PR #6, opened 2019-11-27
Serve frontend even with empty database (fix #5)

### LINKED ISSUE #5: Frontend panics when started with an empty database
When starting the frontend with an initialized but empty database it will throw the following error:
```
INFO[0000] Error retrieving latest epoch from the database: sql: Scan error on column index 0, name "max": converting NULL to uint64 is unsupported module=services
```

guybrush:


alokeutpal:
Mother C

PR #9, opened 2019-12-03
Implement search with typeahead

guybrush:


cmd/explorer/main.go
```diff
@@ -96,6 +96,7 @@ func main() {
 		router.HandleFunc("/validators/data/active", handlers.ValidatorsDataActive).Methods("GET")
 		router.HandleFunc("/validators/data/ejected", handlers.ValidatorsDataEjected).Methods("GET")
 		router.HandleFunc("/search", handlers.Search).Methods("POST")
+		router.HandleFunc("/searchAhead/{type}/{search}", handlers.SearchAhead).Methods("GET")
```
ppratscher: I don't believe we need a dedicated enpoint for the new search. You should be able to use the old /search endpoint with the new parameters.

handlers/search.go
```diff
@@ -13,17 +17,67 @@ func Search(w http.ResponseWriter, r *http.Request) {
...
-		http.Redirect(w, r, "/validator/" + search, 301)
+		http.Redirect(w, r, "/validator/"+search, 301)
 	} else {
 		http.Error(w, "Not found", 404)
 	}
-}
\ No newline at end of file
+}
+
+type SearchAheadBlocksResult []struct {
```
ppratscher: Please move type definitions into the dedicated types package

cmd/explorer/main.go
guybrush: I am now doing:
```golang
router.HandleFunc("/search", handlers.Search).Methods("POST")
router.HandleFunc("/search/{type}/{search}", handlers.SearchAhead).Methods("GET")
```

handlers/search.go
guybrush: Moved it to `types/templates.go`

PR #13, opened 2019-12-17
Make validators sortable

guybrush:
I think it is useful to be able to sort validators


handlers/validators.go
```diff
@@ -16,6 +16,79 @@ import (
 
 var validatorsTemplate = template.Must(template.New("validators").ParseFiles("templates/layout.html", "templates/validators.html"))
 
+type ValidatorDataQueryParams struct {
```
guybrush: Was not sure about this, maybe we should put this also into types-module? (helper-function inclusive)

PR #15, opened 2019-12-18
Fix chart-glitching (#14)

### LINKED ISSUE #14: Glitching axis labels on validator chart
See validator https://beaconcha.in/validator/b8a113e73760699c593762f7816660553b6ce21dadac05b7c75eeafcf56fbcfa6af876166cd6652a1e441e919d1474fe

The axis labels on the first and second y axis do not match the displayed data.

guybrush:
This fixes #14, though I am pretty sure it was `toFixed(2)` already some time ago and changed to `toFixed(0)` for some reason?

### DIFF
--- a/templates/validator.html
+++ b/templates/validator.html
@@ -96,39 +96,17 @@
                     }
                 },
                 labels: {
-                    formatter: function () {
-                        return this.value.toFixed(0);
-                    },
                     style: {
                         color: 'black'
                     }
                 },
                 opposite: false
-            }, {
-                title: {
-                    text: 'Effective Balance [ETH]',
-                    style: {
-                        color: '#26232780',
-                        'font-size': '0.8rem'
-                    }
-                },
-                labels: {
-                    formatter: function () {
-                        return this.value.toFixed(0);
-                    },
-                    style: {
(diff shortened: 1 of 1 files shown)


guybrush:
Actually maybe we should just use the default formatter that comes with highcharts, it will determine the amount of decimals on its own depending on the chart-data (which is 3 digits after comma most of the time).

ppratscher:
Please also move both series to the same Y axis and remove the second Y axis.

guybrush:
This PR includes now:
* use default formatter in balance-graph (instead of `toFixed(0)`) which will determine amount of digits after comma on its own
* use only 1 y-axis for balance-graph (`Balance [ETH]`) on the left side
* use same styling for proposal- and balance-graph (labels were gray and black mixed differently between the 2 graphs)
* fix error when there are no proposals (just make an empty array for now)

The graphs look like this now:
Screenshot from 2019-12-18 09-15-29

When there are no proposals:
Screenshot from 2019-12-18 09-16-01

guybrush:
With the latest change the proposal-graph will not be displayed at all if there are no proposals

PR #18, opened 2019-12-18
Lint

mkinney:
Fix most golint warnings. There are still 12 warnings.

Will need help/input on remaining warnings. (such as "type RpcClient should be RPCClient" or "don't use underscores in Go names; struct field Header_1 should be Header1".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ docker/
 templates/imprint.html
 config.yml
 config-lighthouse.yml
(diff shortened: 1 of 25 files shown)


ppratscher:
Thanks a lot for your efforts. The pr looks good to me, I will merge it as it is as it touches a lot of code. Regarding the remaining renames feel free to make them in a separate pr.

PR #23, opened 2019-12-19
looks like it is working

mkinney:
Note: Not persisting the postgres db but probably should.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,41 @@
+version: '2.4'
+services:
+
+  prysm:
+    image: gcr.io/prysmaticlabs/prysm/beacon-chain:latest 
+    ports:
+      - "4000:4000"
+      - "7000:7000"
+    volumes:
+      - $HOME/prysm-data:/data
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+    command: --archive --grpc-gateway-port=7000 --datadir=/data
+
+  postgres:
+    image: postgres
+    environment:
+      - POSTGRES_PASSWORD=secret
+      - POSTGRES_USER=explorer
+      - POSTGRES_DB=explorer
+    volumes:
+    - ./tables.sql:/docker-entrypoint-initdb.d/init.sql
```
guybrush: minor nit: use same indentation as with other fields (same with `services.explorer.volumes`)

mkinney: Thanks. Will do.

mkinney:
I'm going to close this PR as I want to add a docker-compose for lighthouse, too.

I'll re-open a new PR.

PR #25, opened 2019-12-27
add lighthouse chain link

mkinney:


asaneh:
#394 #1241 #384 #1267 #1268

PR #28, opened 2020-01-16
Estimate activationEpoch for pending validators

guybrush:


db/db.go
```diff
@@ -334,6 +334,21 @@ func saveValidatorSet(epoch uint64, validators []*types.Validator, validatorIndi
 	}
 	defer stmtValidators.Close()
 
+	var lenActivatedValidators int
+	var lastActivatedValidatorIdx uint64
+	for _, v := range validators {
+		isActive := v.ActivationEpoch <= epoch && epoch < v.ExitEpoch
```
ppratscher: I don't believe the temp variable here is really required

guybrush: Thought its easier to read, anyway removed with force-push!

PR #35, opened 2020-01-20
Store validators-state of latest epoch in validators-table

guybrush:
With this PR we get much faster response-timings for `/validators/data/{pending,active,ejected}` routes (~70ms vs ~1000ms).

It expands the already existing `validators`-table to match the `validator_set`-table (plus the already existing publickey-field).


guybrush:
Ah with this, the integration-tests also need update - on it.

PR #40, opened 2020-01-23
Show offline valiators on validators-page

guybrush:
Offline validators are validators that have not attested in both, the current epoch and the previous epoch.

Maybe we should also show when they last attested?


guybrush:
Please do not merge yet, the query is currently too slow when we include last-attestation-info. I am trying to make this more performant, but maybe it is better to just no show last-attestation-info for now.

guybrush:
Ok I got it fast again, by changing the database-schema. 

This PR introduces yet another column to the `validators`-table: `lastattestationslot` which tracks the last attestations of each validator. This column is then used to see which validators are offline. An offline validator is active for >1 epochs and did not attest in the last 2 epochs (including the current).

Changing the database for this PR:
```sql
alter table validators add lastattestationslot bigint;

update validators 
set lastattestationslot = a.lastslot
from (
    select validatorindex, max(attesterslot) as lastslot
    from attestation_assignments
    group by validatorindex
) a where a.validatorindex = validators.validatorindex;
```

guybrush:
This PR now changes:
- backend
  - add `lastattestationslot` to `validators`-table
- frontend
  - validators-page
    - use 1 table instead of 3 and add filter for validator-status
    - table-columns will be displayed depending on validator-status-filter
    - mark offline validators with icon (offline validators are {active,slashing,exiting} for at least 2 epochs and did not attest in the last 2 epochs)
  - validator-page
    - show offline-icon if validator is offline
    - remove slashed-information, now the validator state is `Slashing` if the validator got slashed

We still need to fix:
- validators are not marked as `Exiting` yet, we only show when a validator exited
- offline-indication depends on getting the correct data from beacon-node, which is currently not heavily tested

PR #128, opened 2020-06-16
Design/clean

qu0b:
* switched from development version bootstrap to stable version 4.5
* added the layout styles to the to the bootstrap css theme
* changed the dark/light toggle to change style sheets instead of just css variables


qu0b:
* Removed boostrap folder from the project
* Added a home button

Issue #124, opened 2020-06-14
Header: "Epoch Slot" to "Epoch current_slot/scheduled_slot" 


Buttaa:
Currently shows the last scheduled slot until a new epoch starts. 
Better UX by showing `Epoch, current_slot / last_scheduled_slot`. 

![firefox_2020-06-14_12-09-42](https://user-images.githubusercontent.com/26490734/84590569-1dd32580-ae38-11ea-903f-1fef5386da55.png)

qu0b:
Also add a finalization-gap indicator to the status field if the gap is high.

PR #139, opened 2020-06-22
Feature/statusbar

qu0b:
Added to the banner:
* finalization delay
* sync status
* current current slot
* current epoch
* live updates every 12 seconds


qu0b:
The status bar now shows the current slot that is getting proposed #124

PR #164, opened 2020-07-09
Poap - Proof of Attendance Protocol design update

qu0b:


guybrush:
This features is only relevant for the next upcoming multi-client-testnet as far as I understand @Buttaa ?

Buttaa:
@guybrush yes, let's not confuse current users and merge it ones the new deposit contract is deployed

Issue #114, opened 2020-06-01
Add commas to Beaconcha.in to separate numbers


pjindal:
Small request, but can you add commas to all the large numbers on beaconcha.in -- currently it is a little difficult to read (e.g. instead of 1000, write 1,000): 

![image](https://user-images.githubusercontent.com/26806409/83428023-a943c400-a3ff-11ea-9c6d-1812147431bd.png)

PR #216, opened 2020-08-24
Update/index page

qu0b:


qu0b:
Adds commas to balances as referenced in #114

guybrush (approved):

PR #289, opened 2020-10-12
clarifies balance types on the frontpage

Buttaa:


qu0b (approved):
"Average staked (current) balance of all validators" - whats the added information with the (current) comments

PR #297, opened 2020-10-19
Adds pubkey lookups to v1 API

CarlBeek:
As discussed asynchronously, we're really keen to use your API to enhance the Launchpad. To do so, we need a few extra API interfaces, primarily to look up validators by their `pubkey` (we don't know their `index` at the time of deposit).

This PR adds the following API options:
* `ApiValidatorByPubkey` - This is the same as `ApiValidator` except look-ups are by `pubkey`, not `index`
* `ApiValidatorByEth1DepositPubkey` - This interface is added to be able to look up validators who have deposited on eth1, but not had their deposits processed on eth2 yet.

I welcome any and all changes to this PR (including scraping it and starting again) as I am not really familiar with your code-base, this is my first time using golang, and I have not tested this code.

PR #298, opened 2020-10-19
Enable CORS for the v1 API

CarlBeek:
As discussed in #297, we're keen to use this API with the eth2 launchpad, but in order to do so, we need some changes. This PR implements what I think is needed to enable CORS on the v1 API. Again, I have not tested it, I'm new to golang, and I have very little idea how `gorilla/mux` works. :)

PS: If #297 gets merged first, then the header must be extended to allow cross origin requests too for `ApiValidatorByEth1DepositPubkey`, and `ApiValidatorByPubkey`.


guybrush (approved):

CarlBeek:
Closing due to the cleaner solution in #300. Thanks @guybrush!

PR #300, opened 2020-10-20
Api: query validators by indexOrPubkey and enable CORS

guybrush:
This incorporates ideas and implementation from #297 and #298 (can be closed when merging this).


qu0b (approved):

CarlBeek:
This is freaking awesome, @guybrush. 🎉  A much cleaner solution overall due to the reuse of the existing interface instead of adding extra commands to the existing API.

PR #355, opened 2020-11-12
add inclusion-distance-chart

guybrush:


guybrush:
last 7 days for now, we need to calculate the avg inclusion delay per epoch and put that into the db in order to have performant enough chart-generation for this (and other charts)

qu0b (approved):

PR #384, opened 2020-11-20
fix onlyAPI

guybrush:

PR #394, opened 2020-11-24
Multilang update

Vahegian:
Put the language selector at the top, and using cookies instead of query strings.


qu0b (dismissed):

qu0b (approved):

asaneh:
#1241 #384 #1268 #1267 #1264

PR #396, opened 2020-11-24
List Allnodes as staking service

3eph1r0th:
List Allnodes as staking service


Buttaa:
Done in https://github.com/gobitfly/eth2-beaconchain-explorer/pull/398

PR #443, opened 2020-12-02
Updatedocs

Buttaa:


kevinzen (commented):
Documentation strings are improved by this commit. It may be worth considering adding some reference to how/why i81n not supported.

PR #508, opened 2020-12-07
Translation index page

Vahegian:
add utils.GetTemplateFuncs() call to template renders that didn't have it,
replaced words on index page with {{ trLang .Lang "original word" }} (multiple files were modified with this.)
add "Lang" function to pageData type, it is used for getting the user's browser language or preference from cookies.


Vahegian:
merging to feature/staging

PR #518, opened 2020-12-09
Feature/staging

qu0b:


positivenoise (approved):

db/db.go
```diff
@@ -1181,6 +1181,13 @@ func GetTotalValidatorsCount() (uint64, error) {
 	return totalCount, err
 }
 
+// GetActiveValidatorCount will return the total-validator-count
+func GetActiveValidatorCount(currentEpoch uint64) (uint64, error) {
+	var count uint64
+	err := DB.Get(&count, "SELECT COUNT(*) FROM validators WHERE activationepoch >= $1 AND $2 < exitepoch", currentEpoch, currentEpoch)
```
peterbitfly, line 1187: Needs an db index in order to be performant, otherwise the query will cause a full table scan

handlers/validator.go
```diff
@@ -870,3 +917,161 @@ func ValidatorSave(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "Internal server error", 503)
+		return
+	}
+	if length > 100 {
+		length = 100
+	}
+
+	var totalCount uint64
+	err = db.DB.Get(&totalCount, `
+		select
```
peterbitfly, line 960: why a nested select? also this will always be equal to the number of epochs the validator is active because there is one entry per epoch in the validator_balances table and each validator is scheduled to attest once per epoch

guybrush (dismissed):

peterbitfly (changes requested):

peterbitfly (approved):

PR #585, opened 2021-01-04
Db partitioning

peterbitfly:
Implements partitioning of the validator_balances and the attestation_assignment tables


rpc/prysm.go
```diff
@@ -50,32 +50,32 @@ func NewPrysmClient(endpoint string) (*PrysmClient, error) {
 	}
 	client.assignmentsCache, _ = lru.New(10)
 
-	streamChainHeadClient, err := chainClient.StreamChainHead(context.Background(), &ptypes.Empty{})
-	if err != nil {
```
Nieks777: rpc/prysm.go

PR #589, opened 2021-01-05
add streak count to validator view

Vahegian:


handlers/validator.go
```diff
@@ -346,8 +346,30 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 		validatorPageData.AttestationsCount = validatorPageData.ExitEpoch - validatorPageData.ActivationEpoch
 	}
 
+	var attestationStatusHistory []int64
+	err = db.DB.Select(&attestationStatusHistory, "select status from attestation_assignments_p where validatorindex=$1 order by epoch desc;", index)
```
peterbitfly, line 350: this query will not scale well as it will fetch the complete attestation history of the validator, please put this on hold until the new stats module is deployed

peterbitfly (changes requested):

handlers/validator.go
Vahegian: okay, thank you

axljoker03 (approved):

PR #612, opened 2021-01-14
Ensure pixelated rendering for the graffitiwall

qix:
Zooming into an area of the wall right now will blur the pixels, but in this case the image is better pixelated.

I wasn't sure if you would prefer throwing this in the `define "css"` block but since it's such a minimal and specific edit just putting it in the HTML tag felt appropriate.


guybrush (approved):

guybrush:
Thank you!

PR #661, opened 2021-02-15
Validator stats table view

peterbitfly:
Add pagination and table export


qu0b (changes requested):
Would be great to store the dependencies in our static folder instead of using a cdn

qu0b (approved):

PR #697, opened 2021-03-15
Feature/pools

Vahegian:


gigilove2001 (dismissed):

gigilove2001 (approved):
Amazing work guys I'm blown away.. 
thank you ...

handlers/pools.go
```diff
@@ -0,0 +1,268 @@
...
+	var pieChart chart
+
+	indexStats := services.LatestIndexPageData()
+
+	pieChart.DepositDistribution.Data = chartData
+	pieChart.DepositDistribution.Height = 500
+	pieChart.DepositDistribution.Path = "deposits_distribution"
+	pieChart.StakedEther = indexStats.StakedEther
+
+	updateMux.Lock()
```
qu0b, line 88: It would make more sense to have this in a service instead of in the handler. We should be able to call handlers concurrently.

services/charts_updater.go
```diff
@@ -1467,24 +1469,174 @@ func depositsDistributionChartData() (*types.GenericChartData, error) {
...
+
+		if skipAppend {
 			continue
 		}
-		seriesData = append(seriesData, seriesDataItem{
-			Name: string(utils.FormatEth1AddressString(rows[i].Address)),
-			Y:    rows[i].Count,
-		})
+
+		if drillSeries[len(drillSeries)-1].ID == "Others" {
```
qu0b, line 1632: if len(drillSeries) is 0 this will cause a runtime exception

qu0b (changes requested):

services/charts_updater.go
Vahegian: The drillSeries  is hard-coded so it always has a length grater than one, just have to make sure that the last one is always "Others"

handlers/pools.go
Vahegian: this is one of the thing that concerns me too, I wanted to see if it becomes a bottleneck, but you are right maybe I should move it to services

qu0b (approved):

PR #734, opened 2021-04-02
users: fix updating email-addresses

### LINKED ISSUE #733: Bug: Can't Change Email Address to one that includes a Plus symbol
The Plus symbol "+" is a valid character in an email address. However it is not currently possible to change your email address on Beaconcha.in to one that includes a plus symbol

Beaconcha.in correctly allows users to subscribe to emails using an email address that includes a plus sign.

But if a user attempts to change their email address to one that includes a plus sign, it doesn't work. The change email process breaks down after clicking the link to validate the new email. When the user tries to log in with their updated email, the system says there is no such address. At that point it is also no longer possible to log in with the old email address.

One reason this is important to fix: As a helpful bonus feature, Gmail allows the use of the plus symbol for email filtering. Everything after the plus (inclusive) but before the @ sign is ignored for email delivery purposes (it all goes to the same Gmail account). But Gmail allows the user to sort and filter email based on the extra information after the plus symbol.

guybrush:
fix #733 

q.Get() is already unescaping

https://play.golang.org/p/N4hAN9U26Y0

### DIFF
--- a/handlers/user.go
+++ b/handlers/user.go
@@ -645,9 +645,11 @@ func UserConfirmUpdateEmail(w http.ResponseWriter, r *http.Request) {
 	}
 
 	q := r.URL.Query()
-	newEmail, err := url.QueryUnescape(q.Get("email"))
-	if err != nil {
-		utils.SetFlash(w, r, authSessionName, "Error: Could not update your email please try again.")
+
+	newEmail := q.Get("email")
+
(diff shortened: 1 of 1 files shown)


guybrush:
`q.Get()` is already unescaping

https://play.golang.org/p/N4hAN9U26Y0

qu0b (approved):

PR #741, opened 2021-04-06
add func to insert chart data

Vahegian:
Please insert this table into db before merging
```
create table staking_pools_chart
(
    epoch                      int  not null,
    name                       text not null, 
    income                     bigint not null, 
    balance                    bigint not null, 
    PRIMARY KEY(epoch, name, income, balance)
);
```

Values older than 6 months will be automatically removed from the table


alisscco (approved):

l2ichshop2 (approved):
Oh Arthit 
ขอบคุณครับ

l2ichshop2 (dismissed):
OhArthit
ขอบคุณครับ

alisscco (dismissed):

guybrush (approved):

PR #742, opened 2021-04-06
Revert "add new ad-zone, fix stakingServices"

qu0b:
Reverts gobitfly/eth2-beaconchain-explorer#740


alisscco (commented):
Thanks the list ..

PR #793, opened 2021-04-27
staging

qu0b:
- update/bundle (#788)
- update dockerfile
- API: validator queue endpoint
- Bundle: fix paths
- Bundle: fix paths


qu0b:
adds endpoint for entering and exiting queue  #757

guybrush (approved):

PR #837, opened 2021-05-26
more accurate Rocket Pool staker commission

dEEtooo:
regular stakers (non-operators) pay the average of all commissions across the protocol, which is weighted heavily to 10%.  it'll be a variable % correlating with demand for operators and most likely it'll be around 10%.  tried to make this clear with the ~ symbol and the use of "(variable)".  do not want persons viewing this page to think they're likely to pay a 5% commission or a 20% commission, which probably isn't possible given the math/reality.


guybrush:
ping @Buttaa

PR #1041, opened 2021-09-11
Fix bitlist format

k1rill-fedoseev:
In the fixed line, `y` should be a start byte index, not a line number. Otherwise, the same bits are displayed again and again, for example:

Screenshot 2021-09-11 at 16 49 22


guybrush:
this is in #1123 

thank you!

PR #1206, opened 2022-01-19
@RainerBitfly: Added View Voluntary Exits and fixed the view of other…

RainerBitfly:
… states (active online, active offline,...)

### DIFF
--- a/handlers/validators.go
+++ b/handlers/validators.go
@@ -34,33 +34,46 @@ func Validators(w http.ResponseWriter, r *http.Request) {
 	}
 
 	latestEpoch := services.LatestEpoch()
-
 	validatorOnlineThresholdSlot := GetValidatorOnlineThresholdSlot()
+
 	for _, validator := range validators {
-		validatorsPageData.TotalCount++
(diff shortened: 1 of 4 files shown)


types/templates.go
```diff
@@ -232,6 +232,7 @@ type ValidatorsPageData struct {
 	ExitingOnlineCount   uint64
 	ExitingOfflineCount  uint64
 	ExitedCount          uint64
+	VoluntaryExitsCount  uint64
```
guybrush, line 235: Exiting = voluntary exit
Slashing = forced exit

handlers/validators.go
```diff
@@ -34,33 +34,46 @@ func Validators(w http.ResponseWriter, r *http.Request) {
 	}
 
 	latestEpoch := services.LatestEpoch()
-
 	validatorOnlineThresholdSlot := GetValidatorOnlineThresholdSlot()
+
 	for _, validator := range validators {
```
guybrush, line 39: all the checks in this for-loop can be replaced by checking against `validator.Status`

Issue #1304, opened 2022-05-03
Can't unsubscribe client notification


xanatos:
In the https://beaconcha.in/user/ethClients page, if I have subscribed to a client notification, and I try to unsubscribe, the change won't be saved. There doesn't seem to be a way to unsubscribe. Note that it seems that the page auto-saves the changes, so if you click a checkbox to subscribe, it is immediately saved, while if you click to unsubscribe the change isn't saved correctly.

qu0b:
The UI seems to work fine for me. Would it be possible for you go give us more details such as console logs etc.

xanatos:
Using Chrome 100 on Windows: "Versione 100.0.4896.127 (Build ufficiale) (a 64 bit)" and/or Ede: "Versione 101.0.1210.32 (Build ufficiale) (64 bit)" (I know that they are both "chromium" inside). I'm using an Italian computer if it is important.

Just tested if there is the same problem if I use Chrome + the Incognito Mode (so that all the extensions are disabled) and the problem remains.

The only two errors (and technically they are warnings) I get are:

DevTools failed to load source map: Could not load content for https://beaconcha.in/bundle/js/popper.min.js.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE
112 The resource <URL> was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.

The 112 is because the second warning "spawns" 112 times

xanatos:
Just checked and the POST to https://beaconcha.in/user/notifications/unsubscribe?filter=teku&event=eth_client_update is done and the return value is 200, so it is a backend problem and/or account corruption

qu0b:
Okay thank you for the update, we are looking into it

qu0b:
We've pushed an update, let us know if that helped and thanks for bringing up the issue.

xanatos:
Works correctly thanks.

PR #1308, opened 2022-05-04
fix/ethClientSubscriptions

qu0b:
- Tekumetrics (#1265)
- gitcoinfeed: increase fetch-interval from 5s to 120s
- Update Rocket Pool and Btifrost (#1270)
- Stripetx (#1294)
- Statslogcleanup (#1295)
- db: limit connections
- frontend: set blocks search timeout to 30s
- fix ethclient notification unsubscribe


qu0b:
This should fix issue #1304

PR #1347, opened 2022-05-30
add https to mailgun

Buttaa:


xstaticxgpx:
Shouldn't this link to https://beaconcha.in/slot/... now?

https://beaconcha.in/block/4704415

vs.

https://beaconcha.in/slot/4704415

PR #1352, opened 2022-06-01
Update default config files

RomiRand:
The config file layout has been recently changed in d0a7488f742dbaad474509310f24860712c6309b. This change isn't reflected in the default config files yet, which makes it hard for new users to get started.
So I updated the default config files.

### DIFF
--- a/config-example.yml
+++ b/config-example.yml
@@ -1,17 +1,26 @@
 # Database credentials
-database:
+readerDatabase:
+  user: "<dbuser>"
+  name: "<dbname>"
+  host: "<dbhost>"
+  port: "<dbport>"
+  password: "<dbpassword>"
+writerDatabase:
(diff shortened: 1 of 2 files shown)


fengling229:
> The config file layout has been recently changed in [d0a7488](https://github.com/gobitfly/eth2-beaconchain-explorer/commit/d0a7488f742dbaad474509310f24860712c6309b). This change isn't reflected in the default config files yet, which makes it hard for new users to get started. So I updated the default config files.

Now the default config is behind the explorer program again. I'm a new one and want to run this explorer program. I have run a goerli testnet using geth 1.0.26 as the execution client and prysm 3.1.2 as the consensus client, and they worked well. But I don't know how to configure this config file, would you wish to help me?

axljoker03 (approved):

PR #1355, opened 2022-06-07
feature/webhookDebug

qu0b:
- add simle webhook debug modal
- add webhook debug modal


qu0b:
migrations necessary on the users database

```sql
alter table users_webhooks add column request jsonb;
alter table users_webhooks add column response jsonb;
```

PR #1372, opened 2022-06-15
New validator staking pool view

beaconchain:
This PR replaces the currently very complicated pool view generation with a simplified version. Validator pool tagging is now done via a dedicated table called validator_pool. When deploying this PR the database has to be updated to include the new table:

```
drop table if exists validator_pool;
create table validator_pool
(
    publickey bytea not null,
    pool      varchar(40),
    primary key (publickey)
);
```

Validator pool tags need to be written into the new table either via direct sql inserts or via a dedicated tagging tool that tags new validators based on on-chain heuristics (deposit addresses, withdrawal credentials, deposit behavior).

The `/pools` view has been simplified and currently only shows the current pool share distribution and the average pool performance calculated over the last 1, 7 and 31 days.

This PR is the basis of a more comprehensible and scalable staking pool analytics module.


templates/pools.html
```diff
@@ -0,0 +1,577 @@
...
+                    </table>
+                </div>
+                <span class="px-2" id="lastUpdate"></span>
+            </div>
+        </div>
+    </div>
+</section>
+<div id="tempcont" class=""></div>
+<script src="/js/jquery.min.js"></script>
+<script src="/js/luxon.min.js"></script>
```
qu0b, line 486: This should also be in the JS block

templates/pools.html
```diff
@@ -0,0 +1,577 @@
...
+<script>
+
+    window.addEventListener('load', function () {
+        Highcharts.chart('pools-distribution-chart', {
```
qu0b, line 490: This can stay here, but calling a function from the JS block and passing the golang object would be a bit cleaner

PR #1373, opened 2022-06-15
Fix linter warnings

beaconchain:
Fixes all warning raised by go-staticcheck

### DIFF
--- a/exporter/exporter.go
+++ b/exporter/exporter.go
@@ -75,7 +75,7 @@ func Start(client rpc.Client) error {
 			logger.Fatal(err)
 		}
 
-		for epoch := head.HeadEpoch - 1; epoch >= 0; epoch-- {
+		for epoch := head.HeadEpoch - 1; true; epoch-- {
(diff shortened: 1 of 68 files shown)


exporter/exporter.go
```diff
@@ -75,7 +75,7 @@ func Start(client rpc.Client) error {
 			logger.Fatal(err)
 		}
 
-		for epoch := head.HeadEpoch - 1; epoch >= 0; epoch-- {
+		for epoch := head.HeadEpoch - 1; true; epoch-- {
```
qu0b, line 78: true here, is that right?

qu0b: there might be a chance that epoch goes below 0 when the client returns an error and continue is called

beaconchain: epoch is an uint hence can never go below 0

qu0b: https://go.dev/play/p/w3U4mMxbPCL

axljoker03 (approved):

PR #1380, opened 2022-06-21
fix getPoolsPageData

guybrush:


Athayde3586:
Need to add my self as a validator

axljoker03 (approved):

PR #1408, opened 2022-06-30
Fix uint64 overflow on initial sync in `eth1DepositsExporter()`

mksh:
Problem: while synchronizing deposits from private eth1 network (ganache),
which has eth2 deposit contract deployed at one of the very first blocks (for example, `1`),
the logic around computing `fromBlock` can detract 100 from `toBlock` value that is less than 100,
due to small block height at a time.

This leads to overflowing `fromBlock` value, passing a value near `2**64` to eth1 rpc, 
and the log in eth2 explorer like following:

time="2022-06-30T21:46:28Z" level=error msg="error fetching eth1-deposits" error="error getting logs from eth1-client: Number can only safely store up to 53 bits" fromBlock=18446744073709551582 module=exporter toBlock=66

This patch makes sure there will be no overflow when importing deposits on the very start of private network,
even if it will fetch less than 100 block on the initial import.


guybrush:
thank you!

PR #1476, opened 2022-08-10
Feature/add accurate queue estimator

InvisibleBitfly:
adds accurate activation estimation and queue position to validator page.

the estimation error i have observed over a set of around 900 validators is pretty much 0, with a small subset of validators being one epoch earlier than predicted (but not really worth it to look into that much deeper).

also change the validator lifecycle steps a bit to better visualize being queued.
| pending and queued | pending and not queued |
| ------ | ----- |
| pending_and_queud | pending_to_active |

also added a tooltip to show the current churn rate when queued
<p align="center">
  
</p>

-----

required migration queries:
```sql
CREATE TABLE validator_queue_deposits (
	validatorindex int4 NOT NULL,
	block_slot int4 NULL,
	block_index int4 NULL,
	FOREIGN KEY (block_slot,block_index) REFERENCES blocks_deposits(block_slot,block_index),
	FOREIGN KEY (validatorindex) REFERENCES validators(validatorindex)
);
CREATE INDEX idx_validator_queue_deposits_block_slox ON validator_queue_deposits (block_slot);
CREATE UNIQUE INDEX idx_validator_queue_deposits_validatorindex ON validator_queue_deposits (validatorindex);
```

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -1700,7 +1700,7 @@ func GetQueueAheadOfValidator(validatorIndex uint64) (uint64, error) {
 	select count(*)
 	from (
 		select validatorindex, block_slot, block_index 
-		from validator_queue_deposits vqd
+		from validator_queue_deposits
 	) as vqd 
 	inner join validators on
 		validators.validatorindex = vqd.validatorindex and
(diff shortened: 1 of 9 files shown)


InvisibleBitfly:
made the queue position bold to better indicate that one can hover over it, the same way the epoch and date above it are.
Screenshot from 2022-08-10 14-20-38

PR #1477, opened 2022-08-15
Fix a pet peeve

jshufro:
image


guybrush:
thanks, indeed it was kind of annoying :)

PR #1486, opened 2022-08-17
exporter: update epoch finalization each exported epoch

InvisibleBitfly:
the 100% correct solution would be to remove the finalized column from the blocks table and only store what the finalized epoch is, but this will work for now. 
query is already optimized to only update the epochs that change in status, so there should be no real perf impact.


InvisibleBitfly:
deprecated by #1497

PR #1497, opened 2022-08-22
Feature/remove finalized column

InvisibleBitfly:
> **Note**
> needs to be tested with bigtable changes

-----

- the `finalized` column has been renamed to `completeparticipationstats` and now tracks if the participation stats have been requested using the epoch after it.
    - this flag is used to style the participation progress bar as an intermediate type if false: image
    - this flag would also allow us to more easily save redundant epoch statistics exports in the future (dont need to update them if they are complete and the epoch is finalized) 
 - a value of 0 for the eligible ether column is now rendered as `Exporting ...` 
image
- epochs that haven't been exported yet are shown on the index page image
- fixed the "syncing" state not being set correctly on initial index load image 
- the index page progress bar is now aware of the sync state image
- finality delay is now always calculated using epoch numbers from the node, preventing false positives during explorer sync
- API should be unaffected, as the finality for epochs is simply calculated on demand. unexported epochs are not exposed over the API and only exist virtually on the index page.
- the app checks for finality issues using epoch participation, so unaffected by this change.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -1591,23 +1591,6 @@ func UpdateEpochStatus(stats *types.ValidatorParticipation) error {
 	return err
 }
 
-// UpdateEpochFinalization will update finalized-flag of unfinalized epochs
-func UpdateEpochFinalization(finality_epoch uint64) error {
-	// to prevent a full table scan, the query is constrained to update only between the last epoch that was tagged finalized and the passed finality_epoch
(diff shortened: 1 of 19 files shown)


InvisibleBitfly:
too much has changed on staging for this to be merge-able, needs a rewritten version from scratch, i cant solve these conflicts without wasting too much time

PR #1499, opened 2022-08-24
Feature/exporter improvements

peterbitfly:
This PR represents a major refactoring of the eth2 data export module that addresses the performance issues on networks with a high degree of validators (mainnet, görli).

On average this PR brings down the full epoch export time to around 1 minute where the most time is currently still spent in updating the validators table. This issue will be addressed in a subsequent PR.

With this PR the explorer will have a hard dependency on bigtable which is used to store the validator balance history, proposal history, attestation history and sync committee duty history. Bigtable allows us to store this data in a scalable and performant way.

We understand that the reliance on bigtable might impose a challenge for small testnet setups. For this use case we suggest to use an open source bigtable port called little bigtable available at https://github.com/bitly/little_bigtable which uses sqlite as data storage engine.


giacomolicari:
@peterbitfly is there available any config example explaining how to make eth2-beaconchain-explorer work with [little_bigtable](https://github.com/bitly/little_bigtable)?
It would be very beneficial also for other chains if we could share that info.
Thank you!

Issue #1523, opened 2022-09-06
FR: Graffiti data request parameters


benv666:
Right now the only way to fetch the Graffiti Wall pixels is using the [/api/v1/graffitiwall](https://beaconcha.in/api/v1/docs/index.html#/Graffitiwall/get_api_v1_graffitiwall) API, which as advertised returns all pixels on the wall.

However, for certain applications such as Rocket Pool's Graffiti Wall Writer addon that only care about a specific section of the wall it would be nice to be able to fetch only a segment, as the rest of the data is irrelevant.
Currently this is no big deal yet with "only" 6.5K pixels, but given the size of the canvas this might become painful sooner or later. E.g. imagine pulling in 1000x1000 pixels in json, with ~63 bytes per pixel this is almost 64MB of results.

So the feature request is simple:
Please add a set of parameters to allow for fetching a dedicated segment of the graffitiwall, e.g. starting X,Y coordinates and height/width, or even starting X,Y and ending X,Y.

Example:
```
# GET /api/v1/graffitiwall/{startX}/{startY}/{endX}/{endY}
GET /api/v1/graffitiwall/100/100/200/200
```
Would return all pixels on the wall between (100,100)-(200,200).

recy21:
Done, will be part of next Release (tomorrow)

PR #1525, opened 2022-09-06
Api: graffitiwall api params [#1523]

benv666:
This is my naive stab at implementing #1523.
Untested.


InvisibleBitfly:
if this used optional query parameters instead i think this would be quite an easy merge. the resulting endpoint functionality would then be `/api/v1/graffitiwall?startx=500&endy=500`. just have sensible defaults to params that haven't been passed (start = 0, end = 1000)

axljoker03 (approved):

PR #1563, opened 2022-09-16
Update rocketscan url from .dev to .io

grok13:
Rocketscan moved to .io in April 2022.
There is a redirect in place, but we may as well update it.


Buttaa:
This was fixed a while ago :) thanks

PR #1628, opened 2022-10-21
optimized eth1 deposits fetching

alexqrid:
Hey!
### Problem
Noticed that you do not fully utilize `FilterLogs` capabilities. I've tested almost a year ago the perfomance with filtering topic on geth level, and found it much more efficient than filtering a bunch of logs in the code.
### Solution
Providing the topic to the RPC method `FilterLogs` does the thing better, than trying to make comparison of the topic inside the code, as log filters are more optimized on geth level.
Making this PR as it brings a huge optimization for eth1 deposits fetching.


slumericanB (approved):

guybrush:
thank you!

axljoker03 (approved):

dm7324154:
How do I get my deposit

guybrush:
this lands in #2050

thank you and sorry for the long delay

PR #1648, opened 2022-11-07
Improve validator offline detection

D13ce:
The threshold for validators to appear offline now is based on finalized epochs rather than on slots.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -127,7 +127,7 @@ func GetEth1Deposits(address string, length, start uint64) ([]*types.EthOneDepos
 
 var searchLikeHash = regexp.MustCompile(`^0?x?[0-9a-fA-F]{2,96}`) // only search for pubkeys if string consists of 96 hex-chars
 
-func GetEth1DepositsJoinEth2Deposits(query string, length, start uint64, orderBy, orderDir string, latestEpoch, validatorOnlineThresholdSlot uint64) ([]*types.EthOneDepositsData, uint64, error) {
+func GetEth1DepositsJoinEth2Deposits(query string, length, start uint64, orderBy, orderDir string, latestEpoch, validatorOnlineThresholdEpoch uint64) ([]*types.EthOneDepositsData, uint64, error) {
 	deposits := []*types.EthOneDepositsData{}
 
 	if orderDir != "desc" && orderDir != "asc" {
@@ -202,9 +202,9 @@ func GetEth1DepositsJoinEth2Deposits(query string, length, start uint64, orderBy
 				CASE 
 					WHEN exitepoch <= $3 then 'exited'
 					WHEN activationepoch > $3 then 'pending'
-					WHEN slashed and activationepoch < $3 and (lastattestationslot < $4 OR lastattestationslot is null) then 'slashing_offline'
+					WHEN slashed and activationepoch < $3 and (lastattestationslot / $5 < $4 OR lastattestationslot is null) then 'slashing_offline'
 					WHEN slashed then 'slashing_online'
-					WHEN activationepoch < $3 and (lastattestationslot < $4 OR lastattestationslot is null) then 'active_offline'
+					WHEN activationepoch < $3 and (lastattestationslot / $5 < $4 OR lastattestationslot is null) then 'active_offline'
 					ELSE 'active_online'
 				END AS state
 				FROM validators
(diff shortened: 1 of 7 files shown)


qu0b (commented):
Please use readerdb for reads & use services for things such as the latest epoch. Services will provide the latest information that is frequently retreived

qu0b (commented):
Why do you need the "Greatest" function here?

db/db.go
```diff
@@ -1295,14 +1107,14 @@ func saveValidators(data *types.EpochData, tx *sqlx.Tx, client rpc.Client) error
 					WHEN EXCLUDED.exitepoch <= %[1]d THEN 'exited'
 					WHEN EXCLUDED.activationeligibilityepoch = 9223372036854775807 THEN 'deposited'
 					WHEN EXCLUDED.activationepoch > %[1]d THEN 'pending'
-					WHEN EXCLUDED.slashed AND EXCLUDED.activationepoch < %[1]d AND GREATEST(COALESCE(validators.lastattestationslot, 0), EXCLUDED.lastattestationslot) < %[2]d THEN 'slashing_offline'
+					WHEN EXCLUDED.slashed AND EXCLUDED.activationepoch < %[1]d AND GREATEST(COALESCE(validators.lastattestationslot / %[4]d, 0), EXCLUDED.lastattestationslot / %[4]d) < %[2]d THEN 'slashing_offline'
```
qu0b, line 1110: Why do you need the GREATEST func here?

db/db.go
```diff
@@ -1013,10 +1013,12 @@ func saveValidators(data *types.EpochData, tx *sqlx.Tx, client rpc.Client) error
 		return err
 	}
 
-	thresholdSlot := latestBlock - 64
-	if latestBlock < 64 {
-		thresholdSlot = 0
+	var lastFinalizedEpoch uint64
```
qu0b, line 1016: A query here for the latest finalized epoch should not be necessary. We have services that store frequently used values such has the latest finalized epoch in memory. Check the services.go file.

db/db.go
```diff
@@ -127,7 +127,7 @@ func GetEth1Deposits(address string, length, start uint64) ([]*types.EthOneDepos
 
 var searchLikeHash = regexp.MustCompile(`^0?x?[0-9a-fA-F]{2,96}`) // only search for pubkeys if string consists of 96 hex-chars
 
-func GetEth1DepositsJoinEth2Deposits(query string, length, start uint64, orderBy, orderDir string, latestEpoch, validatorOnlineThresholdSlot uint64) ([]*types.EthOneDepositsData, uint64, error) {
+func GetEth1DepositsJoinEth2Deposits(query string, length, start uint64, orderBy, orderDir string, latestEpoch, validatorOnlineThresholdEpoch uint64) ([]*types.EthOneDepositsData, uint64, error) {
```
qu0b, line 130: Isn't it enough to show a validator as offline after more than one finalized epoch instead of more than two?

qu0b (changes requested):

db/db.go
D13ce: Quoting the BIDS' description: `The "Active" status on the validator page and dashboard, should only turn red if the validator was not attesting for at least 2 finalized epochs.`

If the last finalized epoch is 1000, the active status should turn red if the latest attestation happened in or before epoch 998. The **validatorOnlineThesholdEpoch** in that case is set to 999 any by comparing it with '<', this should be correct, should it not?

db/db.go
D13ce: This is correct but not straight forward. Using services in this file causes a circular dependency. I tried that but did not want to touch the architecture for this issue, hence the query.
I guess this is also the reason for the `latestBlock` being queried above to get `latestEpoch` instead of using `services.LatestEpoch`.

How do you want me to proceed?

db/db.go
D13ce: We get here when an existing `validatorindex` has to be updated. The whole `ON CONFLICT` block takes the greater value for `lastattestationslot` from either the existing or the new data. This is even done when `lastattestationslot` is set within `ON CONFLICT` in line 1103.

I interpret this as "use the latest of both available `lastattestationslots` as it will be the truly latest one". But to be honest, this handling was here before and I did not question it too much.

db/db.go
D13ce: Regarding the use of the `WriterDb` here:
Based on this [answer](https://stackoverflow.com/c/bitfly/a/124/17), the `WriterDb` should be used for read access if the data may have been changed very recently. Within `func saveEpoch`, the function we are looking at (`func saveValidators`) is called and afterwards the epoch table is updated. So yes, the `ReaderDb`should work here. But if this order is changed for whatever reason in the future, using the `ReaderDb` would break this handling and it would be hard to spot that.

I guess there is a similar reason why the `WriterDb` is also used above to access the `latestBlock`.

If you still want me to change that, I will do that of course. To be consistent, I would suggest to use the `ReaderDb`for the `latestBlock` too in that case. I just wanted to share my train of thoughts.

db/db.go
qu0b: Yes in the bids it says at least 2 finalized epochs. Although in my opinion that is an overkill. If you pass the latestEpoch and the latestFinalizedEpoch to the GetEth1DepositsJoinEth2Deposits function you don't need the validatorOnlineThresholdEpoch. In my opinion it would be cleaner to have an inequality:
`lastattestationslot / $5 < $4` where $5 is the slots per epoch and $4 is the lastFinalizedEpoch.

There are always at least 1 - 2 epochs above the latest finalized epoch and using a value of at least 1 finalzed epoch would mean that the validator has failed to attest in at least 2(-3) epochs.

db/db.go
qu0b: Ah I see its part of the exporter and not of the frontend. Then its fine how it is. I would however similar to above remove the threshold epoch and use the lastFinalizedEpoch.

qu0b: Hard to see what functions some of the code belongs to without looking at the whole file..

db/db.go
qu0b: For example, if the head epoch is 1000 and the latestFinalized epoch is 998 the validator has a chance to attest in epoch 1000, 999, 998 before being considered offline.

qu0b (changes requested):

PR #1655, opened 2022-11-09
feature/mempool

qu0b:
- add mempool route
- add a mempool service
- move types
- improve formatting
- add mempool to nav
- mempool nav logo


qu0b:
This PR adds a mempool view that is updated through a service. The mempool transactions are requested from a geth node every 10 seconds and stored in a shared cache between the frontends. Requests to the mempool page are served via the frontend from the cache.

K0P01IL:
I

PR #1679, opened 2022-11-19
Returned search box in validators leaderboard template

sauliusgrigaitis:
As talked with @Buttaa, this feature was very handy on Goerli network in order to roughly figure out the performance of a group of labeled validators. Hopefully, the search feature is optimized at some point, and this PR will be merged.

### DIFF
--- a/templates/validators_leaderboard.html
+++ b/templates/validators_leaderboard.html
@@ -7,7 +7,7 @@
         processing: true,
         serverSide: true,
         ordering: true,
-        searching: false,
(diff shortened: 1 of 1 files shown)


slumericanB (approved):
yes

guybrush:
Sorry right now this does not work for performance reasons, see https://github.com/gobitfly/eth2-beaconchain-explorer/blob/e0aaa419a6/handlers/validators_leaderboard.go#L103

PR #1692, opened 2022-11-24
fix: avoid crashing the frontend when gas prices can't be calculated

skylenet:
Avoids the frontend crashing when it's not possible to calculate gas fees and we get "nil" instead. 
This was happening in our devnet setup as there are not pending transactions to get an estimate.

``` 
time="2022-11-23T12:39:48Z" level=error msg="error executing template for / route: template: layout.html:195:122: executing \"layout\" at <.GasNow.Data.Fast>: nil pointer evaluating *types.GasNowPageData.Data" module=handlers
2022/11/23 12:39:48 http: superfluous response.WriteHeader call from github.com/urfave/negroni.(*responseWriter).WriteHeader (response_writer.go:53)
```

### DIFF
--- a/templates/layout.html
+++ b/templates/layout.html
@@ -192,7 +192,13 @@
               <div data-toggle="tooltip" title="" data-original-title="Gas Price" class="d-none d-lg-block">
                 <div id="banner-slot" class="info-item d-flex mr-2 mr-lg-3">
                   <div class="info-item-body">
-                    <a id="banner-gpo-data" href="/gasnow"><i class="fas fa-gas-pump mr-1"></i>{{ formatBalanceWei .GasNow.Data.Fast "GWei" 0 }}</a>
+                    <a id="banner-gpo-data" href="/gasnow"><i class="fas fa-gas-pump mr-1"></i>
+                      {{- if .GasNow }}
+                        {{ formatBalanceWei .GasNow.Data.Fast "GWei" 0 }}
+                      {{- else }}
+                        ?
+                      {{- end }}
+                    </a>
                   </div>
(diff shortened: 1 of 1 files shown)


qu0b:
Thanks for the contribution, could you extend the {{ if .Gasnow }} to the complete block. then nothing is shown if there is not GasNow value available.

templates/layout.html
```diff
@@ -192,7 +192,13 @@
               <div data-toggle="tooltip" title="" data-original-title="Gas Price" class="d-none d-lg-block">
                 <div id="banner-slot" class="info-item d-flex mr-2 mr-lg-3">
                   <div class="info-item-body">
-                    <a id="banner-gpo-data" href="/gasnow"><i class="fas fa-gas-pump mr-1"></i>{{ formatBalanceWei .GasNow.Data.Fast "GWei" 0 }}</a>
+                    <a id="banner-gpo-data" href="/gasnow"><i class="fas fa-gas-pump mr-1"></i>
+                      {{- if .GasNow }}
```
qu0b, line 196: Move this below L191 and remove the else ? part

qu0b:
You should be able to spin up the frontend if you run the newest version of the frontend-data-updater against your cache. The error is caused because the cache does not have a value :upside_down_face:

skylenet:
> You should be able to spin up the frontend if you run the newest version of the frontend-data-updater against your cache. The error is caused because the cache does not have a value upside_down_face

That's correct. The cache doesn't have a value because my network doesn't have any pending transactions so that a value can be calculated and set.

PR #1693, opened 2022-11-24
feat(eth1indexer): get chainId directly from config

skylenet:
We're running some temporary testnets that all have different chainIds, so this make it possible to directly pass it in.

### DIFF
--- a/cmd/eth1indexer/main.go
+++ b/cmd/eth1indexer/main.go
@@ -31,9 +31,11 @@ import (
 
 func main() {
 
+	var chainId string
+
+	flag.StringVar(&chainId, "chainId", "", "Chain Identifier")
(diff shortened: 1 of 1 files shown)


qu0b:
We support prater and mainnet chain configs. Maybe a better solution would be to support a custom chain config via a cli flag?
e.g. https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/config/mainnet.chain.yml

qu0b:
I started working on this here: https://github.com/gobitfly/eth2-beaconchain-explorer/compare/master...feature/testnet

qu0b:
yes sounds great!

qu0b:
Just make sure, that its defined

PR #1718, opened 2022-12-15
implement new landing page (BIDS-435)

LuccaBitfly:


D13ce:
You are following the figma here, but may I suggest to add commas for the epoch in the progress bar as well?
image

services/services.go
```diff
@@ -644,34 +635,36 @@ func getIndexPageData() (*types.IndexPageData, error) {
...
+		epoch.GlobalParticipationRateFormatted = utils.FormatGlobalParticipationRateStyle(epoch.VotedEther, epoch.GlobalParticipationRate, currency)
+		epoch.ExecutionReward = big.NewInt(0)
+		epochsMap[epoch.Epoch] = epoch
 	}
 
-	var blocks []*types.IndexPageDataBlocks
-	err = db.ReaderDb.Select(&blocks, `
+	// get slots from db
+	var slots []*types.IndexPageDataBlocks
+	err = db.ReaderDb.Select(&slots, `
```
D13ce, line 662: This SQL statement should be limited to 10 rows as well, should it not?

D13ce:
According to the Figma, the explorer should only show 7 epochs and 7 slots (see "Web 1920 - Landing/Overview"). The current version shows 10 each though. What is correct?

utils/format.go
```diff
@@ -332,6 +332,30 @@ func FormatBlockStatusShort(status uint64) template.HTML {
...
+		return `<div title="This block has been orphaned" data-toggle="tooltip" class="style-badge style-badge-single-char style-bg-neutral-2"><p class="style-status-tag-text text-white" >O<span class="d-none d-sm-inline">rphaned</span></p></div>`
+	} else {
+		return `<div title="This shouldn't be possible" data-toggle="tooltip" class="style-badge style-bg-neutral-2"><p class="style-status-tag-text text-white" >?</p></div>`
+	}
+}
+
+func FormatEpochStatus(finalized bool, participationRate float64) template.HTML {
+	if finalized {
+		return `<div title="This epoch is finalized" data-toggle="tooltip" class="style-badge style-badge-single-char style-bg-good"><p class="style-status-tag-text text-white" >F<span class="d-none d-sm-inline">inalized</span></p></div>`
+	} else if participationRate > 0.66 && participationRate < 1 {
```
D13ce, line 352: Personally I think this `< 1` is confusing as it excludes 1. Why would a perfect participation rate of 1 not be considered save?
Epochs with a participationRate of exactly 1, to my knowledge, are more than rare but sill. Since this is a workaround for the newest epoch in the db, I suggest switching this to `<= 1` as the newest epoch is not finalized anyways.

This applies to line 648 in services.go (https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1718/files#diff-3b37e5382c8ce5b2244a5256b612ea8e965943bc02ec783669c2e679e69d6760R648). A check for finalized should suffice, should it not?

And I know, this is very minor yet the page would show a (more than unlikely) perfect epoch as not finalized.

services/services.go
```diff
@@ -684,70 +677,118 @@ func getIndexPageData() (*types.IndexPageData, error) {
...
 	}
 
-	blocksMap := make(map[uint64]*types.IndexPageDataBlocks)
-	for _, block := range blocks {
-		if blocksMap[block.Slot] == nil || len(block.BlockRoot) > len(blocksMap[block.Slot].BlockRoot) {
-			blocksMap[block.Slot] = block
+	// keep slots with higher blockroot length
+	slotsMap := make(map[uint64]*types.IndexPageDataBlocks)
+	for _, slot := range slots {
+		if slot.Slot > (slots[0].Slot-10) && (slotsMap[slot.Slot] == nil || len(slot.BlockRoot) > len(slotsMap[slot.Slot].BlockRoot)) {
```
D13ce, line 683: You could `minSlot := slots[0].Slot-10` before the for loop and use this here to save some time.

services/services.go
```diff
@@ -684,70 +677,118 @@ func getIndexPageData() (*types.IndexPageData, error) {
...
+		slot.ProposerFormatted = utils.FormatValidatorWithNameAlt(slot.Proposer, slot.ProposerName)
+		slot.ExecutionRewardFormatted = template.HTML("-")
+		slot.ExecutionRewardRecipient = template.HTML("-")
+		slot.Ts = utils.SlotToTime(slot.Slot)
 
-		if !epochsMap[block.Epoch] {
-			epochs = append(epochs, &types.IndexPageDataEpochs{
-				Epoch:                            block.Epoch,
-				Ts:                               utils.EpochToTime(block.Epoch),
+		if _, containsEpoch := epochsMap[slot.Epoch]; !containsEpoch {
```
D13ce, line 696: Why not stay with `if !epochsMap[slot.Epoch]`?

Furthermore, I don't really understand why epochs have to be added like this. Could you please tell me?

D13ce:
The tooltips for the gas chart sometimes are cut off:
image

services/services.go
LuccaBitfly: As you noted, it may be possible for multiple slots to have the same slot number. Although I haven't seen this happen in practice, the old code did check for that possibility. To avoid potential issues, I didn't want to limit the slotsMap to only 10 rows.

utils/format.go
LuccaBitfly: The participation rate for the most recent epoch in the database is always 1, which is why I check for < 1 instead of <= 1. Checking for <= 1 would incorrectly mark the most recent epoch as safe. I realize that this is a convoluted workaround and I agree that it's not ideal.
Regarding line 648, I believe that the EpochParticipationRate field in the data object represents the participation rate of the most recent safe epoch, which does not necessarily entail said epoch being finalized. I also noticed that the first condition of the if statement is unnecessary, so I plan to remove it.

services/services.go
LuccaBitfly: The slots are retrieved from the database after the epochs, so it's possible that newer slots from a more recent epoch than the most recent one in the epoch query get exported to the database in the interim. In this case, I want to display that epoch on the page since the page now also shows slots of that epoch. That's why I add it to the epochsMap, rather than skipping it when calculating the MEV reward.

services/services.go
```diff
@@ -782,15 +823,55 @@ func getIndexPageData() (*types.IndexPageData, error) {
...
+
+	group := make(map[int64]float64, 0)
+	for i := 0; i < len(history); i++ {
+		_, ok := group[history[i].Ts.Truncate(time.Hour).Unix()]
+		if !ok {
+			group[history[i].Ts.Truncate(time.Hour).Unix()] = float64(history[i].Fast.Int64())
+		} else {
+			group[history[i].Ts.Truncate(time.Hour).Unix()] = (group[history[i].Ts.Truncate(time.Hour).Unix()] + float64(history[i].Fast.Int64())) / 2
+		}
+	}
```
D13ce, line 862: You are using `history[i].Ts.Truncate(time.Hour).Unix()` as index 2-3 times per iteration. It would be better to save the result an re-use it.

services/services.go
```diff
@@ -782,15 +823,55 @@ func getIndexPageData() (*types.IndexPageData, error) {
...
+		return nil, fmt.Errorf("error retrieving gas price history: %v", err)
+	}
+
+	group := make(map[int64]float64, 0)
+	for i := 0; i < len(history); i++ {
+		_, ok := group[history[i].Ts.Truncate(time.Hour).Unix()]
+		if !ok {
+			group[history[i].Ts.Truncate(time.Hour).Unix()] = float64(history[i].Fast.Int64())
+		} else {
+			group[history[i].Ts.Truncate(time.Hour).Unix()] = (group[history[i].Ts.Truncate(time.Hour).Unix()] + float64(history[i].Fast.Int64())) / 2
```
D13ce, line 860: Is this formula correct? `group` works with hour precision and within `history` many entries for a specific hour may be present.
Let's say we have 3 entries with a gas price of 10, 5 and 1.
1) 10 gets saved as first value
2) (10 + 5) / 2 = 7.5 gets saved
3) (7.5 + 1) / 2 = **4.25** gets saved

But: (10 + 5 + 1) / 3 = **5,333333333**

Shouldn't 5,3 be the desired outcome?

services/services.go
```diff
@@ -756,7 +797,7 @@ func getIndexPageData() (*types.IndexPageData, error) {
 		return nil, fmt.Errorf("error retrieving validator queue count: %v", err)
 	}
 	data.EnteringValidators = queueCount.EnteringValidators
-	data.ExitingValidators = queueCount.ExitingValidators
+	data.ExitingValidators = queueCount.ExitingValidators */
 
 	var epochLowerBound uint64
 	if epochLowerBound = 0; epoch > 1600 {
```
D13ce, line 803: I know this has not been touched by your changes but one could see it in the scope of this task, so:

Why are 1600 epochs checked here? Only the last finalized epoch and the generally last epoch are used. Wouldn't it suffice to, for example, only grab the last 10 epochs and if it does not contain a finalized one only THEN grab the last, let's say, 1600?

I have made a quick benchmark an the duration for the complete `epochHistory` handling would drop from ~950ms to ~220ms.

Thoughts?

D13ce:
On mobile (I am using chrome's developer tools) the gap between the blocks list and the epochs list is missing:
image

D13ce:
Again on mobile, according to the figma, a finalized epoch should be labeled "FIN" but it is labeled "F" instead:
image

D13ce:
According to the figma, the status labels for epochs and blocks (like "Not finalized" or "Proposed") should all have the same dimensions (with the only exception being the epoch's "Not finalized (Safe)" status).
Within the current implementation, the label's dimensions seem always to be "as small as possible based on the text shown":
image

types/templates.go
```diff
@@ -176,16 +180,35 @@ type IndexPageDataEpochs struct {
...
+	ProposerName             string        `db:"name" json:"-"`
+	ExecutionBlockNumber     int           `db:"exec_block_number" json:"exec_block_number"`
+	ExecutionReward          *big.Int      `json:"-"`
+	ExecutionRewardFormatted template.HTML `json:"exec_reward"`
+	ExecutionRewardRecipient template.HTML `json:"exec_reward_recipient"`
+	BlockRoot                []byte        `db:"blockroot" json:"-"`
+}
+
+// OldIndexPageDataBlocks is a struct to hold detail data for the old main web page, which for some reason is used in the validator and epoch handler
+type OldIndexPageDataBlocks struct {
```
D13ce, line 211: I feel like this has to be removed to avoid confusion. `BlocksPageDataBlocks` a few lines below looks like something that should be used instead, does it not?

utils/format.go
```diff
@@ -836,6 +876,27 @@ func FormatValidatorWithName(validator interface{}, name string) template.HTML {
 	}
 }
 
+func FormatValidatorWithNameAlt(validator interface{}, name string) template.HTML {
+	var validatorRead string
+	var validatorLink string
+	switch v := validator.(type) {
+	case []byte:
+		if len(v) > 2 {
+			validatorRead = fmt.Sprintf("0x%x…%x", v[:2], v[len(v)-2:])
```
D13ce, line 885: If `len(v)` is <= 2, validatorRead is not set. This is not intended, is it?

utils/format.go
```diff
@@ -332,6 +332,30 @@ func FormatBlockStatusShort(status uint64) template.HTML {
 	}
 }
 
+func FormatBlockStatusStyle(status uint64) template.HTML {
+	if status == 0 {
+		return `<div title="This slot is still scheduled" data-toggle="tooltip" class="style-badge style-badge-single-char style-bg-neutral-1"><p class="style-status-tag-text" >S<span class="d-none d-sm-inline">cheduled</span></p></div>`
+	} else if status == 1 {
+		return `<div title="This block has been proposed" data-toggle="tooltip" class="style-badge style-badge-single-char style-bg-good"><p class="style-status-tag-text text-white" >P<span class="d-none d-sm-inline">roposed</span></p></div>`
+	} else if status == 2 {
+		return `<div title="This block has been proposed" data-toggle="tooltip" class="style-badge style-badge-single-char style-bg-bad"><p class="style-status-tag-text" >M<span class="d-none d-sm-inline">issed</span></p></div>`
```
D13ce, line 341: The text here ("This block has been proposed") is the same as the one for status == 1. Is that intended? Unfortunately I do not see what status == 1 and status == 2 means here.

templates/index/hero.html
```diff
@@ -1,4 +1,4 @@
-{{ define "hero" }}
+{{ define "hero_old" }}
```
D13ce, line 1: If I am not mistaken, this is not used anywhere and should be deleted. The same applies to `hero_css_old`.

D13ce (changes requested):
Looking good! Everything seems to work properly and the design is almost exactly as requested.
I have found some minor flaws and made some suggestions though.

handlers/index.go
```diff
@@ -6,40 +6,64 @@ import (
...
-}
-
-var indexTemplate = template.Must(template.New("index").Funcs(utils.GetTemplateFuncs()).ParseFS(templates.Files,
-	indexTemplateFiles...,
-))
-
 // Index will return the main "index" page using a go template
 func Index(w http.ResponseWriter, r *http.Request) {
 
+	/* var indexTemplateFiles = []string{
```
D13ce, line 15: Generally I am not a fan of leaving unused code within the sources. I would remove that (and some other unused stuff).

Buttaa:
> The mobile version shows an additional "Ad" label above the ad itself which should not be here based on the figma: image

That's my/philpps bad -- the ad label is set on our ad-server and it's supposed to be there :)

axljoker03 (approved):
😁👍

LuccaBitfly:
The suggested changes (and some other oversights) have either been implemented/fixed or commented on as to the reasoning why it was done that way.

utils/format.go
LuccaBitfly: This function (which I have deleted since) was basically a copy of the one above. I can't tell for sure if this is intended as the code was written by somebody not working for bitfly anymore. Also I doubt whether the validator parameter is ever passed as a byte array. I will leave this be as is for now and probably replace this function over the course of implementing all new designs.

services/services.go
LuccaBitfly: You're right, 1600 epochs is completely overkill. I've implemented changes so that the relevant db columns are already selected and evaluated in the first epoch query and left the original epoch history code as fallback if no finalized slot is found within the first queried epochs.

LuccaBitfly:
missclick, pls don't review

utils/format.go
D13ce: Got it. Thanks for adding the comments => Resolved

utils/format.go
```diff
@@ -334,25 +334,26 @@ func FormatBlockStatusShort(status uint64) template.HTML {
...
 	} else if status == 1 {
-		return `<div title="This block has been proposed" data-toggle="tooltip" class="style-badge style-badge-single-char style-bg-good"><p class="style-status-tag-text text-white" >P<span class="d-none d-sm-inline">roposed</span></p></div>`
+		return `<div title="This block has been proposed" data-toggle="tooltip" class="style-badge style-badge-single-char style-bg-good"><div class="style-status-tag-text text-white" >P<span class="d-none d-sm-inline">roposed</span></div></div>`
 	} else if status == 2 {
-		return `<div title="This block has been proposed" data-toggle="tooltip" class="style-badge style-badge-single-char style-bg-bad"><p class="style-status-tag-text" >M<span class="d-none d-sm-inline">issed</span></p></div>`
+		return `<div title="This blocks proposal has been missed" data-toggle="tooltip" class="style-badge style-badge-single-char style-bg-bad"><div class="style-status-tag-text" >M<span class="d-none d-sm-inline">issed</span></div></div>`
```
D13ce, line 341: Missing the apostrophe, should be "block's". Or change it to "block proposal" which would make sense too imho.

services/services.go
D13ce: Thank you for the explanation and also thank you for adding the comment in the sources => Resolved

services/status.go
```diff
@@ -2,15 +2,11 @@ package services
...
-	"eth2-exporter/db"
-	"eth2-exporter/version"
-	"os"
-	"time"
 )
 
 // Report the status of a particular service, will add current Pid and executable name
 func ReportStatus(name, status string, metadata *json.RawMessage) {
-	pid := os.Getpid()
+	/* pid := os.Getpid()
```
D13ce, line 9: This change does not seem right. Why have you removed the function's body?

templates/index/recentEpochs.html
```diff
@@ -75,7 +34,7 @@ <h3 class="style-card-header d-flex justify-content-between align-items-center m
             <div class="style-paragraph-6 style-grid-header pr-2" role="columnheader">Time</div>
             <div class="style-paragraph-6 style-grid-header pr-2" role="columnheader"><span class="d-none d-sm-flex">Final</span></div>
             <div class="style-paragraph-6 style-grid-header pr-2" role="columnheader">Voted</div>
-            <div class="style-paragraph-6 style-grid-header pr-2" role="columnheader">Fee Recipient Reward</div>
```
D13ce: Why has this been changed? According to the FIGMA, "Fee Recipient Reward" was correct.

utils/eth1.go
```diff
@@ -346,6 +347,110 @@ func formatAmount(amount *big.Int, unit string, digits int, maxPreCommaDigitsBef
...
+	default:
+		displayUnit = " Wei"
+		unitDigits = 0
+	}
+
+	// Initialize preComma and postComma variables to "0"
+	preComma := "0"
+	postComma := "0"
+
+	if amount != nil {
```
D13ce, line 383: Passing nil as amount shows "0.0 ETH" and a tool-tip of "0.0".

Personally I think this function should not work if amount = nil so I would remove this if here and add a
```
if amount == nil {
	logger.Error("nil passed as amount to NewFormat")
	return template.HTML(`NA`)
}
```
at the very beginning of the function. Otherwise, errors may stay undetected and potentially wrong values are shown.

Thoughts?

utils/eth1.go
```diff
@@ -346,6 +347,110 @@ func formatAmount(amount *big.Int, unit string, digits int, maxPreCommaDigitsBef
...
+//   - `amount` : the Wei(!) amount to be formatted.
+//   - `unit` : the unit in which the amount should be formatted. Supported values are ETH and GWei.
+//   - `digits` : the number of digits after the decimal point to be displayed.
+//   - `mobileDigitsHide` : the number of trailing decimal digits to hide on mobile views.
+//   - `maxPreCommaDigitsBeforeTrim` : the maximum number of digits before the decimal point that should be displayed before trimming excess decimal digits. If this is set to 0, no digits are trimmed.
+//
+// **NOTE** Every number is rounded down
+//
+// TODO Replace every formatAmount with this
+func NewFormat(amount *big.Int, unit string, digits int, mobileDigitsHide int, maxPreCommaDigitsBeforeTrim int) template.HTML {
```
D13ce, line 362: This should most likely be named something like **FormatWeiAmount**.

D13ce, line 362: I may have found some unintentional behavior in this function. 

Here is my testcode:
```
myTest := utils.NewFormat(big.NewInt(1234567890123456789), "ETH", 7, 2, 0)
logger.Info(myTest)
myTest = utils.NewFormat(big.NewInt(1234000000000000000), "ETH", 7, 2, 0)
logger.Info(myTest)
```
This will show the following logs:
> msg="<span data-toggle=\"tooltip\" data-placement=\"top\" title=\"1.234567890123456789\">1.23456<span class=\"d-none d-sm-inline\">78</span> ETH</span>"
> msg="<span data-toggle=\"tooltip\" data-placement=\"top\" title=\"1.234\">1.2<span class=\"d-none d-sm-inline\">34</span> ETH</span>"

**mobileDigitsHide** is used to limit the digital points on mobile. But it works differently based on the passed **amount**. If the **amount** results in trailing 0 that are cut off, **mobileDigitsHide** hides different decimal point positions.

In the first test case, the decimal points on positions 6 and 7 have been hidden. This is as I would expect it as I want to show 7 decimal places (see **digits**) but hide 2 on mobile. The result of "1.23456(78)" (78 hidden on mobile) seems perfectly fine.
In the second test case, the decimal points on positions 2 and 3 have been hidden resulting in only 1 decimal place on mobile: "1.2(34)"

A possible solution would be to add trailing zeros but I am not sure how all this should behave.

This whole comment might be super confusing so I suggest we talk about this face2face.

services/services.go
```diff
@@ -637,27 +637,80 @@ func getIndexPageData() (*types.IndexPageData, error) {
...
+			if epochs[i].Finalized {
+				data.CurrentFinalizedEpoch = epochs[i].Epoch
+				data.FinalityDelay = FinalizationDelay()
+				data.AverageBalance = string(utils.FormatBalance(uint64(epochs[i].AverageValidatorBalance), currency))
+				break
+			}
+		}
+
+		data.StakedEther = string(utils.FormatBalance(epochs[len(epochs)-1].EligibleEther, currency))
+		data.ActiveValidators = epochs[len(epochs)-1].ValidatorsCount
```
D13ce, line 669: This code now is duplicated (see code below from lines 684 to 694). I suggest putting this in a not-exported function that takes the epoch slice (I guess you would have to use **IndexPageDataEpochs** for both cases and get rid of **IndexPageEpochHistory**) and the **IndexPageData** as parameters.

services/services.go
D13ce: Love it!

services/services.go
D13ce: Many thanks for the explanation and for adding the comment in the sources.
One more thing though: The slotsMap gets converted into a slice later and some slots may be removed from there, see [HERE](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1718/files/08a6614eee2a63eeb96bd65716101a2de37f2578..a8bf44365f2fe8cce60ba64fc4d107fae40df585#diff-3b37e5382c8ce5b2244a5256b612ea8e965943bc02ec783669c2e679e69d6760R822).

I think it would be better to add missing epochs after the slots have been reduced. Otherwise it MIGHT happen that you add an epoch for a slot that is not shown in the end.

types/templates.go
D13ce: @LuccaBitfly , thoughts?

D13ce (commented):
Good improvements! I have only found minor things that at least require feedback from the developer.

templates/index/index.html
```diff
@@ -207,168 +183,237 @@
...
+    .style-grid:hover::-webkit-scrollbar-thumb {
+      background-color: var(--table-scroll-hover);
     }
   </style>
+
+  {{ template "hero_css" }}
+  {{ template "postGenesis_css" }}
 {{ end }}
 
 {{ define "content" }}
```
D13ce, line 415: I am sorry, I am not sure where to put this conversation, but there is now a big gap between the latest epochs/slots and the footer:
image

LuccaBitfly (commented):
test

types/templates.go
LuccaBitfly: Somehow overlooked this, sorry. You are probably right, but as both the validator and epoch page handler use the old block struct from the index page (which shouldn't have happened in the first place imo) and I don't see changing and testing these handlers within the scope of this task, I won't change this for now.

services/status.go
LuccaBitfly: Temporary workaround until a implementation for BIDS-1329 gets merged into this branch, won't change

templates/index/recentEpochs.html
LuccaBitfly: We wan't to unify all terms for execution layer rewards (including MEV) across the site, so "EL Reward" is used for now. This is prone to change

utils/eth1.go
LuccaBitfly: My goal was to replace the formatAmount function, once every call of formatAmount has been replaced with NewFormat over the course of implementing all new designs I will change the function name

utils/eth1.go
LuccaBitfly: Good suggestion, was implemented without logging an error. I'm not 100% sure if passing a nil amount is always faulty behavior, once more designs have been implemented I will be able to evaluate this better.

utils/eth1.go
LuccaBitfly: After talking with the designer, having a different number of decimal places on desktop and mobile views was not intended (all ETH amounts should display 5 decimal places). Everything related to hiding digits on mobile was removed.

services/services.go
LuccaBitfly: Implemented in new commit

templates/index/index.html
LuccaBitfly: The height of the main content in a view changes based on the height of the browser window, probably to make sure the footer is not visible without scrolling but I'm not sure. Definitely wonky behavior but won't change for now.

handlers/index.go
LuccaBitfly: Removed in new commit

types/templates.go
D13ce: As discussed, we have created a BIDS for this => Resolved

PR #1720, opened 2022-12-20
add ETH.STORE page (BIDS-1040)

LuccaBitfly:


axljoker03 (approved):
thank you

services/services.go
```diff
@@ -518,6 +546,59 @@ func slotVizUpdater(wg *sync.WaitGroup) {
 	}
 }
 
+func getEthStoreStatisticsData() (*types.EthStoreStatistics, error) {
+
```
D13ce, line 550: Empty line, should be removed.

templates/ethstore.html
```diff
@@ -0,0 +1,524 @@
...
+		},
+		],
+		navigator: {
+			series: {
+				type: 'line'
+			}
+		},
+		series: [
+			{
+				name: 'Total Staking Rewards',
```
D13ce, line 496: https://staking.ethermine.org/statistics labels this "Total Staking Rewards (ETH)" i.e. adds "(ETH)" at the end. Should the same be done here?

LuccaBitfly: In my opinion it is unnecessary to specify the unit when every datapoint has "X ETH" as its tooltip, @Buttaa thoughts?

cmd/explorer/main.go
```diff
@@ -54,6 +54,7 @@ func init() {
 }
 
 func main() {
+	logrus.SetLevel(logrus.TraceLevel)
```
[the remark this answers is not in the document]
LuccaBitfly: Correct, will remove

templates/ethstore.html
```diff
@@ -0,0 +1,524 @@
...
+        </div>
+
+        <button type="button" class="collapsible">How are total rewards calculated in ETH.STORE?</button>
+        <div class="content">
+          <p>Total staking rewards consists of consensus rewards and transaction fees:</p>
+          <ol>
+            <li>Consensus rewards: Rewards received by validators for proposing and attesting in accordance with the consensus rules of the Ethereum protocol</li>
+            <li>Transaction fees: Fees received by validators for processing transactions on the Ethereum network</li>
+          </ol>
+          <p>Total staking rewards equals the sum of consensus rewards and transaction fees</p>
```
D13ce, line 256: This point is missing punctations:
- "... rules of the Ethereum protocol" (missing dot at the end)
- "... transactions on the Ethereum network" (missing dot at the end)
- "... rewards and transaction fees" (missing dot at the end)

Again, this matches the behavior over on https://staking.ethermine.org/statistics but I think this should be changed on both sides.
Thoughts?

templates/ethstore.html
```diff
@@ -0,0 +1,524 @@
...
+              <thead>
+                <tr>
+                  <th>Variable</th>
+                  <th>Description</th>
+                </tr>
+              </thead>
+              <tbody>
+                <tr>
+                  <td>Start epoch (SE):</td>
+                  <td>First epoch after 1200 UTC every day</td>
```
D13ce, line 272: Above the notation has been "12:00:23 UTC" but here the ":" is missing in "1200 UTC". Shouldn't this be harmonized?

Again, this matches the behavior over on https://staking.ethermine.org/statistics but I think this should be changed on both sides.

templates/ethstore.html
```diff
@@ -0,0 +1,524 @@
...
+                  <td>Start epoch (SE):</td>
+                  <td>Epoch {{ .Data.StartEpoch }}</td>
+                </tr>
+                <tr>
+                  <td>End epoch (EE):</td>
+                  <td>Epoch {{ addUI64 .Data.StartEpoch 225 }}</td>
+                </tr>
+                <tr>
+                  <td>Active validators (VS):</td>
+                  <td>in timeframe [{{ .Data.StartEpoch }},{{ addUI64 .Data.StartEpoch 225 }})</td>
```
D13ce, line 322: Three times a "[" is combined with a ")". This should be changed to "[" and "]" for all cases, should it not?

Again, this matches the behavior over on https://staking.ethermine.org/statistics but I think this should be changed on both sides.

LuccaBitfly: I believe this is written in mathematical interval notation, as in [ is including that number while ) is excluding that number

templates/ethstore.html
```diff
@@ -0,0 +1,524 @@
...
+                  <td>Sum rewards (RS):</td>
+                  <td>{{ .Data.YesterdayRewards }} Ether</td>
+                </tr>
+                <tr>
+                  <td>Sum effective balances (EB):</td>
+                  <td>{{ .Data.YesterdayEffectiveBalance }} Ether</td>
+                </tr>
+                <tr>
+                  <td>ETH.STORE:</td>
+                  <td>{{ .Data.ProjectedAPR }} % = ({{ .Data.YesterdayRewards }} / ({{ .Data.YesterdayEffectiveBalance }})) * 365</td>
```
D13ce, line 334: This page uses the "e+X" notation while https://staking.ethermine.org/statistics uses "^10". Should this be harmonized?

templates/ethstore.html
```diff
@@ -0,0 +1,524 @@
...
+                </tr>
+              </tbody>
+            </table>
+          </div>
+        </div>
+
+        <button type="button" class="collapsible">Is the ETH.STORE code Open-Source?</button>
+        <div class="content">
+          <p>Yes, you can run the calculations yourself with our open source ETH.STORE tool and verify the integrity of the ETH.STORE listed here. Check out <a href="https://github.com/gobitfly/eth.store/" target="_blank">ETH.STORE on Github</a>.</p>
```
D13ce, line 343: The link to github in https://staking.ethermine.org/statistics is orange while it is blue here. Should this be harmonized?

templates/ethstore.html
```diff
@@ -0,0 +1,524 @@
...
+{{ end }}
+
+{{ define "js" }}
+  <script src="/js/highcharts/highstock.min.js"></script>
+  <script src="/js/highcharts/highcharts-global-options.js"></script>
+  <script>
+	var coll = document.getElementsByClassName("collapsible");
+	var i;
+
+	for (i = 0; i < coll.length; i++) {
```
D13ce, line 359: Why declare i above?

templates/ethstore.html
D13ce: FYI, something similar also applies to the chart's tooltip: The explorer now shows something like "Total Staking Reward: 212 ETH" while https://staking.ethermine.org/statistics shows something like "Total Staking Reward (ETH): 212 ETH" (the latter is worse in my humble opinion).

services/services.go
```diff
@@ -491,6 +494,31 @@ func indexPageDataUpdater(wg *sync.WaitGroup) {
...
+		if err != nil {
+			logger.Errorf("error retrieving ETH.STORE statistics data: %v", err)
+			time.Sleep(time.Second * 10)
+			continue
+		}
+
+		cacheKey := fmt.Sprintf("%d:frontend:ethStoreStatistics", utils.Config.Chain.Config.DepositChainID)
+		err = cache.TieredCache.Set(cacheKey, data, time.Hour*24)
+		if err != nil {
+			logger.Errorf("error caching indexPageData: %v", err)
```
D13ce, line 510: Copy&Paste error ("indexPageData" versus "ethStoreStatistics").

services/services.go
```diff
@@ -904,6 +985,17 @@ func LatestBurnData() *types.BurnPageData {
 	return &types.BurnPageData{}
 }
 
+func LatestEthStoreStatistics() *types.EthStoreStatistics {
+	wanted := &types.EthStoreStatistics{}
+	cacheKey := fmt.Sprintf("%d:frontend:ethStoreStatistics", utils.Config.Chain.Config.DepositChainID)
+	if wanted, err := cache.TieredCache.GetWithLocalTimeout(cacheKey, time.Second*60, wanted); err == nil {
+		return wanted.(*types.EthStoreStatistics)
+	} else {
+		logger.Errorf("error retrieving burn data from cache: %v", err)
```
D13ce, line 994: Copy&Paste error.

services/services.go
```diff
@@ -518,6 +546,59 @@ func slotVizUpdater(wg *sync.WaitGroup) {
...
+	daysLastIndex := len(ethStoreDays) - 1
+
+	effectiveBalances := [][]float64{}
+	totalRewards := [][]float64{}
+	aprs := [][]float64{}
+	for _, stat := range ethStoreDays {
+		ts := utils.EpochToTime(stat.Day * 225)
+
+		effectiveBalances = append(effectiveBalances, []float64{
+			float64(ts.Unix()) * 1000,
```
D13ce, line 573: Minor: `float64(ts.Unix()) * 1000` is done 3 times per loop. You could do it just once and save the result temporarily.

services/services.go
```diff
@@ -904,6 +985,17 @@ func LatestBurnData() *types.BurnPageData {
...
 
+func LatestEthStoreStatistics() *types.EthStoreStatistics {
+	wanted := &types.EthStoreStatistics{}
+	cacheKey := fmt.Sprintf("%d:frontend:ethStoreStatistics", utils.Config.Chain.Config.DepositChainID)
+	if wanted, err := cache.TieredCache.GetWithLocalTimeout(cacheKey, time.Second*60, wanted); err == nil {
+		return wanted.(*types.EthStoreStatistics)
+	} else {
+		logger.Errorf("error retrieving burn data from cache: %v", err)
+	}
+	return wanted
```
D13ce, line 996: Shouldn't this return `&types.EthStoreStatistics{}`?

templates/ethstore.html
LuccaBitfly: Yes, this is exactly what I meant. Every datapoint on the chart already specifies the unit, which is why I deem the "(ETH)" unnecessary. In the same vein, the total effective balance is also denominated in ETH but doesn't specify its unit.

templates/ethstore.html
```diff
@@ -0,0 +1,524 @@
...
+          </div>
+        </div>
+      </section>
+
+      <section class="my-5">
+        <h1 class="title-head">ETH.STORE Explained</h1>
+
+        <button type="button" class="collapsible">What is ETH.STORE?</button>
+        <div class="content">
+          <p>ETH.STORE (Ether <strong>St</strong>aking <strong>O</strong>ffered <strong>R</strong>at<strong>e</strong>) is a transparent Ethereum staking reward reference rate. It represents the average financial return validators on the Ethereum network have achieved in a 24-hour period as published on https://staking.ethermine.org/statistics.</p>
```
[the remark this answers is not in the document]
LuccaBitfly: I did not notice this, URL should probably be changed to the beaconcha.in explorer in any case.

templates/ethstore.html
D13ce: I totally understand your reasoning and I am not saying that anything is wrong. It is just different to the EMS.
Maybe @Buttaa can let us know how it should be.

templates/ethstore.html
D13ce: TIL!

Thanks for your explanation => Resolved.

services/services.go
LuccaBitfly: Fixed in new commit

templates/ethstore.html
```diff
@@ -0,0 +1,524 @@
...
+        </div>
+
+        <button type="button" class="collapsible">How can I reproduce the ETH.STORE calculation on my own?</button>
+        <div class="content">
+          <div class="table-responsive">
+            <table class="table">
+              <thead>
+                <tr>
+                  <th>Variable</th>
+                  <th>Description</th>
```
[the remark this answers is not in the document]
LuccaBitfly: No, won't change

templates/ethstore.html
LuccaBitfly: No, won't change

templates/ethstore.html
LuccaBitfly: The entire JS code was copied from another project and barely touched, won't change for now

services/services.go
LuccaBitfly: Fixed in new commit

services/services.go
LuccaBitfly: Fixed in new commit

services/services.go
LuccaBitfly: Fixed in new commit

services/services.go
LuccaBitfly: Don't know if this makes any differences, but fixed to keep in line with similar functions

templates/ethstore.html
LuccaBitfly: After talking to him privately this change has been approved

templates/ethstore.html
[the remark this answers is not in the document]
LuccaBitfly: Fixed in new commit

templates/ethstore.html
LuccaBitfly: Fixed in new commit

templates/ethstore.html
LuccaBitfly: Fixed in new commit

templates/ethstore.html
LuccaBitfly: Fixed in new commit

D13ce: You have missed one: 
image
(see the second number)

Also: Does the change with `formatThousandsFancy` work as intended? The numbers seem way too high.

Here is the data for the 4th of January on Prater with your changes:
image

And here is the data for the 8th of January on Mainnet:
image

Compare, for example, the "Sum rewards (RS)".

LuccaBitfly: Oversight, sorry! Forgot to convert from Wei to ETH, should be fixed in new commit

templates/ethstore.html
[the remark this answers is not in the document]
LuccaBitfly: Yes

D13ce (approved):
lgtm

PR #1729, opened 2023-01-06
(BIDS-1334) Concurrent map iteration and map write - bigtable_eth1.go…

recy21:
…#L213

### DIFF
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -2123,31 +2123,39 @@ func (bigtable *Bigtable) GetInternalTransfersForTransaction(transaction []byte,
 	}
 
 	names := make(map[string]string)
-	// init
-	for _, t := range transfers {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
-	}
+	{
+		forNames := make(map[string]string)
+		for _, t := range transfers {
+			forNames[string(t.From)] = ""
+			forNames[string(t.To)] = ""
+		}
 
-	g := new(errgroup.Group)
-	g.SetLimit(25)
-	for address := range names {
-		address := address
-		g.Go(func() error {
-			name, err := bigtable.GetAddressName([]byte(address))
-			if err != nil {
-				return err
+		g := new(errgroup.Group)
+		g.SetLimit(25)
+		for address := range forNames {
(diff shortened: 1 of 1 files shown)


db/bigtable_eth1.go
```diff
@@ -2123,31 +2123,39 @@ func (bigtable *Bigtable) GetInternalTransfersForTransaction(transaction []byte,
...
-	for address := range names {
-		address := address
-		g.Go(func() error {
-			name, err := bigtable.GetAddressName([]byte(address))
-			if err != nil {
-				return err
+		g := new(errgroup.Group)
+		g.SetLimit(25)
+		for address := range forNames {
+			address := address
```
D13ce, line 2136: I know this was not changed by you, but the naming of this variable is VERY confusing (as the name is already used).
Feel free to change or to ignore.

db/bigtable_eth1.go
```diff
@@ -2123,31 +2123,39 @@ func (bigtable *Bigtable) GetInternalTransfersForTransaction(transaction []byte,
...
+					mux.Lock()
+					names[address] = name
+					mux.Unlock()
+					return nil
+				})
+			} else {
+				mux.Lock()
+				names[address] = ""
+				mux.Unlock()
 			}
```
D13ce, line 2152: I don't understand this else-block:
Why would you put something at `names[""]`?
Please share your thoughts.

D13ce (commented):
One open question (see https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1729/files?diff=unified&w=0#r1064421795), once this is tackled everything's fine.

db/bigtable_eth1.go
recy21: Yes, I agree, but this seems to be the Golang way of doing it. :man_shrugging:

db/bigtable_eth1.go
recy21: - To keep current / previous behavior
- There will be a map read later in the code for index "", even if names["random-index"] will be "" anyway, I feels more accurate to me.

But sure, the else block can be removed without any issue, I guess

D13ce (commented):
lgtm

PR #1731, opened 2023-01-09
- Fix race condition (BIDS-1344)

D13ce:


db/db.go
```diff
@@ -39,32 +39,45 @@ var logger = logrus.StandardLogger().WithField("module", "db")
...
+			case <-endTimeoutDetection:
+				return
+			}
+			time.Sleep(10 * time.Millisecond)
+		}
 	}()
-	err = dbConnWriter.Ping()
+	err := dbConn.Ping()
 	if err != nil {
-		logger.Fatal(err)
```
D13ce, line 57: This is an intentional behavior change. A db that cannot be pinged most likely will not work so why should we not quit here?

PR #1747, opened 2023-01-23
Ethstore API improvement/bugfix (BIDS-1443)

D13ce:
* Break down EL and CL rewards
* Fix timeframe bug for AVG values (BIDS-1443)


manuelsc:
lgtm

PR #1749, opened 2023-01-24
(BIDS-1453) Some epoch errors which should be avoid / switched to war…

recy21:
…nings

### DIFF
--- a/handlers/epoch.go
+++ b/handlers/epoch.go
@@ -36,7 +36,6 @@ func Epoch(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		SetPageDataTitle(data, fmt.Sprintf("Epoch %v", epochString))
 		data.Meta.Path = "/epoch/" + epochString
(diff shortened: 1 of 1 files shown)


handlers/epoch.go
```diff
@@ -156,10 +155,15 @@ func Epoch(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "Internal server error", http.StatusInternalServerError)
 		return
 	}
-	err = db.ReaderDb.Get(&epochPageData.PreviousEpoch, "SELECT epoch FROM epochs WHERE epoch < $1 ORDER BY epoch DESC LIMIT 1", epochPageData.Epoch)
-	if err != nil {
-		logger.Errorf("error retrieving previous epoch for epoch %v: %v", epochPageData.Epoch, err)
+
+	if epochPageData.Epoch == 0 {
```
guybrush, line 159: `if epochPageData.Epoch <= 1`

previous epoch is also 0 if the current epoch is 1

PR #1761, opened 2023-01-26
(BIDS-1486) Fix for Sepolia: error collection notifications: error co…

recy21:
…llecting rocketpool commission: sql: no rows in result set


manuelsc (approved):
lgtm

PR #1766, opened 2023-01-30
BIDS-1495/BetterHandlingForFirebaseErrors

recy21:


notify/firebase.go
```diff
@@ -13,42 +14,80 @@ import (
...
+		resultFailureCount = result.FailureCount
+
+		newMessages := make([]*messaging.Message, 0, resultFailureCount)
+		if resultFailureCount > 0 {
+			for i, response := range result.Responses {
+				if isRelevantError(response) {
+					newMessages = append(newMessages, currentMessages[i])
+				}
+			}
+		}
```
D13ce, line 75: This causes weird behavior in the following scenario:

1. You have 3 messages: **A**, **B** and **C**.
2. In your first try you successfully send **A**, **B** triggers a relevant error and **C** triggers an irrelevant error.
3. In your second try you only resend **B** and it triggers a relevant error again.
4. In your third try you resend **B** again and it works.

`resultSuccessCount` will be 2 and `resultFailureCount` will be 0 even though 3 messages were sent.

recy21: fixed

notify/firebase.go
```diff
@@ -13,42 +14,82 @@ import (
...
+		if err != nil {
+			logger.Errorf("error sending push notifications: %v", err)
+			return err
+		}
+
+		resultSuccessCount += result.SuccessCount
+		resultFailureCount += result.FailureCount
+
+		newMessages := make([]*messaging.Message, 0, resultFailureCount)
+		if resultFailureCount > 0 {
```
manuelsc, line 69: Super unimportant but using `result.FailureCount` instead of `resultFailureCount` would make the code a bit easier to read since you don't have to look ahead for the decrement.

manuelsc (approved):
otherwise lgtm

PR #1779, opened 2023-02-02
(BIDS-1521) Fix underflow errors in bigtable for epoch 0

D13ce:
* Use empty end range to include epoch 0 when required in GetValidatorBalanceStatistics
* Only delete duplicated info for epoch >= 3 in SaveEpoch
* Prevent infinite loop for day 0 when calling GetValidatorMissedAttestationsCount


db/bigtable.go
```diff
@@ -938,6 +938,11 @@ func (bigtable *Bigtable) GetValidatorMissedAttestationsCount(validators []uint6
 				}
 			}
 		}
+
```
recy21, line 941: I don't like this for at all (including your fix)

If startEpoch is 100.000 and endEpoch is 99.999, you are checking from 100.000 down to 1. In that case, endEpoch is not endEpoch, it's the amount of Epochs to be checked?

db/bigtable.go
```diff
@@ -1016,9 +1021,11 @@ func (bigtable *Bigtable) GetValidatorBalanceStatistics(startEpoch, endEpoch uin
 	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Minute*10))
 	defer cancel()
 
-	// logger.Info(startEpoch, endEpoch)
 	rangeStart := fmt.Sprintf("%s:e:b:%s", bigtable.chainId, reversedPaddedEpoch(endEpoch)) // Reverse as keys are sorted in descending order
-	rangeEnd := fmt.Sprintf("%s:e:b:%s", bigtable.chainId, reversedPaddedEpoch(startEpoch-1))
+	rangeEnd := ""
```
recy21, line 1025: `var rangeEnd string` will have the same result

D13ce: Sure, changed. Will keep this in mind for my further changes :)

db/bigtable.go
D13ce: Changed.

db/bigtable.go
```diff
@@ -856,7 +862,12 @@ func (bigtable *Bigtable) GetValidatorSyncDutiesHistory(validators []uint64, sta
 	defer cancel()
 
 	rangeStart := fmt.Sprintf("%s:e:%s:s:", bigtable.chainId, reversedPaddedEpoch(startEpoch))
-	rangeEnd := fmt.Sprintf("%s:e:%s:s:", bigtable.chainId, reversedPaddedEpoch(startEpoch-uint64(limit)))
+	var rangeEnd string
+	if startEpoch >= uint64(limit) {
+		rangeEnd = fmt.Sprintf("%s:e:%s:s:", bigtable.chainId, reversedPaddedEpoch(startEpoch-uint64(limit)))
+	} else {
+		rangeEnd = fmt.Sprintf("%s:e:%s:s:%s", bigtable.chainId, reversedPaddedEpoch(0), "\x00") // add \x00 to rangeEnd so epoch 0 is included
```
qu0b, line 869: From my understanding this will always return an empty array as the rangeEnd will be smaller than the rangeStart. I don't think a query where limit > startEpoch makes sense and I would suggest to return an error instead. Especially, considering that we currently cannot query for epoch 0.

qu0b: It might be a good Idea to return an error here instead of running a query that will always return a zero length array.

recy21 (approved):

PR #1790, opened 2023-02-06
(BIDS-1196) Add copy button for 0x00 withdrawals too

D13ce:


utils/format.go
```diff
@@ -508,31 +508,34 @@ func FormatHash(hash []byte, trunc_opt ...bool) template.HTML {
 	return template.HTML(fmt.Sprintf("<span class=\"text-monospace\">%#x</span>", hash))
 }
 
-func formatWithdrawalHash(hash []byte) template.HTML {
+func formatWithdrawalHash(hash []byte) string {
```
recy21, line 511: Why did you switched to a return value of "string", when you return html & HTML is just an alias of string?

D13ce: The whole method now returns a string that is used within `FormatWithdawalCredentials` below to construct the `text` variable. The text variable was a string initially.

The final return value from `FormatWithdawalCredentials` then is HTML again.

recy21 (approved):

PR #1795, opened 2023-02-07
(BIDS-1095) Remove Open Ethereum from client list

beaconchain:


D13ce (changes requested):
Searching the sources for "openethereum" shows a lot of hits. Please check what can be removed.

You may want to take a look at this BIDS as an example (in the other direction): https://bitfly1.atlassian.net/browse/BIDS-894 (the according commit is linked in the comments)

D13ce (approved):
lgtm

PR #1799, opened 2023-02-08
(BIDS-1548) Eth Data Block Hash broken

Eisei24:
Keep in mind that the missing "0x" prefix happens only on the mainnet. See the missing '#' here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/524b86ffe4ec17f50e0de4e7ab1eb0152a38ec58/utils/format.go#L420


D13ce (commented):
lgtm

Also thanks for cleaning up.

D13ce (approved):

PR #1800, opened 2023-02-09
BIDS-1550/calldata copy button

Eisei24:
*CallData* is already a string (see *Eth1TxData struct*) so it can't be shown with "%x".

### DIFF
--- a/handlers/eth1tx.go
+++ b/handlers/eth1tx.go
@@ -27,12 +27,11 @@ func Eth1TransactionTx(w http.ResponseWriter, r *http.Request) {
 	data := InitPageData(w, r, "blockchain", "/tx", "Transaction")
 	data.HeaderAd = true
 
+	SetPageDataTitle(data, fmt.Sprintf("Transaction %v", txHashString))
+	data.Meta.Path = "/tx/" + txHashString
+
 	txHash, err := hex.DecodeString(strings.ReplaceAll(txHashString, "0x", ""))
-
 	if err != nil {
-
-		SetPageDataTitle(data, fmt.Sprintf("Transaction %v", txHashString))
-		data.Meta.Path = "/tx/" + txHashString
 		logger.Errorf("error parsing tx hash %v: %v", txHashString, err)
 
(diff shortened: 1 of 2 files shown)


D13ce (changes requested):
While this is working for the a transaction's overview tab ([EXAMPLE](http://localhost:8081/tx/0x479666bd6f179cbb2576442fbcf0b15f44a0afd891d351aa14239442efd6436d)), it does not work for the transaction's events tab ([EXAMPLE](http://localhost:8081/tx/0x479666bd6f179cbb2576442fbcf0b15f44a0afd891d351aa14239442efd6436d#events)).

Furthermore, I have searched for the string "copy public key" in the sources and found another occurrence that might be wrong:
[Transaction Hash](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/c6789e88cfff5b6a2773d66b28548b6013cabf74/templates/eth1tx.html#L55): Please check if that has to be changed to something like "Copy transaction hash to clipboard".

handlers/eth1tx.go
```diff
@@ -22,7 +22,7 @@ func Eth1TransactionTx(w http.ResponseWriter, r *http.Request) {
```
Eisei24, line 22: The changes in this method don't relate to the task but while I had a look at the transaction view I noticed that some updates could be done here.

D13ce: This changes the behavior oh so slightly:
If an error is thrown, the first if caused a page title of "Transaction 123" while the second if caused a page title of "Transaction 0x123"
With your changes, both ifs write "Transaction 0x123".

To me, that's an improvement. But was that an intentional change by you?

And of course I am curious to know why you have extracted the `SetPageDataTitle` and the setting of `data.Meta.Path` out of the second if but not out of the first if.
Wouldn't it work to set both right above the first if?

Eisei24: You are right, moved it up.
he important behaviour change is that *data.Meta.Path* now always contains the "0x" while before in the error cases it didn't.

D13ce (approved):
lgtm

PR #1810, opened 2023-02-14
(BIDS-1158) Prevent scam urls as token symbol

Eisei24:
I found 4 places where the symbol of the token gets shown:

-  The address page top right where the list of all the tokens for this address is, see e.g. /address/0x78359e7df948834cafecbe0494b010c6f7f7fa74#erc20Txns
- The address page "Erc20 Token Txns" tab where the list of all the token transactions is, see e.g. same page as before
- The token page, see e.g. /token/0x620bf7f257055c14b6ef1826b2e913cbe7efa886?a=0x78359e7df948834cafecbe0494b010c6f7f7fa74
- Some transaction pages where tokens were transferred, see e.g. /tx/0xff34c1afe4757d3e0ba4ed3eceb7738b48160192551ecf00ec87daa100cb8db3

To test it I hardcoded something like "880$ Visit https://agbonus.site to claim reward." as the symbol.

### DIFF
--- a/templates/execution/token.html
+++ b/templates/execution/token.html
@@ -245,7 +245,7 @@
       <h1 class="h4 mb-1 mb-md-0 header-token text-truncate">
         <div class="my-md-3 font-weight-bold">
           <span class="d-flex">
-            <span class="mr-1 d-flex align-items-center" title={{ .Data.Metadata.Symbol | formatTokenSymbolTitle }}>
+            <span class="mr-1 d-flex align-items-center" title="{{ .Data.Metadata.Symbol | formatTokenSymbolTitle }}">
               {{ if .Data.Metadata.Logo }}
                 <img class="mr-1" style="height: 1.5rem;" src="data:image/png;base64, {{ toBase64 .Data.Metadata.Logo }}" />
               {{ else }}
--- a/templates/execution/token.html
+++ b/templates/execution/token.html
@@ -245,14 +245,14 @@
       <h1 class="h4 mb-1 mb-md-0 header-token text-truncate">
         <div class="my-md-3 font-weight-bold">
           <span class="d-flex">
-            <span class="mr-1 d-flex align-items-center">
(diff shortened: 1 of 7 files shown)


recy21 (approved):

templates/execution/token.html
```diff
@@ -245,14 +245,14 @@
       <h1 class="h4 mb-1 mb-md-0 header-token text-truncate">
         <div class="my-md-3 font-weight-bold">
           <span class="d-flex">
-            <span class="mr-1 d-flex align-items-center">
+            <span class="mr-1 d-flex align-items-center" title="{{ .Data.Metadata.Symbol | formatTokenSymbolTitle }}">
```
D13ce, line 248: There is no need to use the pipe symbol '|' here. Instead, writing
`{{ formatTokenSymbolTitle .Data.Metadata.Symbol }}`
would achieve the same but be more readable imho.

This applies to more of your changes.

D13ce (changes requested):
As discussed, we have to adapt the handling slightly (see comments in Jira).

Regarding your uses of the pipe in the Go templates, I guess it is a matter of taste. But please reply to my comment what you think.

templates/execution/token.html
Eisei24: I slightly prefer the pipeline because if I see the '|' character I immediately know that a function is getting called.
It's already used like this with only one parameter and one function in some files but if you want it changed we can talk about it.

Eisei24: We now also want the replacement text to be [hidden-symbol].
For this review please also keep in mind that I show the tooltip also on mouseover of the logo (if it has one).
I think this makes sense as both together form a href, see e.g. here:
https://beaconcha.in/address/0xa4580ea641b98a25f3549e2712415028f8178824#erc20Txns

D13ce (approved):
utACK

Testing this on Prater is hard as I was unable to find any potential spam token. I hardcoded a few parts to emulate the behavior and everything worked as expected.

PR #1814, opened 2023-02-15
(BIDS-68) fix validator sync participation

LuccaBitfly:


handlers/validator.go
```diff
@@ -1811,10 +1827,15 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
...
-		syncPeriods = syncPeriods[1:]
+
+	//remove scheduled committees
+	for _, syncPeriod := range syncPeriods {
+		if syncPeriod.EndEpoch > latestEpoch {
+			syncPeriods = syncPeriods[1:]
+		} else {
+			break
+		}
 	}
```
recy21, line 1839:
```
for i, syncPeriod := range syncPeriods {
	if syncPeriod.EndEpoch <= latestEpoch {
		syncPeriods = syncPeriods[i:]
		break
	}
}
```

recy21: Maybe faster?

handlers/validator.go
```diff
@@ -626,10 +628,21 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
+
+	// remove scheduled committees
+	latestEpoch := services.LatestEpoch()
+	for _, syncPeriod := range syncPeriods {
+		if syncPeriod.FirstEpoch > latestEpoch {
+			syncPeriods = syncPeriods[1:]
+		} else {
+			break
+		}
+	}
```
recy21, line 640:
```
for i, syncPeriod := range syncPeriods {
	if syncPeriod.FirstEpoch <= latestEpoch {
		syncPeriods = syncPeriods[i:]
		break
	}
}
```

recy21: maybe faster?

recy21 (changes requested):

handlers/validator.go
LuccaBitfly: Is it guaranteed that no more than 1 sync committee is scheduled? Otherwise you could also just do
`if syncPeriods[0].EndEpoch > latestEpoch {
	syncPeriods = syncPeriods[1:]
}`

handlers/validator.go
LuccaBitfly: See my comment in conversation above

handlers/validator.go
recy21: No idea, but based on the original code, I guess no guarantee at all :S

handlers/validator.go
```diff
@@ -1811,10 +1827,15 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	//remove all future sync periods
 	latestEpoch := services.LatestEpoch()
-	for syncPeriods[0].EndEpoch > services.LatestEpoch() {
-		syncPeriods = syncPeriods[1:]
+
+	//remove scheduled committees
+	for _, syncPeriod := range syncPeriods {
```
D13ce, line 1833: Because of this change, the block starting at line 1842 is not required anymore, is it?

D13ce (changes requested):
Please change line 642 (see comment there).

Additionally, I am not sure if everything works correctly. Here the easiest possibly broken edge case I can think of (all line numbers reference `validator.go`):

Have a validator that has never had any sync duties but is part of the scheduled sync committee.
`syncPeriods` (line 615) will not contain anything after the postgres quey (line 621) once scheduled sync committees are removed (line 634).
That way, a lot of `validatorPageData` fields (line 688) will never be filled because of line 644.
But bigtable (line 665) would contain data about the scheduled sync committee, would it not?

Since stuff like this is quite hard to communicate over Github comments, feel free to hit me up in Discord or even better, directly in the office.

handlers/validator.go
```diff
@@ -626,10 +628,21 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
+	latestEpoch := services.LatestEpoch()
+	for _, syncPeriod := range syncPeriods {
+		if syncPeriod.FirstEpoch > latestEpoch {
+			syncPeriods = syncPeriods[1:]
+		} else {
+			break
+		}
+	}
+	// expected count of sync duties, this is set later to the actual value
 	validatorPageData.SyncCount = uint64(len(syncPeriods)) * utils.Config.Chain.Config.EpochsPerSyncCommitteePeriod * utils.Config.Chain.Config.SlotsPerEpoch
```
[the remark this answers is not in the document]
LuccaBitfly: Fixed in [new commit](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1814/commits/5614a705f1f4b6e76a79aa6b9022d89de0f43770)

LuccaBitfly (commented):
testest

LuccaBitfly (commented):
asfsadffsdfas

handlers/validator.go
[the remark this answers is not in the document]
D13ce: Thanks you.

handlers/validator.go
D13ce: Duplicated conversation, we will resolve this one and remember it in the first one :)

handlers/validator.go
D13ce: Let us keep in mind that the same exact issue appears twice: Once here and once in line 634.

handlers/validator.go
LuccaBitfly: It is required, can give you explanation after lunch lol

LuccaBitfly: So basically the block starting at line 1842 does a completely different thing. `StartEpoch` in that code block refers to the epoch the sync table first shows, which per default is that sync periods chronologically **latest** epoch due to the descending order. If that sync periods latest epoch has yet to happen, the `StartEpoch` is set to the current epoch, so that the current epoch is shown on the first page in the table, i.e. the table "starts" with that epoch.

LuccaBitfly: This is also the reason why I use the terminology `FirstEpoch` and `LastEpoch` in the validator sync participation, rather than `StartEpoch` and `EndEpoch`. Depending on the ordering of the sync table, `StartEpoch` could be the first or the last epoch.

handlers/validator.go
LuccaBitfly: In order to implement the changes suggested in [my comment](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1814#issuecomment-1436570717) I will keep the code as is and not implement your suggested changes.

D13ce (approved):
utACK

handlers/validator.go
LuccaBitfly: Nvm, did not read through @recy21 's suggestion correctly, implemented in [new commit](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1814/commits/ecedce684e794aa20496267997b91decca420ffd)

recy21 (approved):

PR #1821, opened 2023-02-16
(BIDS-1569) Fixing incorrect proposer formating

Eisei24:


handlers/api.go
```diff
@@ -1205,7 +1205,29 @@ func getRocketpoolValidators(queryIndices []uint64) ([]interface{}, error) {
 }
```
Eisei24, line 1205: I just did some formating on these 2 queries because you cannot read it if it is just one line.
No functionality change.

utils/utils.go
```diff
@@ -943,5 +943,6 @@ func GetTimeToNextWithdrawal(distance uint64) time.Time {
 }
 
 func EpochsPerDay() uint64 {
```
Eisei24, line 945: Just some cleanup.
Not related to the task and no functionality change.

D13ce (approved):
lgtm

I have added a lengthy comment that can just seen as info.

PR #1831, opened 2023-02-20
BIDS-1608/fix_skip_epoch

peterbitfly:


cmd/rewards-exporter/main.go
```diff
@@ -82,29 +82,35 @@ func main() {
...
 			}
 
-			*epoch = int64(head.FinalizedEpoch)
+			if *epoch == -1 {
+				*epoch = int64(head.FinalizedEpoch) - 1
+			}
 
-			start := time.Now()
-			logrus.Infof("retrieving rewards details for epoch %v", *epoch)
+			for i := *epoch + 1; i <= int64(head.FinalizedEpoch); i++ {
```
D13ce, line 95: I don't understand this for loop. If I read this correctly, this for loop will ALWAYS only be executed once (for `i = head.FinalizedEpoch`):

We only get here if the exporter is started with `*epoch` being set to -1. And let's say that the current `head.FinalizedEpoch` is 100.

So in the first iteration of the endless loop (line 79), `*epoch` will be be set to 99 (see line 92).
The for loop therefore goes from 100 (`*epoch + 1`) to `<= 100`: It will be executed exactly once for epoch 100 which is `head.FinalizedEpoch`.

Once that is done, `*epoch` is set to 100 (line 113) and the exporter waits until `head.FinalizedEpoch` is set to 101 (line 80 and if in line 84).

The new for loop therefore now starts with `i` being set to 101 (remember, `*epoch` is still 100) and loops exactly once again. This behavior then continues from here.

If all that is correct, I suggest to remove the for loop.

cmd/rewards-exporter/main.go
```diff
@@ -82,29 +82,35 @@ func main() {
...
+					logrus.Infof("retrieved %v reward details for epoch %v in %v", len(rewards), i, time.Since(start))
+				}
 
-			err = bt.SaveValidatorIncomeDetails(uint64(*epoch), rewards)
-			if err != nil {
-				logrus.Fatalf("error saving reward details to bigtable: %v", err)
+				err = bt.SaveValidatorIncomeDetails(uint64(i), rewards)
+				if err != nil {
+					logrus.Fatalf("error saving reward details to bigtable: %v", err)
+				}
```
D13ce, line 110: This whole block is duplicated code, see block starting at line 117.
I suggest to move this to a function.

D13ce (changes requested):
I don't understand the change and have left a comment explaining my thoughts.

I also suggest to remove duplicated code (not really related to your changes but since the code is touched anyways, this would be cleaner) within this file.

cmd/rewards-exporter/main.go
D13ce: As discussed, `head.FinalizedEpoch` might be incremented by more than 1 and the for loop ensures that nothing is missed. I will resolve this conversation.

recy21 (approved):

D13ce (approved):
lgtm

PR #1833, opened 2023-02-20
BIDS 1211/ui fixes

qu0b:
- (BIDS-1211) add withdrawals to epochs data handler
- (BIDS-1211) add finalized epoch to pagdata
- (BIDS-1211) show text for non-finalized bls changes
- (BIDS-1211) don't show validator history if validator has exited
- (BIDS-1211) don't show validator history if validator has exited
- (BIDS-1211) (NOBIDS) add missing index
- (BIDS-1211) (BIDS-1363) add withdrawals and bls api endpoints
- (BIDS-1211) add withdrawals amount to api daily stats
- (BIDS-1211) (BIDS-1603) don't show withdrawal estimation before capella


handlers/api.go
```diff
@@ -1620,6 +1622,135 @@ func ApiValidatorIncomeDetailsHistory(w http.ResponseWriter, r *http.Request) {
...
+
+	q := r.URL.Query()
+
+	latestEpoch := services.LatestEpoch()
+	epoch, err := strconv.ParseUint(q.Get("epoch"), 10, 64)
+	if err != nil {
+		epoch = latestEpoch
+	}
+
+	endEpoch := epoch - 100
```
D13ce, line 1658: Out of interest, why are only 100 epochs checked? I guess, this is because of performance reasons.

One minor detail though: We don't only check the last 100 epochs as stated via the swagger comments above but the last 101. The start and end epochs are both inclusive [HERE](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1833/files#diff-30c25f2a7d8ae335d2aada4b0a6013ad9d159595bb6a2a14b1d0022451b6753fR2615).

Thoughts?

handlers/validator.go
```diff
@@ -359,58 +360,60 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
-	// get validator withdrawals
-	withdrawalsCount, err := db.GetValidatorWithdrawalsCount(validatorPageData.Index)
-	if err != nil {
-		logger.Errorf("error getting validator withdrawals count from db: %v", err)
-		http.Error(w, "Internal server error", http.StatusInternalServerError)
-		return
-	}
-	validatorPageData.WithdrawalCount = withdrawalsCount
-
-	validatorPageData.ShowWithdrawalWarning = hasMultipleWithdrawalCredentials(validatorPageData.Deposits)
```
D13ce, line 372: At least the `ShowWithdrawalWarning` information has to be shown even pre-Capella. Please move it outside of this if.

`IsWithdrawableAddress` on the other hand is a bit more tricky. It should be set to true if the validator has 0x01 creds even pre-Capella, right? This would hide the following warning:
image
On the other hand, the Withdrawals tab would become active even pre-Capella which would not be perfect.

Thoughts?

**Sidenote:** Now that withdrawals are supported by the explorer, the variable name `ShowWithdrawalWarning` is too vague and I suggest to change it to something like `ShowMultipleWithdrawalCredentialsWarning`.

handlers/api.go
```diff
@@ -1620,6 +1622,135 @@ func ApiValidatorIncomeDetailsHistory(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+}
+
+// ApiValidatorBlsChange godoc
+// @Summary Gets the BLS withdrawal address change for up to 100 validators
+// @Tags Validator
+// @Produce  json
+// @Param  indexOrPubkey path string true "Up to 100 validator indicesOrPubkeys, comma separated"
+// @Success 200 {object} types.ApiResponse{data=[]types.ApiValidatorBalanceHistoryResponse}
```
D13ce, line 1700: Wrong type for return value, should be `ApiValidatorBlsChangeResponse`.

handlers/api.go
```diff
@@ -1628,7 +1759,7 @@ func ApiValidatorIncomeDetailsHistory(w http.ResponseWriter, r *http.Request) {
 // @Param  latest_epoch query int false "The latest epoch to consider in the query"
 // @Param  offset query int false "Number of items to skip"
 // @Param  limit query int false "Maximum number of items to return, up to 100"
-// @Success 200 {object} types.ApiResponse{data=[]types.ApiValidatorBalanceHistoryResponse}
+// @Success 200 {object} types.ApiResponse{data=[]types.ApiValidatorBlsChangeResponse}
```
D13ce, line 1762: Please revert, see comment for line 1700.

D13ce (commented):
I have found some issues that require changes.

handlers/api.go
qu0b: There is a query parameter epoch with which you can view historical epochs as well. To limit the response size at most 100 epochs are returned. We have 100 validators that are possible in a single query and 100 epochs, which would return in the worst case an array with around 100 results

qu0b: Yes its inclusive and returns 101 results, can be fixed

handlers/api.go
qu0b: Yes please fix

handlers/api.go
qu0b: Seems the two were mixed up :face_exhaling:

handlers/validator.go
qu0b: Ah yes I think I went a bit overboard with this, should move this outside the capella check

qu0b: All the suggestions sound good

D13ce:
Changes will be tackled in a new branch.

PR #1835, opened 2023-02-21
(BIDS-1598) Add graffitiwall coordinate params

Eisei24:


handlers/api.go
```diff
@@ -2147,7 +2148,25 @@ func ApiGraffitiwall(w http.ResponseWriter, r *http.Request) {
...
 	}
-	rows, err := db.ReaderDb.Query("SELECT x, y, color, slot, validator FROM graffitiwall WHERE slot <= $1 AND slot >= $2 ORDER BY slot desc", slotQuery, slotQuery-10000)
+
+	startPxl := uint64(0)
+	endPxl := uint64(999)
+
+	startX := math.MinU64(parseUintWithDefault(r.URL.Query().Get("startx"), startPxl), endPxl)
+	startY := math.MinU64(parseUintWithDefault(r.URL.Query().Get("starty"), startPxl), endPxl)
+	endX := math.MinU64(parseUintWithDefault(r.URL.Query().Get("endx"), endPxl), endPxl)
+	endY := math.MinU64(parseUintWithDefault(r.URL.Query().Get("endy"), endPxl), endPxl)
```
D13ce, line 2158: New parameters have not been added to swagger comment above this function. Please add them. Maybe mention that both values are inclusive (as querying from 50/50 to 50/50 works).

Please think about further required changes for the swagger comment too.

handlers/api.go
```diff
@@ -2147,7 +2148,25 @@ func ApiGraffitiwall(w http.ResponseWriter, r *http.Request) {
 	if slotQuery < 10000 {
 		slotQuery = 10000
 	}
-	rows, err := db.ReaderDb.Query("SELECT x, y, color, slot, validator FROM graffitiwall WHERE slot <= $1 AND slot >= $2 ORDER BY slot desc", slotQuery, slotQuery-10000)
+
+	startPxl := uint64(0)
+	endPxl := uint64(999)
```
D13ce, line 2153: **Minor:** These sould have more meaningful names like `defaultStartPxl` and `defaultEndPxl`.

handlers/api.go
D13ce, line 2158: Please add a sanity check as to ensure that start values <= end values and return an error if the check fails. Think about adding this info to the swagger comment.

D13ce (changes requested):
Swagger comment has not been updated. I have also added some suggestions.

handlers/api.go
Eisei24: Adjusted everything in the swagger comment.

handlers/api.go
Eisei24: I had those at first but didn't quite like them because I also use **endPxl** as an upper boundary but by now I probably like those names more than startPxl/endPxl so I changed it.

handlers/api.go
Eisei24: As discussed we don't run into any issues for invalid areas since the result will just be empty but it is useful return an error with the info that the area is invalid.

Eisei24:
This pull request is now obsolete because https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1865 was created.

PR #1841, opened 2023-02-23
Bids 1627/minor withdrawal improvements

D13ce:


qu0b (changes requested):
The old api endpoints need to stay for backwards compatability

db/db.go
```diff
@@ -2044,8 +2046,9 @@ func GetSlotVizData(latestEpoch uint64) ([]*types.SlotVizEpochs, error) {
 				Epoch:          b.Epoch,
 				Finalized:      b.Finalized,
 				Particicpation: b.Globalparticipationrate,
-				Slots:          [32]*types.SlotVizSlots{},
+				Slots:          []*types.SlotVizSlots{},
 			}
+			r.Slots = make([]*types.SlotVizSlots, utils.Config.Chain.Config.SlotsPerEpoch)
```
qu0b, line 2051: Have you tested this change and made sure there are no unwanted sideffects?

db/db.go
```diff
@@ -2467,12 +2472,12 @@ func GetWithdrawals(query string, length, start uint64, orderBy, orderDir string
 	return withdrawals, nil
 }
 
-func GetTotalAmountWithdrawn() (uint64, uint64, error) {
+func GetTotalAmountWithdrawn() (sum uint64, count uint64, err error) {
```
qu0b, line 2475: Whats the point of naming the return values here, these are extra allocations?

db/db.go
```diff
@@ -2955,15 +2943,13 @@ func GetWithdrawableCountFromCursor(epoch uint64, validatorindex uint64, cursor
...
-		AND
-		validatorindex > $4 AND validatorindex < $3
 		AND (
 			(
 				$3::int > $4 AND validatorindex > $4 AND validatorindex < $3
 			)
 			OR
 			(
-				$3::int < $4 AND (validatorindex > $4 OR validatorindex < $3)
+				$3::int < $4 AND (validatorindex < $4 AND validatorindex > $3)
```
qu0b, line 2952: please revert

qu0b (changes requested):

db/db.go
D13ce: I have tested this change with a local Prater instance by simply calling `GetSlotVizData` at the end of the explorer's main function.
Everything seemed to work correctly.

db/db.go
D13ce: Only readability improvements like this:
image

The two extra allocations imho have too little impact to neglect any readability improvement.
If you see this otherwise, I will revert this change.

db/db.go
D13ce: As discussed, my bad. Reverted.

D13ce:
> The old api endpoints need to stay for backwards compatability

No endpoint should be removed. Personally I think that that many `// deprecated` comments for single lines within up2date code is confusing. That's why I have grouped all deprecated block endpoints.

db/db.go
qu0b: For future reference you can add a comment as well to document the function with additional information that will show up on hover.

qu0b (approved):

PR #1842, opened 2023-02-23
(BIDS-1595) Pending withdrawal credentials

MauserBitfly:


recy21 (approved):

utils/format.go
```diff
@@ -525,6 +525,18 @@ func FormatWithdawalCredentials(hash []byte, addCopyButton bool) template.HTML {
...
+
+	var text template.HTML = template.HTML(fmt.Sprintf("<a href=\"/address/%#x\"><span class=\"text-monospace text-success\">0x01</span><span class=\"text-monospace\">00…%x</span></a>", address, address[len(address)-2:]))
+
+	if addCopyButton {
+		text += template.HTML(fmt.Sprintf("<i class=\"fa fa-copy text-muted p-1\" role=\"button\" data-toggle=\"tooltip\" title=\"Copy to clipboard\" data-clipboard-text=\"%s\"></i>", combined))
+	}
+
+	return text
+}
+
```
D13ce, line 539: This can be replaced by something like this:

```
func FormatAddressToWithdawalCredentials(address []byte, addCopyButton bool) template.HTML {
	credentials, err := hex.DecodeString("010000000000000000000000")
	if err != nil {
		return "INVALID CREDENTIALS"
	}
	credentials = append(credentials, address...)

	return FormatWithdawalCredentials(credentials, addCopyButton)
}
```

That way, credential formatting logic like the coloring (`text-success` and `text-warning`) is not duplicated.

Please change and test.

D13ce (changes requested):
Please adapt `FormatAddressToWithdawalCredentials` as described to avoid code duplication.

utils/format.go
MauserBitfly: Thanks, works great!

D13ce (approved):
lgtm

I have found a typo though which I will fix. Please review my small incoming commit and the merge on your own.

MauserBitfly:
> lgtm
> 
> I have found a typo though which I will fix. Please review my small incoming commit and the merge on your own.

thanks looks good

PR #1847, opened 2023-02-23
(BIDS-1297) Replace logrus/logger.Error/Fatal

Eisei24:


exporter/exporter.go
```diff
@@ -142,17 +142,17 @@ func Start(client rpc.Client) error {
...
 		head, err := client.GetChainHead()
 		if err != nil {
-			logger.Fatal(err)
+			utils.LogError("getting chain head from client error", err).Fatal()
 		}
 
 		dbBlocks, err := db.GetLastPendingAndProposedBlocks(1, head.HeadEpoch)
 		if err != nil {
-			logger.Fatal(err)
+			utils.LogError("getting proposed and pending blocks error ", err).Fatal()
```
D13ce, line 150: This very example made me think: We might want to log more information in scenarios like this. Adding the `head.HeadEpoch` information could help.

You could do that by using something like fmt.sprintf and add it as string into the `infoIdentifier`.

@recy21 had a different suggestion though:
Maybe use the `Variadic Functions` feature, see here: https://gobyexample.com/variadic-functions

What are your thoughts?

D13ce (commented):
I have left a few suggestions. Feel free to implement them or close the conversations. As of now, I will neither approve this PR nor request changes.

And here is something else to mind: iirc the whole `runtime.Caller(1)` call can be considered to be super slow. Have you measured the performance?
If it is slow we might not want to show the callstack for errors but only for fatals.

And it should most likely not be used within the API, see here: https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1847/files#diff-6a59605178bab56012cab99c23c5c9de613ff8ecc0af6a1baf5d010c757dda31R2101

We most likely should discuss this topic with @recy21 .

utils/utils.go
```diff
@@ -1044,3 +1046,27 @@ func ForkVersionAtEpoch(epoch uint64) *types.ForkVersion {
...
+		lineIdentifier = msg
+	}
+
+	return logrus.WithFields(logrus.Fields{
+		"file":       fileIdentifier,
+		"line":       lineIdentifier,
+		"function":   functionIdentifier,
+		"info":       infoIdentifier,
+		"error type": fmt.Sprintf("%T", err),
+	}).WithError(err)
```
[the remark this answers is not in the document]
recy21: The output is kinda strange, as normally the "info" is part of Error("..."), so there is now a big empty space instead.

utils/utils.go
```diff
@@ -1044,3 +1046,27 @@ func ForkVersionAtEpoch(epoch uint64) *types.ForkVersion {
 		PreviousVersion: MustParseHex(Config.Chain.Config.GenesisForkVersion),
 	}
 }
+
+func LogError(infoIdentifier string, err error) *logrus.Entry {
+	var fileIdentifier, lineIdentifier, functionIdentifier string
+
+	pc, fullFilePath, line, ok := runtime.Caller(1)
```
recy21, line 1053: There should be an option to use Caller(2) as well, I need this feature in case of an uncaught panic, where Caller(1) will be panic.go, which is not a good info at all.

cmd/explorer/main.go
```diff
@@ -148,7 +151,7 @@ func main() {
 		}
 
 		if !(erigonChainId.String() == gethChainId.String() && erigonChainId.String() == fmt.Sprintf("%d", utils.Config.Chain.Config.DepositChainID)) {
-			logrus.Fatalf("chain id missmatch: erigon chain id %v, geth chain id %v, requested chain id %v", erigonChainId.String(), erigonChainId.String(), fmt.Sprintf("%d", utils.Config.Chain.Config.DepositChainID))
+			logrus.Fatalf("chain id mismatch: erigon chain id %v, geth chain id %v, requested chain id %v", erigonChainId.String(), erigonChainId.String(), fmt.Sprintf("%d", utils.Config.Chain.Config.DepositChainID))
```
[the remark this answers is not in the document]
Eisei24: Strange, I thought I searched exactly for that and replaced them but I guess there was some revert after it.
Fixed.

utils/utils.go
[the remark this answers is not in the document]
Eisei24: I rewrote the fields so that only the ones that give actual information depending on callstack and error are written.

"runtime.Caller" surely is something that takes longer than other calls but it was still just 7.5µs.

I moved the "info" part back into Error(...)/Fatal(...)

services/monitoring.go
```diff
@@ -46,7 +46,7 @@ func startClDataMonitoringService() {
 
 		if time.Since(utils.SlotToTime(maxAttestationSlot)) > time.Minute*15 {
 			errorMsg := fmt.Errorf("error: max attestation slot is older than 15 minutes: %v", time.Since(utils.SlotToTime(maxAttestationSlot)))
-			logger.Error(errorMsg)
+			utils.LogError(errorMsg.Error(), nil).Error()
```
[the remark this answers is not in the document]
Eisei24: As mentioned above I moved the "info" part back into Error(...)/Fatal(...)

utils/utils.go
```diff
@@ -1044,3 +1046,27 @@ func ForkVersionAtEpoch(epoch uint64) *types.ForkVersion {
...
+	if ok {
+		fileIdentifier = filepath.Base(fullFilePath)
+		functionIdentifier = runtime.FuncForPC(pc).Name()
+		lineIdentifier = strconv.Itoa(line)
+	} else {
+		msg := "Cannot read callstack"
+		fileIdentifier = msg
+		functionIdentifier = msg
+		lineIdentifier = msg
```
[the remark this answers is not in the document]
Eisei24: Good idea, I implemented it like that.

cmd/explorer/main.go
```diff
@@ -225,15 +228,15 @@ func main() {
 				for _, epoch := range utils.Config.Indexer.OneTimeExport.Epochs {
 					err := exporter.ExportEpoch(epoch, rpcClient)
 					if err != nil {
-						logrus.Fatal(err)
+						utils.LogError("exporting epoch error", err).Fatal()
```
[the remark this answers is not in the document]
Eisei24: I changed the error messages so that now all of them should be unique but I allowed it if one is an LogError and the other LogFatal since you can still distinguish them by the level.

exporter/exporter.go
Eisei24: You can now add an arbitrary amount of additional info strings to the log but this is not yet used by anything.

utils/utils.go
Eisei24: You now have to set the skip parameter when calling the new log functions.

utils/utils.go
```diff
@@ -1058,26 +1057,36 @@ func ForkVersionAtEpoch(epoch uint64) *types.ForkVersion {
...
+	}
+
+	if err != nil {
+		logFields = logFields.WithField("error type", fmt.Sprintf("%T", err)).WithError(err)
+	}
+
+	infoCount := 0
+	for _, info := range infoIdentifiers {
+		logFields = logFields.WithField(fmt.Sprintf("info_%v", infoCount), info)
 	}
```
D13ce, line 1089: `infoCount` is not increment, so passing more than 1 `infoIdentifiers` will only show the last one labeled `info_0`.

```suggestion
	for i, info := range infoIdentifiers {
		logFields = logFields.WithField(fmt.Sprintf("info_%v", i), info)
	}
```

utils/utils.go
```diff
@@ -1058,26 +1057,36 @@ func ForkVersionAtEpoch(epoch uint64) *types.ForkVersion {
...
+func LogFatal(err error, errorMsg interface{}, callerSkip int, infoIdentifiers ...string) {
+	logErrorInfo(err, callerSkip, infoIdentifiers...).Fatal(errorMsg)
+}
+
+func LogError(err error, errorMsg interface{}, callerSkip int, infoIdentifiers ...string) {
+	logErrorInfo(err, callerSkip, infoIdentifiers...).Error(errorMsg)
+}
 
-	pc, fullFilePath, line, ok := runtime.Caller(1)
+func logErrorInfo(err error, callerSkip int, infoIdentifiers ...string) *logrus.Entry {
```
D13ce, line 1068: `infoIdentifiers` is not the best name. It does not identify info, the info will be simply named info_x (see below) ... this is the ACTUAL info itself, is it not?

Maybe
```suggestion
func logErrorInfo(err error, callerSkip int, additionalInfo ...string) *logrus.Entry {
```
?

utils/utils.go
```diff
@@ -1058,26 +1057,36 @@ func ForkVersionAtEpoch(epoch uint64) *types.ForkVersion {
...
 
-func LogError(infoIdentifier string, err error) *logrus.Entry {
-	var fileIdentifier, lineIdentifier, functionIdentifier string
+func LogFatal(err error, errorMsg interface{}, callerSkip int, infoIdentifiers ...string) {
+	logErrorInfo(err, callerSkip, infoIdentifiers...).Fatal(errorMsg)
+}
+
+func LogError(err error, errorMsg interface{}, callerSkip int, infoIdentifiers ...string) {
+	logErrorInfo(err, callerSkip, infoIdentifiers...).Error(errorMsg)
+}
```
D13ce, line 1066: These two could really use a comment above the `func`. From the outside you would not have any idea what `callerSkip` is or what to do with `infoIdentifiers`.

utils/utils.go
```diff
@@ -1058,26 +1057,36 @@ func ForkVersionAtEpoch(epoch uint64) *types.ForkVersion {
...
+
+	pc, fullFilePath, line, ok := runtime.Caller(callerSkip + 2)
 	if ok {
-		fileIdentifier = filepath.Base(fullFilePath)
-		functionIdentifier = runtime.FuncForPC(pc).Name()
-		lineIdentifier = strconv.Itoa(line)
+		logFields = logFields.WithFields(logrus.Fields{
+			"file":     filepath.Base(fullFilePath),
+			"line":     line,
+			"function": runtime.FuncForPC(pc).Name(),
```
D13ce, line 1076: You are super lucky that the alphabetical order here is "file", "function", "line". This is exactly the order I would like the info to appear.

You are not so lucky that the additional info passed to this function is then labeled `info_X` (see below). This separates `line` from the rest of the call stack info.

Maybe
```suggestion
			"cs_file":     filepath.Base(fullFilePath),
			"cs_line":     line,
			"cs_function": runtime.FuncForPC(pc).Name(),
```
or something similar?

That way you would get `level`, `msg`, `call stack info`, `additional info`.

D13ce (changes requested):

utils/utils.go
Eisei24: Ah, sorry for that. Fixed.

utils/utils.go
Eisei24: Sure, let's rename it. I would just use the plural **additionalInfos**.
It was called that in
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/96de9fe6db245b4064a268482967bea19b3535dd/handlers/common.go#L347
but I don't really know why.

utils/utils.go
Eisei24: I added comments for them.

utils/utils.go
Eisei24: I see, so it's in alphabetic order but only for the fields themselves with level and msg always being first.
I did as you suggested and also moved line down in the code.

D13ce (approved):
lgtm

D13ce:
@recy21, this now looks really good in my opinion. Please take another look regarding performance and think about possible rules like "Do not use this in any API endpoint".

recy21 (approved):

PR #1860, opened 2023-02-27
Bids 1632/remove warning for0x01 creds

D13ce:


templates/validator/overview.html
```diff
@@ -235,12 +235,14 @@
       <span id="depositCount" style="cursor: pointer;" data-toggle="tooltip" title="Deposits made to the deposit contract"><i class="fas fa-wallet"></i> {{ .DepositsCount }}</span>
     </div>
     <div class="mx-3">
-      <span id="withdrawalCount" style="cursor: pointer;" data-toggle="tooltip" {{ if .IsWithdrawableAddress }}title="Withdrawals processed for this validator"{{ else }}title="To receive payouts you need 0x01 withdrawal credentials"{{ end }}>
+      <span id="withdrawalCount" style="cursor: pointer;" data-toggle="tooltip" {{ if or (.IsWithdrawableAddress) (not .CappellaHasHappened) }}title="Withdrawals processed for this validator"{{ else }}title="To receive payouts you need 0x01 withdrawal credentials"{{ end }}>
```
MauserBitfly, line 238: I think we should hide the whole ` <div class="mx-3"> ` block before cappella.
Does not make sense to just show the icon. 

Then you can remove the two CappellaHasHappened conditions within the block

image

one-three-three-seven: If we hide the icon, then we should hide the withdrawal tab too.

handlers/validator.go
```diff
@@ -380,10 +389,7 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 		validatorPageData.BLSChange = blsChange
 
 		if bytes.Equal(validatorPageData.WithdrawCredentials[:1], []byte{0x00}) && blsChange != nil {
```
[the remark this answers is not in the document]
D13ce: No, line 367 sets the variable if the credentials are 0x01. The line you are talking about sets it for credentials of 0x00 in combination with a blsChange.

templates/validator/overview.html
D13ce: Good suggestions, done

MauserBitfly (approved):

PR #1861, opened 2023-02-27
Bids 1612/spot finder

MauserBitfly:


templates/slot/components/slotfinder.html
```diff
@@ -0,0 +1,111 @@
...
+        const slotDateLocal = moment(slotDate)
+        const slotDateUtc = moment.utc(slotDate)
+        datepickerUtc.data("daterangepicker").setStartDate(slotDateUtc)
+        datepickerUtc.data("daterangepicker").setEndDate(slotDateUtc)
+        datepickerLocal.data("daterangepicker").setStartDate(slotDateLocal)
+        datepickerLocal.data("daterangepicker").setEndDate(slotDateLocal)
+
+        const timePassed = slotDate.getTime() / 1000 - genesisTimeStamp
+        const daysPassed = Math.max(1, 1 + Math.floor(timePassed / 60 / 60 / 24))
+        beaconchainDay.innerText = daysPassed
```
D13ce, line 53: This is tricky: According to the slot finder, epoch 0 is part of beaconchain day 1.
If you check our API though:
* `/api/v1/validator/stats/{index}` returns values for day 0.
* `/api/v1/ethstore/{day}` returns something different for day 0 compared to day 1.

=> Please check if we should display beaconchain 0 for epoch 0.

templates/slot/components/upgradescheduler.html
```diff
@@ -0,0 +1,161 @@
...
+      const hour = tzDate.plus({ minutes: tzDate.o }).hour
+      if (hour >= 8 && hour <= 16) {
+        return "perfect"
+      } else if (hour >= 7 && hour <= 20) {
+        return "good"
+      } else if (hour >= 5 && hour <= 23) {
+        return "ok"
+      }
+      return "bad"
+    }
```
D13ce, line 97: The color scheme implemented matches the one used in https://slots.symphonious.net/ but does NOT match the according BIDS' description. Is that intended?

templates/slot/components/upgradescheduler.html
```diff
@@ -0,0 +1,161 @@
...
+      row.appendChild(col)
+    }
+
+    function setupTable(date) {
+      const table = document.getElementById("upgrade-table")
+      table.innerHTML = ""
+      const utc = luxon.DateTime.utc(date.getFullYear(), date.getMonth() + 1, date.getDate(), 12)
+      const dif = utc.ts / 1000 - genesisTimeStamp
+      const firstSlot = Math.max(1, dif / secondsPerSlot)
+      const firstEpoch = 1 + Math.floor(firstSlot / slotsPerEpoch)
```
D13ce, line 121: Why the +1 here?

Currently it is impossible to view epoch 0.

D13ce (changes requested):
I have left a few comments, changes are requested.

Furthermore:

1. The Upgrade Scheduler on https://slots.symphonious.net/ shows all epochs for a day (based on UTC). Select a random day and the first epoch shown will be right after midnight of the given day and the last one will be right before midnight of the next day.
Our Upgrade Scheduler is different. For example, on Zhejiang, select `02/02/2023` and take a look at the UTC column: The first epoch is 197 which happened at noon and not at midnight (midnight would be the start of the day).
Please discuss with our product owner what the desired behavior should be.
2. **Minor:** To the right of the Upgrade Scheduler's table there is a lot of empty space and the scrollbar is to the right of that. This could be improved.
3. **Minor:** To me it is not clear what the Upgrade Schedule does immediately. We might want to add some tool-tips or description above the table?

templates/slot/components/upgradescheduler.html
MauserBitfly: Yes, discussed it with Inan ... will add a comment in the BIDS

templates/layout.html
```diff
@@ -139,6 +139,7 @@
               }
             })
         </script> {{ template "css" .Data }}
+      <script src="/js/jquery.min.js"></script>
```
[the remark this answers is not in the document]
MauserBitfly: because I needed in to load before my components, otherwise the external libraries added there would not work

MauserBitfly:
> 

@D13ce : For your first Point ... both https://slots.symphonious.net/  and our new scheduler start at Noon. 12 PM = noon, 12 AM = Midnight

templates/slot/components/slotfinder.html
```diff
@@ -0,0 +1,111 @@
...
+  <small class="d-flex justify-content-end align-items-center m-2">Inpired by <a href="https://slots.symphonious.net" class="ml-2">https://slots.symphonious.net</a></small>
+  <script>
+    var genesisTimeStamp = {{ .ChainGenesisTimestamp }}
+    var secondsPerSlot = {{ .ChainSecondsPerSlot }}
+    var slotsPerEpoch = {{ .ChainSlotsPerEpoch }}
+    var slotNumber, datepickerUtc, datepickerLocal, beaconchainDay, slotLink
+
+    function slotChanged() {
+      const slot = parseInt(slotNumber.value)
+      if (!isNaN(slot) && slot > 0) {
```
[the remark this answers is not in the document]
MauserBitfly: Because I wrongly thought we start with the 1st slot ... will fix it.

handlers/slot_finder.go
```diff
@@ -0,0 +1,20 @@
+package handlers
+
+import (
+	"eth2-exporter/templates"
+	"net/http"
+)
+
+// Blocks will return information about blocks using a go template
```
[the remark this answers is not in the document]
MauserBitfly: replaced (also in the other occurrences )

templates/slot/components/slotfinder.html
MauserBitfly: Wrongly started with 1 ... Slots/Epochs/Days start with 0 ... fixed

templates/slot/components/upgradescheduler.html
MauserBitfly: Wrongly started with 1 ... Slots/Epochs/Days start with 0 ... fixed

D13ce (approved):
lgtm

PR #1862, opened 2023-02-27
Bids 1643/minor withdrawal improvements

D13ce:


db/db.go
```diff
@@ -2523,10 +2523,6 @@ func GetAddressWithdrawals(address []byte, limit uint64, offset uint64) ([]*type
 		limit = 100
 	}
 
-	if limit > (offset + 100) {
-		limit = offset + 100
-	}
```
Eisei24, line 2528: I also noticed such weird settings for APIs.
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d6eda8928fcf92ffdfac1923ea75c1327e9be767/handlers/api.go#L562
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d6eda8928fcf92ffdfac1923ea75c1327e9be767/handlers/api.go#L570
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d6eda8928fcf92ffdfac1923ea75c1327e9be767/handlers/api.go#L1550

Should they immediately be changed or is this not relevant enough for the issue?

handlers/api.go
```diff
@@ -1777,7 +1777,7 @@ func ApiValidatorBlsChange(w http.ResponseWriter, r *http.Request) {
 // @Param  latest_epoch query int false "The latest epoch to consider in the query"
 // @Param  offset query int false "Number of items to skip"
 // @Param  limit query int false "Maximum number of items to return, up to 100"
-// @Success 200 {object} types.ApiResponse{data=[]types.ApiValidatorBlsChangeResponse}
+// @Success 200 {object} types.ApiResponse{data=[]types.ApiValidatorBalanceHistoryResponse}
```
Eisei24, line 1780: *ApiValidatorBalanceHistoryResponse* is correct here but it is not used within the function itself, a new struct that is the same gets created instead.

handlers/dashboard.go
```diff
@@ -345,6 +345,9 @@ func DashboardDataProposals(w http.ResponseWriter, r *http.Request) {
 }
 
 func DashboardWithdrawals(w http.ResponseWriter, r *http.Request) {
+
+	// [WIP]
```
Eisei24, line 349: An intended comment?

handlers/api.go
```diff
@@ -1667,14 +1667,14 @@ func ApiValidatorWithdrawals(w http.ResponseWriter, r *http.Request) {
...
 	epoch, err := strconv.ParseUint(q.Get("epoch"), 10, 64)
 	if err != nil {
-		epoch = latestEpoch
+		epoch = services.LatestEpoch()
 	}
 
-	endEpoch := epoch - 100
-	if epoch < 100 {
+	// startEpoch and endEpoch are both inclusive, so substracting 99 here will result in a limit of 100 epochs
+	endEpoch := epoch - 99
```
Eisei24, line 1676: **Minor:** Since endEpoch is created as uint64 it can get set to something like 18446744073709551527 for a line.
Maybe something like
```suggestion
endEpoch, err := math.Sub(epoch, 99)
if err != nil {
   endEpoch = 0
}
```
would be prefered?

Eisei24, line 1676: **Minor:** endEpoch gets created as an uint64 so for only this line it could be set to something like 18446744073709551527.

handlers/dashboard.go
D13ce: Yes

handlers/api.go
D13ce: Good catch, fixed.

db/db.go
D13ce: I have created BIDS-1665 for that.

Eisei24:
lgtm

Eisei24 (approved):

PR #1883, opened 2023-03-01
(BIDS-1646) show validator history up to his exit

MauserBitfly:


recy21 (approved):

handlers/validator.go
```diff
@@ -1514,6 +1514,10 @@ func ValidatorHistory(w http.ResponseWriter, r *http.Request) {
 	}
 
 	currentEpoch := services.LatestEpoch() - 1
+	// for an exited validator we show the history until his exit
+	if activationAndExitEpoch.ExitEpoch != 9223372036854775807 {
+		currentEpoch = activationAndExitEpoch.ExitEpoch - 1
+	}
```
D13ce, line 1520: This does not meet the requirements specified. According to the [BIDS](https://bitfly1.atlassian.net/browse/BIDS-1646) description, the last entry for an exited validator should be the balance change to 0 ETH.

Let's take your example validator: https://zhejiang.beaconcha.in/validator/62092#deposits

The last entry shown with your changes is their attestation and withdrawal in epoch 3104.

The full withdrawal is not shown though and the problem is that this does not happen during the exit epoch. It happens several epochs after that. The validator in question for example can be fully withdrawn only AFTER epoch 3361 (depending on the witdhdrawal cycle).

So you need to check whether the validator has already been withdrawn completely and if so, add the according entry to the table.

D13ce (changes requested):

templates/validator/tables.html
```diff
@@ -594,19 +594,12 @@ <h6 class="">This table displays the deposits received and processed by the beac
...
           </tr>
         </thead>
-        <tbody>
-          {{ if and (not (or (eq .ExitEpoch 9223372036854775807) (eq .ExitEpoch 0))) (gt $.LatestFinalizedEpoch .ExitEpoch) }}
-            <tr colspan="3">
-              validator has exited and is no longer required to perform any duties.
-            </tr>
-          {{ end }}
-        </tbody>
+        <tbody></tbody>
```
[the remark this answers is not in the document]
MauserBitfly: The table body will be filled by the DataTable library once the data is loaded. 
Yes I tested it and the lib also adds the tbody if its missing, but I think it's nicer to already have it there. And it's also the same for all the other tables in the tables.html that use the DataTable lib.

handlers/validator.go
D13ce: This has been discussed in the according BIDS and @Buttaa approved that the full withdrawal does not need to be shown in the validator history.

D13ce (approved):
lgtm

PR #1888, opened 2023-03-02
(BIDS-1588) Show "Contract" on address page

Eisei24:
An address that should be a contract is:
https://prater.beaconcha.in/address/0xff50ed3d0ec03ac01d4c79aad74928bff48a7b2b

### DIFF
--- a/cmd/explorer/main.go
+++ b/cmd/explorer/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"math/big"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


cmd/explorer/main.go
```diff
@@ -190,8 +191,20 @@ func main() {
...
+			cfg.WriterDatabase.Port,
+			cfg.WriterDatabase.Name}
+		DBStr := strings.Join(DBInfo, "-")
+		frontendDBInfo := []string{
+			cfg.Frontend.WriterDatabase.Username,
+			cfg.Frontend.WriterDatabase.Password,
+			cfg.Frontend.WriterDatabase.Host,
+			cfg.Frontend.WriterDatabase.Port,
+			cfg.Frontend.WriterDatabase.Name}
+		frontendDBStr := strings.Join(frontendDBInfo, "-")
```
Eisei24, line 207: Just a cleanup I thought makes this a lot more readable.
No functionality change.

handlers/eth1Account.go
```diff
@@ -166,6 +169,15 @@ func Eth1Address(w http.ResponseWriter, r *http.Request) {
...
 
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+
+	code, err := eth1data.GetCodeAt(ctx, common.BytesToAddress(addressBytes))
+	if err != nil {
+		logger.WithError(err).Errorf("error retrieving code data for address %v", address)
+	}
+	isContract := len(code) != 0
+
```
Eisei24, line 180: This is how transactions calculate it to save it in the TieredCache.

Maybe this info should also be somewhere in the TieredCache for addresses?
Maybe in "metadata for address"?
Please let me know your thoughts.

Eisei24: **Update:** The main problem should be that a "selfdestruct" could be called for a contract which deletes its code so we could no longer recognize it as a contract here.
However do we not have the same problem for tx?
Does something more far reaching need to be changed?

handlers/eth1Account.go
```diff
@@ -166,6 +169,15 @@ func Eth1Address(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+
+	code, err := eth1data.GetCodeAt(ctx, common.BytesToAddress(addressBytes))
+	if err != nil {
+		logger.WithError(err).Errorf("error retrieving code data for address %v", address)
+	}
+	isContract := len(code) != 0
```
D13ce, line 179: This is adding duplicated code, see here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/1f2df8d694574a6f673a4d309f5499c973cc560b/eth1data/eth1data.go#L80-L84

Please add a `IsContract` function (most likely in `eth1data.go´?) and use that.

D13ce (changes requested):

handlers/eth1Account.go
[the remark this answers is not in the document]
Eisei24: Yeah, that's better.
I also removed the error text to be more consistent with the others and the error itself already was such a text in the first place.

Eisei24: Done.

recy21: > I ping @recy21 though, maybe he has a different opinion.

No, lets do like that, thanks!

handlers/eth1Account.go
D13ce: @recy21 answered here: https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1888#discussion_r1125705285

Eisei24: The new issue to handle this is BIDS-1688.

PR #1892, opened 2023-03-02
(BIDS-1679) fix race condition

peterbitfly:


D13ce (approved):
lgtm

I have found another race condition though and will commit a fix in the next minutes. Once done, I will request a review from @peterbitfly. If everything is fine, he can then approve, merge and delete the branch.

handlers/validator.go
```diff
@@ -492,15 +497,12 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 				break
 			}
 		}
+
+		validatorPageData.ShowMultipleWithdrawalCredentialsWarning = hasMultipleWithdrawalCredentials(validatorPageData.Deposits)
```
D13ce, line 501: This depends on `validatorPageData.Deposits` being set which happens in this very Go routine.

handlers/validator.go
```diff
@@ -398,6 +398,11 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 		validatorPageData.CurrentBalance = vbalance.Balance
 		validatorPageData.EffectiveBalance = vbalance.EffectiveBalance
 
+		if bytes.Equal(validatorPageData.WithdrawCredentials[:1], []byte{0x01}) {
+			// validators can have 0x01 credentials even before the cappella fork
+			validatorPageData.IsWithdrawableAddress = true
+		}
```
D13ce, line 404: Technically it is not required to have this in this Go routine. But since this is related to all the things happening here, I put it here for readability reasons.

PR #1901, opened 2023-03-06
(BIDS-1191) Fix avg block time calculation and rounding

Eisei24:


db/statistics.go
```diff
@@ -611,7 +611,7 @@ func WriteChartSeriesForDay(day int64) error {
 		totalGasLimit = totalGasLimit.Add(decimal.NewFromInt(int64(blk.GasLimit)))
 
 		if prevBlock != nil {
-			avgBlockTime = avgBlockTime.Add(decimal.NewFromInt(prevBlock.Time.AsTime().UnixMicro() - blk.Time.AsTime().UnixMicro())).Div(decimal.NewFromInt(2))
+			avgBlockTime = avgBlockTime.Add(decimal.NewFromInt(prevBlock.Time.AsTime().UnixMicro() - blk.Time.AsTime().UnixMicro()))
```
Eisei24, line 614: The changes in this file I could not test myself but the previous calculation makes no sense to me since it weighs the duration of blocks that happen early during the day much much higher than at the end.

services/charts_updater.go
```diff
@@ -1271,7 +1271,6 @@ func BlockTimeAvgChartData() (*types.GenericChartData, error) {
 		ColumnDataGroupingApproximation: "average",
 		TooltipFormatter: `
 		function (tooltip) {
-			this.point.y = Math.round(this.point.y)
```
Eisei24, line 1274: The db is empty but I could test it by faking some values into **rows** after the query above.

D13ce: This change alone does not remove all rounding, does it? See here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d97a4ac8cd08b81345c836d18874bddd8966ca33/services/charts_updater.go#L1250

D13ce (changes requested):
I have left a few comments. Please take a look.

db/statistics.go
Eisei24: Done.

db/statistics.go
```diff
@@ -675,6 +675,8 @@ func WriteChartSeriesForDay(day int64) error {
 		prevBlock = blk
 	}
 
+	avgBlockTime = avgBlockTime.Div(decimal.NewFromInt(blockCount - 1))
```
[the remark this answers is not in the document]
Eisei24: No, **blocksChan** is a channel.

services/charts_updater.go
Eisei24: Yes, this rounds to 2 digits after the comma like etherscan shows it.

D13ce: I see. I was confused by the Jira comments.

db/statistics.go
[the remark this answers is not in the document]
D13ce: mb

D13ce (approved):
utACK

PR #1904, opened 2023-03-07
BIDS-1685/fix_validator_view_pending_logged_in

peterbitfly:


handlers/validator.go
```diff
@@ -222,6 +223,25 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
+						EventLabel: ev.Desc,
+						EventName:  ev.Event,
+						Active:     false,
+						Warning:    ev.Warning,
+						Info:       ev.Info,
+					})
+				}
+				validatorPageData.AddValidatorWatchlistModal = &types.AddValidatorWatchlistModal{
+					Events:         events,
+					ValidatorIndex: validatorPageData.Index,
```
D13ce, line 240: **Minor:** This is the if branch where no validator index is available. This line therefore sets the `ValidatorIndex` from 0 to 0, right?

It does no harm but I thought that this line is confusing here and suggest to remove it:
```suggestion
```

handlers/validator.go
```diff
@@ -222,6 +223,25 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
+					CsrfField:      csrf.TemplateField(r),
+				}
+			}
```
D13ce, line 243: **Minor:** This is adding more duplicated code to the "no validator index found" if branch. I suggest to clean this up.

D13ce (approved):
lgtm

I have just posted two comments regarding possible code improvements that are minor. Feel free to merge.

PR #1906, opened 2023-03-07
(BIDS-1546) New Validator Page Income

LuccaBitfly:


handlers/common.go
```diff
@@ -130,27 +128,214 @@ func GetValidatorEarnings(validators []uint64, currency string) (*types.Validato
...
+	}
+	// Timeframe constants
+	fiveDays := time.Hour * 24 * 5
+	oneWeek := time.Hour * 24 * 7
+	oneMonth := time.Hour * 24 * 30
+	sixWeeks := time.Hour * 24 * 45
+	twoMonths := time.Hour * 24 * 60
+	threeMonths := time.Hour * 24 * 90
+	fourMonths := time.Hour * 24 * 120
+	fiveMonths := time.Hour * 24 * 150
```
Eisei24, line 230: Could use `utils.Day` and `utils.Month`.

handlers/common.go
```diff
@@ -130,27 +128,214 @@ func GetValidatorEarnings(validators []uint64, currency string) (*types.Validato
...
+	// Return the luck as the ratio of qualified proposals to expected slot proposals
+	return float64(qualifiedProposalCount) / expectedSlotProposals
+}
+
+// calcExpectedSlotProposals calculates the expected number of slot proposals for a certain time frame and validator count
+func calcExpectedSlotProposals(timeframe time.Duration, validatorCount int, activeValidatorsCount uint64) float64 {
+	if validatorCount == 0 || activeValidatorsCount == 0 {
+		return 0
+	}
+	slotsPerTimeframe := timeframe.Seconds() / float64(utils.Config.Chain.Config.SecondsPerSlot)
```
Eisei24, line 287: **Minor:** I think `slotsPerTimeframe` is a bad name since it's not "slots per time" but just a slot count, how about `slotsInTimeframe`?

handlers/common.go
```diff
@@ -130,27 +128,214 @@ func GetValidatorEarnings(validators []uint64, currency string) (*types.Validato
...
+	return (slotsPerTimeframe / float64(activeValidatorsCount)) * float64(validatorCount)
+}
+
+// getNextBlockEstimateTimestamp will return the estimated timestamp of the next block proposal
+// given the blocks proposed by the validators and the number of validators
+//
+// precondition: proposedBlocks is sorted by ascending block number
+func getNextBlockEstimateTimestamp(slots []uint64, validatorsCount int) *time.Time {
```
Eisei24, line 295: - We only use the latest block that was proposed for the calculation so there is no need to give all of them as a parameter.
- Does it make sense to show the estimated time based on the that last block? If something would happen once a year based on some probability and he checks after (1 year - 5 seconds) the text would say `(exp. in 5s)` so the reader might expect there to be a really high chance for it to happen now but it is still as always a low "once a year" chance.
Maybe a text like `(exp. avg. once every 1 year)` would be more appropriate?

handlers/validator.go
```diff
@@ -574,6 +593,40 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
+			// get el data
+			execBlocks, err := db.BigtableClient.GetBlocksIndexedMultiple(proposedToday, 10000)
+			if err != nil {
+				return fmt.Errorf("error retrieving execution blocks data from bigtable: %v", err)
+			}
+
+			// get mev data
+			relaysData, err := db.GetRelayDataForIndexedBlocks(execBlocks)
+			if err != nil {
+				return fmt.Errorf("error retrieving mev bribe data from bigtable: %v", err)
```
Eisei24, line 615: The data is not in the bigtable.

types/templates.go
```diff
@@ -328,6 +328,7 @@ type ValidatorsPageDataValidators struct {
 	Performance7d              int64  `db:"performance7d"`
 }
 
+// ValidatorPageData is a struct to hold data for the validators page
```
Eisei24, line 331: Double comment.

Eisei24 (commented):
Everything should work as required but please have a look at my comments.

handlers/common.go
```diff
@@ -130,27 +128,214 @@ func GetValidatorEarnings(validators []uint64, currency string) (*types.Validato
 		return nil, nil, err
 	}
 
-	apr = (((float64(c.EarningsLastWeek) / 1e9) / (float64(32))) * 365) / 7
-	if apr < float64(-1) {
-		apr = float64(-1)
+	// calculate combined el and cl earnings
+	earnings1d := ((income.ClIncome1d * 1e9) + income.ElIncome1d) / 1e9
+	earnings7d := ((income.ClIncome7d * 1e9) + income.ElIncome7d) / 1e9
+	earnings31d := ((income.ClIncome31d * 1e9) + income.ElIncome31d) / 1e9
```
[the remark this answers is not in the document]
LuccaBitfly: you're right, I assumed that I would need to calculate everything in Wei before converting to Gwei in order to avoid precision loss, but after thinking about it, it makes no difference.

adressed in new commit, ptal

handlers/common.go
```diff
@@ -130,27 +128,214 @@ func GetValidatorEarnings(validators []uint64, currency string) (*types.Validato
...
+
+	// since only the first 5 digits are shown in the frontend, the lost precision is probably negligible
+	income.ElIncome1d /= 1e9
+	income.ElIncome7d /= 1e9
+	income.ElIncome31d /= 1e9
+	income.ElIncome365d /= 1e9
+	income.ElIncomeTotal /= 1e9
```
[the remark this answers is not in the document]
LuccaBitfly: adressed in new commit, ptal

handlers/common.go
LuccaBitfly: adressed in new commit, ptal

handlers/common.go
LuccaBitfly: I honestly don't even know where the difference between "in" and "per" would be, but addressed in new commit, ptal

handlers/common.go
LuccaBitfly: after talking to @Buttaa we're going to show both in the tooltip, ptal

handlers/validator.go
LuccaBitfly: addressed in new commit, ptal

types/templates.go
LuccaBitfly: addressed in new commit, ptal

handlers/validator.go
```diff
@@ -606,7 +621,13 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 		}
 
 		validatorPageData.ProposalLuck = getProposalLuck(slots, 1)
-		validatorPageData.ProposalEstimate = getNextBlockEstimateTimestamp(slots, 1)
+		avgSlotInterval := uint64(getAvgSlotInterval(1))
+		avgSlotIntervalAsDuration := time.Duration(utils.Config.Chain.Config.SecondsPerSlot * avgSlotInterval * 1e9)
```
Eisei24, line 625:
```suggestion
		avgSlotIntervalAsDuration := time.Duration(utils.Config.Chain.Config.SecondsPerSlot * avgSlotInterval ) * time.Second
```

Like above.

Eisei24 (changes requested):
There is still one bug I noticed at the end.
If you load multiple validators at the same time only one of them gets their **Luck** times filled.

E.g. I loaded validators 103447 and 86437 and 86437 just showed
```
Block Prop. (exp. ) 41.98% (on avg. every )
Sync  Com. (exp. ) 318.51% (on avg. every )
```

handlers/validator.go
LuccaBitfly: addressed in new commit, ptal

Eisei24 (approved):
lgtm

PR #1908, opened 2023-03-07
Bids 1544/show inactive validators as inactive

Eisei24:
You can check with https://prater.beaconcha.in/dashboard?validators=11877,55293,222906,255562,311255,453473 but the last one might no longer be pending by that time.

### DIFF
--- a/templates/dashboard.html
+++ b/templates/dashboard.html
@@ -229,7 +229,7 @@
                   </div>
                 </span>
                 <div style="font-weight:300;font-size:1rem;display:flex;justify-content:center" class="pb-2">
-                  <span id="dashboard-info">Found 0 pending, 0 active and 0 exited validators</span>
+                  <span id="dashboard-info">0 active (0 online, 0 offline), 0 pending, 0 exited validators</span>
                 </div>
               </div>
             </div>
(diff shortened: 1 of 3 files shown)


templates/dashboard.html
```diff
@@ -229,7 +229,7 @@
                   </div>
                 </span>
                 <div style="font-weight:300;font-size:1rem;display:flex;justify-content:center" class="pb-2">
-                  <span id="dashboard-info">Found 0 pending, 0 active and 0 exited validators</span>
+                  <span id="dashboard-info">0 active (0 online, 0 offline), 0 pending, 0 exited validators</span>
```
Eisei24, line 232: Afaik this is just a dummy text that just serves to illustrate what the formatting looks like.

D13ce: This is not entirely true, this text is shown as long as the dashboard is loading. I surfed to the link you sent to test your changes and saw said text for ~2s.

D13ce (commented):
Your changes look good but slashed validators are not part of the text. I have asked @Buttaa in Jira about this, let's see what he says.

templates/dashboard.html
Eisei24: I see, thanks for the info.

Eisei24:
Slashed validators are now also written if we have some (no "0 slashed").

D13ce (approved):
lgtm

PR #1916, opened 2023-03-09
(NOBIDS) demote status logging to trace

peterbitfly:


db/db.go
```diff
@@ -1115,7 +1115,7 @@ func saveValidators(data *types.EpochData, tx *sqlx.Tx, client rpc.Client) error
 			// }
 
 			if c.Status != v.Status {
-				logger.Infof("Status changed for validator %v from %v to %v", v.Index, c.Status, v.Status)
+				logger.Tracef("Status changed for validator %v from %v to %v", v.Index, c.Status, v.Status)
```
recy21, line 1118: @peterbitfly "Trace", like "Debug", will not show up in the logs, not sure why, but it's like that. Is this you intention?

recy21 (approved):

PR #1918, opened 2023-03-09
Bids 1683/recently activated validators have no income

Eisei24:
To fix/test this I copied all the relevant code for **cl_rewards_gwei** from **WriteValidatorStatisticsForDay** to a new function and let it run near the end of the explorer/main with the last day as a paramter (day 34 at the time of fixing).

To fully fix this we probably need to run **cmd/statistics/main.go** for all days again.

### DIFF
--- a/db/bigtable.go
+++ b/db/bigtable.go
@@ -1355,7 +1355,7 @@ func (bigtable *Bigtable) GetValidatorIncomeDetailsHistory(validators []uint64, 
 		startEpoch = 0
 	}
 
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Second*180))
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*180)
 	defer cancel()
 
 	ranges, err := bigtable.getEpochRanges(startEpoch, endEpoch)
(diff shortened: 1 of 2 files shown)


db/bigtable.go
```diff
@@ -1355,7 +1355,7 @@ func (bigtable *Bigtable) GetValidatorIncomeDetailsHistory(validators []uint64,
 		startEpoch = 0
 	}
 
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Second*180))
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*180)
```
Eisei24, line 1358: Not related to the bug.

db/statistics.go
```diff
@@ -470,11 +477,23 @@ func GetValidatorIncomeHistory(validator_indices []uint64, lowerBoundDay uint64,
...
+		lastDay := int64(0)
+		if len(result) > 0 {
+			lastDay = result[len(result)-1].Day
+		} else {
+			err = ReaderDb.Get(&lastDay, "SELECT COALESCE(MAX(day), 0) FROM validator_stats")
+			if err != nil {
+				return nil, 0, err
+			}
+		}
+
```
Eisei24, line 496: This fixes the first bug that the "Today" income for validators that were created that same day (hence len(result) = 0) is 0.

db/statistics.go
```diff
@@ -119,44 +119,51 @@ func WriteValidatorStatisticsForDay(day uint64) error {
 	}
 
 	batchSize := 16000 // max parameters: 65535
-	for b := 0; b < len(incomeStats); b += batchSize {
```
Eisei24, line 122: This fixes the second bug that "Yesterday" income does not get set (and is then also missing in "Last Week" & "Last Month") for validators that are too new.
This is because **incomeStats** is a map and not a slice and does not contain every validator.

peterbitfly (approved):

PR #1926, opened 2023-03-09
(BIDS-1745) Remove "Yesterday" from validator page

Buttaa:
Unfortunately this shifts the whole validator design. It will be added in the next design


recy21:
Already fixed in https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1928

PR #1929, opened 2023-03-10
(BIDS-1459) Add etermine staking and eversteel to menu

Eisei24:
I can't quite tell but is the new text a few pixels too high up?
Is there even any way to adjust that?


Eisei24:
The changes for the two CSS which regulate light and dark mode are impossible to read here.
The only change is that I added `--body-color-inverted` which flips the `--body-color` entry which already existed between light and dark mode.
If you know of any other way to get the flipped colour this new entry could be deleted.

beacon-light.min.css has a [data-theme=dark] part at the very bottom which has the flipped `--body-color` but I have no idea if there is even any way to access that.

templates/layout/mainnavigation.html
```diff
@@ -55,14 +55,27 @@
...
+      {{ else }}
+        <span class="nav-icon">
+          {{ if .CustomIcon }}
+            {{ includeSvg .CustomIcon }}
+          {{ else }}
+            <i class="fas {{ .Icon }} mr-2"></i>
+          {{ end }}
+        </span>
+        <span class="nav-text">{{ .Label }}</span>
+      {{ end }}
```
D13ce, line 78: This creates duplicated code, consider something like this:
```suggestion
      {{ $buttonClass := "" }}
      {{ $textClass := "nav-text" }}
      {{ if .IsHighlighted }}
        {{ $buttonClass = "btn btn-primary" }}
        {{ $textClass = "text-white" }}
      {{ end }}


      <span class="{{ $buttonClass }}">
        <span class="nav-icon">
          {{ if .CustomIcon }}
            {{ includeSvg .CustomIcon }}
          {{ else }}
            <i class="fas {{ .Icon }} mr-2"></i>
          {{ end }}
        </span>
        <span class="{{ $textClass }}">{{ .Label }}</span>
      </span>
```

D13ce (changes requested):
Editing `beacon-dark.min.css` / `beacon-light.min.css` manually is not advised. I will post a comment a Jira regarding that.

Furthermore: The alignment looks off. Consider contacting @MauserBitfly.
image
I have also left a little refactor suggestion.

templates/layout/mainnavigation.html
Eisei24: Thanks, I did not consider that you could create a `span` with an empty `class`.

Eisei24: I think the horizontal alignment is supposed to be this way. The button starts left aligned and then wants a bit of space until the icon starts. See the screenshot of the mobile version in the BIDS.
I will check if the vertical alignment can be improved.

templates/layout/mainnavigation.html
```diff
@@ -55,14 +55,22 @@
 {{ define "mainNavigationItem" }}
   {{ if not .IsHidden }}
     <a class="dropdown-item" href="{{ .Path }}">
```
MauserBitfly, line 57: Try this to align the menu items:
```suggestion
   <a class="dropdown-item" {{ if .IsHighlighted }}style="padding:0 0.625rem;"{{ end }} href="{{ .Path }}">
```

D13ce (approved):
lgtm

PR #1935, opened 2023-03-11
(NOBIDS) Removed unused field "name" from Relays struct

recy21:


recy21:
As there is no field "name" in database, there shouldn't be a field "name" in the struct.

D13ce (approved):
lgtm

PR #1939, opened 2023-03-13
Bids 1593/validator names index

Eisei24:
Pages where validator changes happened but which should still work unchanged:

- /epoch/{epoch}
- /slots
- /slot/{slot}
- /blocks
- /block/{block}
- /search/{type}/{search}
The search function when its type is "indexed_validators_by_name"
- /validator/{pubkey}/save
This can't be tested yet because you need to edit the names of the validators.
Can be done for Prater later where we have our own Goerli validators, I guess.
- /validators/slashings
- /validators/streakleaderboard
This can't be tested yet because it is empty for Zhejiang/Sepolia.


Eisei24:
**IMPORTANT:** This requires a database update that needs to be mentioned in the "Next Merge Issue".
`validator_names` now has a new column `index` which is a copy of `validators` column `validatorindex`

Eisei24:
The BIDS was rejected so I am closing this PR.

PR #1940, opened 2023-03-13
(BIDS-1491) Add POST option for api and make it the default

Eisei24:
I couldn't test whether this fixes the "URI too long" response body problem since localhost didn't have it in the first place.


handlers/api.go
```diff
@@ -2292,15 +2315,15 @@ func ApiGraffitiwall(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
-	slotQuery = math.MaxU64(slotQuery, 10000)
+	slotQuery = utilMath.MaxU64(slotQuery, 10000)
```
D13ce, line 2318: Why the switch to `utilMath`? This now is only used within this file. Searching `MinU64` at least shows one other `math` use:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/37c6603b943dd149abacf0bf10a3d82ae1a950a6/handlers/validator.go#L1947

handlers/api.go
```diff
@@ -1326,17 +1327,39 @@ func getEpoch(epoch int64) ([]interface{}, error) {
...
+// @Param  indexOrPubkey path string true "Validator indicesOrPubkeys, comma separated"
+// @Success 200 {object} types.ApiResponse{data=[]types.APIValidatorResponse}
+// @Failure 400 {object} types.ApiResponse
 // @Router /api/v1/validator/{indexOrPubkey} [post]
-func ApiValidator(w http.ResponseWriter, r *http.Request) {
+func ApiValidatorPost(w http.ResponseWriter, r *http.Request) {
+	apiValidator(w, r)
+}
 
+func apiValidator(w http.ResponseWriter, r *http.Request) {
```
D13ce, line 1352: Please leave a comment explaining this "odd at the first glance" construction.

Maybe something like this:
```suggestion
// This endpoint supports both GET and POST but requires different swagger descriptions based on the type
func apiValidator(w http.ResponseWriter, r *http.Request) {
```

D13ce (commented):
Seems to work as intended but I have posted a question and a suggestion.

handlers/api.go
Eisei24: I wanted to use the normal math dependency just for a single call of `math.MaxInt` and since 2 dependencies can't have the same name I had to rename one.

handlers/api.go
Eisei24: Sure, added.

handlers/api.go
D13ce: My bad, I misread that.

D13ce (approved):
lgtm

PR #1945, opened 2023-03-14
(BIDS-1763) show countdown for slot viz

MauserBitfly:


handlers/charts.go
```diff
@@ -146,7 +146,11 @@ func SlotViz(w http.ResponseWriter, r *http.Request) {
 		Selector: "checklist",
 		Epochs:   services.LatestSlotVizMetrics(),
 	}
-	data.Data = slotVizData
+	data.Data = struct {
+		SlotVizData types.SlotVizPageData
+	}{
+		SlotVizData: slotVizData,
+	}
```
D13ce, line 153: **Minor:** Consider adding a comment why this is done the way it is done.

D13ce (changes requested):
The slot viz does not update slots automatically anymore. This neither works when the countdown disappears nor when it wasn't even there to begin with.

FYI: On master this currently seems to work.

handlers/charts.go
MauserBitfly: comment added

MauserBitfly:
> The slot viz does not update slots automatically anymore. This neither works when the countdown disappears nor when it wasn't even there to begin with.
> 
> FYI: On master this currently seems to work.

redraw should work again

geenutts (approved):

templates/slotViz.html
```diff
@@ -416,40 +403,46 @@
         if (!slotVizCountdown) {
           slotVizCountdown = document.createElement("div")
           slotVizCountdown.className = "mb-3"
-          $("#" + selector).prepend(slotVizCountdown)
+          $(selector).prepend(slotVizCountdown)
         }
         const slots = Math.ceil((hfTs - now) / 1000 / secondsPerSlot)
-        slotVizCountdown.innerHTML = `<b>Hardfork in:</b> ${getRelativeTime(luxon.DateTime.fromMillis(hfTs))} | ${slots} slots`
+        slotVizCountdown.innerHTML = `<b>${hardforkName} Hardfork in:</b> ${getRelativeTime(luxon.DateTime.fromMillis(hfTs))} | ${slots} slots`
```
D13ce, line 409: This adds an unnecessary space in front of "Hardfork" when no `${hardforkName}` is set.

Consider putting the space into `${hardforkName}` whenever a name is configured directly here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/228994874181d60551b15675d1e8b8ca6a299efa/templates/slotViz.html#L11

D13ce, line 409: I have just noticed that the word "in" is shown twice:
image

This is caused by https://github.com/gobitfly/eth2-beaconchain-explorer/blob/c3c4adb6bbdf09329a6fc6c6e862d9a217b07188/static/js/layout.js#L497
and
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/228994874181d60551b15675d1e8b8ca6a299efa/templates/slotViz.html#L409
both adding "in".

My suggestion is to remove it in slotViz.html as `getRelativeTime` is used in many different locations.
```suggestion
        slotVizCountdown.innerHTML = `<b>${hardforkName} Hardfork</b> ${getRelativeTime(luxon.DateTime.fromMillis(hfTs))} | ${slots} slots`
```
(notice that I would also remove the colon so that the info "flows" better)

D13ce, line 409: This always states "slots" (i.e. plural), even when there is only 1 slot left. This is weird as it works properly for seconds.

D13ce (changes requested):
Sorry, I have found issues I should have found earlier.

templates/slotViz.html
MauserBitfly: image

MauserBitfly: removed the in:

MauserBitfly: slot/slots will now be displayed dynamically

D13ce (approved):
lgtm

PR #1949, opened 2023-03-15
Rocketpool Atlas Support

manuelsc:
- Updated dependencies
- Remove various hardcoded configs and instead rely on rocketpool config
- Remove code that can be accessed via smartnode
- Export deposit credit, node_deposit_balance, node_refund_balance, user_deposit_balance, is_vacant & minipool delegate version
- Return new data via api
- Display new data on rocketpool stats page

TODO:
add new data to validators page


D13ce (approved):
lgtm but please note that I was unable to test the exporter

PR #1951, opened 2023-03-15
(BIDS-1769) fix address casing

MauserBitfly:


recy21 (commented):
@MauserBitfly is common.HexToAddress accepting addresses starting with 0x and without? Because I'm not sure you will always get a 0x in FormatAddressLong for address

MauserBitfly:
> 

yes supports with and w/o 0x

utils/eth1.go
```diff
@@ -140,7 +140,7 @@ func formatAddress(address []byte, token []byte, name string, isContract bool, l
 	}
 
 	// setting tooltip & limit name/address if necessary
-	addressString := fmt.Sprintf("0x%x", address)
+	addressString := FixAddressCasing(fmt.Sprintf("0x%x", address))
```
Eisei24, line 143: An obsolete "0x" since you would remove it with `FixAddressCasing` anyway.

Eisei24:
Pages where I found the addresses still all lower case:
- Address page, e.g. https://prater.beaconcha.in/address/0x8b7f0977bb4f0fbe7076fa22bc24aca043583f5e
In every tab the "From" and "To"
- Transaction page, e.g. https://prater.beaconcha.in/tx/0xdeccdccfcdbd4b34c04ca34214abcb4063b057a1ae33a497260a357bb810bcfc
"From" and "To"

utils/utils.go
```diff
@@ -955,7 +956,7 @@ func FormatThousandsEnglish(number string) string {
 // returns two transparent base64 encoded img strings for dark and light theme
 // the first has a black QR code the second a white QR code
 func GenerateQRCodeForAddress(address []byte) (string, string, error) {
-	q, err := qrcode.New(fmt.Sprintf("0x%x", address), qrcode.Medium)
+	q, err := qrcode.New(FixAddressCasing(fmt.Sprintf("0x%x", address)), qrcode.Medium)
```
Eisei24, line 959: An obsolete "0x" since you would remove it with FixAddressCasing anyway.

Eisei24:
> @MauserBitfly is common.HexToAddress accepting addresses starting with 0x and without? Because I'm not sure you will always get a 0x in FormatAddressLong for address



> @MauserBitfly is common.HexToAddress accepting addresses starting with 0x and without? Because I'm not sure you will always get a 0x in FormatAddressLong for address

FYI it actually improved `FormatAddressLong` because previously it only worked if the given parameter does not have the "0x" prefix but now both ways work.

MauserBitfly:
> Pages where I found the addresses still all lower case (all on prater):
> 
> * Address page, e.g. /address/0x8b7f0977bb4f0fbe7076fa22bc24aca043583f5e
>   In every tab the "From" and "To"
> * Transaction page, e.g. /tx/0xdeccdccfcdbd4b34c04ca34214abcb4063b057a1ae33a497260a357bb810bcfc
>   "From" and "To"
> * Internal transaction, e.g. /tx/0xceeecc986e645374032c3a081350e4471f6944165e0f34369337b021c36ffcf7
>   In both tabs are a few addresses
> * Token transaction, e.g. /tx/0x078e7fc0d829e6078d2dfa8d7357c9e3f52de04f460b5d9691a594c54c7f30e5#overview
>   In both tabs are a few addresses
>   Maybe there's also another type/view of transactions that I can't think of right now.
> * Validator deposits, e.g. /validator/305361#deposits
>   The "Copy to clipboard" button
> * The slot transactions, e.g. /slot/5133293#transactions
>   "From" and "To"
> * Same for the corresponding block, e.g. /block/8607286#transactions
> * Token, e.g. /token/0x14dc98ea86b6fc8df7306ccf3c408af4e3174df9?a=0x4e69e09572be229385218c46bc481ab24fd98b6a
>   "From", "To", the "Copy to clipboard" button and if a filter is active like in the link that filter
> * Deposit Leaderboard, /validators/deposit-leaderboard
>   The "Copy to clipboard" button
> * Deposits, /validators/deposits
>   The "Copy to clipboard" button
> * Withdrawals, /validators/withdrawals
>   The "Address" and "New Withdrawal Address"
> 
> For most or maybe all of them the mouseover popup shows the correct casing but text itself does not.

Thanks for the thorough testing ... hope I got them now all ;-)

utils/eth1.go
```diff
@@ -213,15 +220,16 @@ func FormatAddressAsLink(address []byte, name string, verified bool, isContract
...
+			ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/token/%x?a=%s\">✔ %s (%s…%s)</a> %v", token, addressString, name, addressString[:3], addressString[len(addressString)-3:], CopyButton(addressString))
 		} else {
-			ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/token/0x%x?a=0x%x\">%s 0x%x…%x</a> %v", token, address, name, address[:3], address[len(address)-3:], CopyButton(hex.EncodeToString(address)))
+			ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/token/%x?a=%s\">%s %s…%s</a> %v", token, addressString, name, addressString[:3], addressString[len(addressString)-3:], CopyButton(addressString))
 		}
 	} else {
-		ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/token/0x%x?a=0x%x\">0x%x…%x</a> %v", token, address, address[:3], address[len(address)-3:], CopyButton(hex.EncodeToString(address)))
+		ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/token/%x?a=%s\">%s…%s</a> %v", token, addressString, addressString[:3], addressString[len(addressString)-3:], CopyButton(addressString))
```
Eisei24, line 232: Like above [:3]=>[:5].

MauserBitfly: tnx you are right -fixed

utils/eth1.go
```diff
@@ -140,7 +140,13 @@ func formatAddress(address []byte, token []byte, name string, isContract bool, l
 	}
 
 	// setting tooltip & limit name/address if necessary
-	addressString := FixAddressCasing(fmt.Sprintf("0x%x", address))
+
+	addressString := ""
+	if len(address) > 22 || len(address) < 20 { // It seems this function is also used to format other hashes and not onyl 'normal' addresses
```
Eisei24, line 145: Why do you allow a byte size of 21 & 22 for addresses?
The "0x" at the start would only be one byte but also cannot be/is not allowed to be part of `address`.

You can use the function `IsEth1Address` or `IsValidEth1Address` here.

utils/eth1.go
```diff
@@ -193,15 +199,16 @@ func formatAddress(address []byte, token []byte, name string, isContract bool, l
...
 		if verified {
-			ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/address/0x%x\">✔ %s (0x%x…%x)</a> %v", address, name, address[:3], address[len(address)-3:], CopyButton(hex.EncodeToString(address)))
+			ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/address/%s\">✔ %s (%s…%s)</a> %v", addressString, name, addressString[:5], addressString[len(addressString)-3:], CopyButton(addressString))
 		} else {
-			ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/address/0x%x\">%s 0x%x…%x</a> %v", address, name, address[:3], address[len(address)-3:], CopyButton(hex.EncodeToString(address)))
+			ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/address/%s\">%s %s…%s</a> %v", addressString, name, addressString[:5], addressString[len(addressString)-3:], CopyButton(addressString))
 		}
 	} else {
-		ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/address/0x%x\">0x%x…%x</a> %v", address, address[:3], address[len(address)-3:], CopyButton(hex.EncodeToString(address)))
+		ret = fmt.Sprintf("<a class=\"text-monospace\" href=\"/address/%s\">%s…%s</a> %v", addressString, addressString[:5], addressString[len(addressString)-3:], CopyButton(addressString))
```
Eisei24, line 209: address[:3] gives you the first 6 characters of the address (without "0x") so it has to be changed to addressString[:8] since that one is a string.

db/bigtable_eth1.go
```diff
@@ -2475,11 +2475,11 @@ func (bigtable *Bigtable) GetAddressErc721TableData(address string, search strin
 
 	tableData := make([][]interface{}, len(transactions))
 	for i, t := range transactions {
-		from := utils.FormatHash(t.From)
+		from := utils.FormatAddressWithLimits(t.From, "", false, "", 11, 0, false)
 		if fmt.Sprintf("%x", t.From) != address {
 			from = utils.FormatAddressAsLink(t.From, "", false, false)
 		}
-		to := utils.FormatHash(t.To)
+		to := utils.FormatAddressWithLimits(t.To, "", false, "", 11, 0, false)
```
Eisei24, line 2482: This changes the view a bit.
Before the changes FormatHash showed the first and last 2 bytes (so 4 characters)
With the parameter 11 you only show the first and last 3 characters. (11-5 = 6 => 6/2 = 3).
You want 13.

I used /address/0x78b155d7bc284a887f09e028a360a36cc4c4d4c8#erc721Txns to check.

db/bigtable_eth1.go
```diff
@@ -2559,11 +2559,11 @@ func (bigtable *Bigtable) GetAddressErc1155TableData(address string, search stri
...
```
Eisei24, line 2566: The same issue as in the [comment](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/1951/files/2bd8c741533cffed6a7eca31a73020ace94c5d9f..05aaeceedccc39f58282ec26095b389364c9ae2d#r1140068198) above.

utils/eth1.go
MauserBitfly: jea mixed that up - fixed

db/bigtable_eth1.go
MauserBitfly: tnx, fixed

db/bigtable_eth1.go
MauserBitfly: fixed

Eisei24:
Not completely done yet:

- Internal transaction, e.g. /tx/0xceeecc986e645374032c3a081350e4471f6944165e0f34369337b021c36ffcf7
The "Transfer" line still shows lowercase and in the [Internal Transactions] tab the "From" and "To" is too.
- Token transaction, e.g. /tx/0x078e7fc0d829e6078d2dfa8d7357c9e3f52de04f460b5d9691a594c54c7f30e5#overview
The "Tokens Transferred" line

Eisei24 (changes requested):
Some further changes are required, please have a look at them.

db/bigtable_eth1.go
Eisei24: Only "to" was changed, "from" also needs to be done. (again also in the other method too)

MauserBitfly: true that ... fixed

db/bigtable_eth1.go
MauserBitfly: > * erred" line

It seems the data (incl. formatting) is cached therefor the changes don't take effect emmidatly.

You could circumenvent the cache by adding a +"1" after the cacheKey in line 31 of the eth1data.go . Start the app reload. And it should work. Now when you remove the +"1" start and reload it should have the new version in your cache.

Eisei24 (approved):
lgtm

PR #1962, opened 2023-03-16
(BIDS-1310) Add UTC time to all day and week responses

Eisei24:


handlers/api.go
```diff
@@ -1522,7 +1537,34 @@ func ApiValidatorDailyStats(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	rows, err := db.ReaderDb.Query(`
+	result := []struct {
```
D13ce, line 1540: All this and the block starting at line 1601
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/a8c4a970428a6ccc3ead173ce7a7a2d8412a629e/handlers/api.go#L1601

... this is SO much duplicated code and reduces readability a lot. I see where you are coming from, but consider the following approach:

Do everything like before but instead of calling `returnQueryResultsAsArray` at the end, call
```
data, err := utils.SqlRowsToJSON(rows)

if err != nil {
	sendErrorResponse(w, r.URL.String(), "could not parse db results")
	return
}
```
directly and add your new fields (`DayStart` and `DayEnd`) manually.
Then proceed with `sendOKResponse`.

handlers/api.go
```diff
@@ -1649,16 +1722,21 @@ func ApiValidatorIncomeDetailsHistory(w http.ResponseWriter, r *http.Request) {
...
 		for epoch, income := range epochs {
 			responseData = append(responseData, &responseType{
+				Income:         income,
 				Epoch:          epoch,
 				ValidatorIndex: validatorIndex,
-				Week:           epoch / 1575,
-				Income:         income,
+				Week:           epoch / epochsPerWeek,
+				WeekStart:      utils.EpochToTime((epoch / epochsPerWeek) * epochsPerWeek),
+				WeekEnd:        utils.EpochToTime((epoch/epochsPerWeek)*epochsPerWeek + epochsPerWeek),
```
D13ce, line 1739: **Minor:**
```suggestion
				WeekEnd:        utils.EpochToTime(((epoch / epochsPerWeek) + 1) * epochsPerWeek),
```
This is communicates better what you want to do imho.

Alternatively, you could create a new variable `epochsAtStartOfTheWeek` just below
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/a8c4a970428a6ccc3ead173ce7a7a2d8412a629e/handlers/api.go#L1730
and then do something like
```
WeekStart:      epochsAtStartOfTheWeek,
WeekEnd:        epochsAtStartOfTheWeek + epochsPerWeek,
```

Applies to more of your changes.

handlers/api.go
```diff
@@ -212,9 +212,37 @@ func ApiEthStoreDay(w http.ResponseWriter, r *http.Request) {
...
+
+	dayStart := utils.DayToTime(day).Unix()
+	dayEnd := utils.DayToTime(day + 1).Unix()
+
+	rows, err = db.ReaderDb.Query(`
 		SELECT 
-			day, 
+			day,
+			to_timestamp($1) as day_start,
+			to_timestamp($2) as day_end,
```
D13ce, line 245: This is misusing Postgres to marry the query result with your `dayStart` / `dayEnd` calculations from above, right?

I would not do that. Consider the same approach as commented below in line 1540. That way you would even get back to a single db query.

D13ce (changes requested):
I have not tested your changes yet as I left some suggestions that would revert a bit. Please take a look.

handlers/api.go
Eisei24: I preferred creating a new `epochAtStartOfTheWeek` everywhere.

handlers/api.go
Eisei24: I agree, I just thought it was fine since `/api/v1/epoch/{epoch}` does the exact same thing.
Since I rewrote the query result response code to work for everything I also fixed it for `/api/v1/epoch/{epoch}`.

I didn't search through all APIs to see if this could be improved anywhere else since I don't know how to do so efficiently.

handlers/api.go
Eisei24: I rewrote `returnQueryResults` and `returnQueryResultsAsArray` to take functions as parameters (**adjustQueryEntriesFuncs**)that adjust the query results.
This seems future proof to me and avoids the most duplicate code.

Please let me know what you think of this approach.

handlers/api.go
```diff
@@ -251,7 +251,18 @@ func ApiEthStoreDay(w http.ResponseWriter, r *http.Request) {
 	}
 	defer rows.Close()
 
-	returnQueryResults(rows, w, r)
+	dayTime := func(dataEntryMap map[string]interface{}) error {
```
Eisei24, line 254: This same `dayTime` function exists twice now but I didn't feel it was right to have it outside the functions they are in.
What do you think?

handlers/api.go
```diff
@@ -1507,7 +1522,7 @@ func ApiValidatorDailyStats(w http.ResponseWriter, r *http.Request) {
 			sendErrorResponse(w, r.URL.String(), "invalid start_day parameter")
 			return
 		}
-		if start < endDay {
+		if start > endDay {
```
Eisei24, line 1525: This was a bugfix unrelated to the issue that prevented you from entering any "start" value at all.

handlers/api.go
```diff
@@ -3470,25 +3501,37 @@ func getAuthClaims(r *http.Request) *utils.CustomClaims {
...
 	j := json.NewEncoder(w)
 	data, err := utils.SqlRowsToJSON(rows)
 	if err != nil {
 		sendErrorResponse(w, r.URL.String(), "could not parse db results")
 		return
 	}
 
+	err = adjustQueryResults(data, adjustQueryEntriesFuncs...)
+	if err != nil {
+		sendErrorResponse(w, r.URL.String(), "could not parse db results")
```
Eisei24, line 3514: I'm feeling iffy about this error message. Something went wrong while trying to adjust the query entries with the given functions. Does this still count at a "could not parse db results"?

D13ce: No.

```suggestion
		sendErrorResponse(w, r.URL.String(), "could not adjust query results")
```
?

handlers/api.go
D13ce: I understand your decision.

handlers/api.go
D13ce: Keep in mind that this would have to be changed here too:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/817b4c30dfccb52a0821a6106467af78cd5e0766/handlers/api.go#L3531

handlers/api.go
```diff
@@ -3470,25 +3501,37 @@ func getAuthClaims(r *http.Request) *utils.CustomClaims {
 	return claims.(*utils.CustomClaims)
 }
 
-func returnQueryResults(rows *sql.Rows, w http.ResponseWriter, r *http.Request) {
+func returnQueryResults(rows *sql.Rows, w http.ResponseWriter, r *http.Request, adjustQueryEntriesFuncs ...func(map[string]interface{}) error) {
```
D13ce, line 3504: **Minor:** The new parameter is not really self explanatory, maybe add a comment above the function here and here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/817b4c30dfccb52a0821a6106467af78cd5e0766/handlers/api.go#L3521

handlers/api.go
```diff
@@ -302,17 +313,21 @@ func ApiEpoch(w http.ResponseWriter, r *http.Request) {
...
+		FROM epochs WHERE epoch = $1`, epoch)
 	if err != nil {
 		logger.WithError(err).Error("error retrieving epoch data")
 		sendServerErrorResponse(w, r.URL.String(), "could not retrieve db results")
 		return
 	}
 	defer rows.Close()
 
-	returnQueryResults(rows, w, r)
+	epochTime := func(dataEntryMap map[string]interface{}) error {
```
D13ce, line 325: **Minor:** Personally I would name this `addEpochTime`.

IF you change that, keep in mind that the "add" should be added here too:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/817b4c30dfccb52a0821a6106467af78cd5e0766/handlers/api.go#L254
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/817b4c30dfccb52a0821a6106467af78cd5e0766/handlers/api.go#L1573

D13ce (changes requested):

handlers/api.go
Eisei24: Done.

D13ce (approved):
lgtm

Your second way to implement the changes (adding `adjustQueryEntriesFuncs`) is super smooth and makes me feel good about this code review :)

PR #1965, opened 2023-03-16
(BIDS-1313) Fix error fetching contract metadata for address ...: con…

recy21:
…tract abi not found

### DIFF
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -2990,7 +2990,7 @@ func (bigtable *Bigtable) GetContractMetadata(address []byte) (*types.ContractMe
 		ret, err := utils.TryFetchContractMetadata(address)
 
 		if err != nil {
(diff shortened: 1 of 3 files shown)


peterbitfly (commented):
I‘d suggest to adapt the function that it emits a warning if the abi is not available and an error if there was an issue with the api call. With the current solution we won’t spot errors if the api used to fetch the abi data is unavailable.

recy21:
@peterbitfly current (without this PR) solution will always log a error with "contract abi not found", so no distinct between real errors possible. But you are right, switching all of them to warnings will not solve this issue.

I will have a deeper look to distinguish between these two cases and implement a better solution, thanks for pointing this out.
+ I will check the caching in all this cases, as currently the contract is cached in each case, maybe in some cases it would make sense to try it again, a little bit later. So maybe caching it not for 24h in that case and only for 1.

peterbitfly (approved):

PR #1967, opened 2023-03-17
Rocketpool exporter regularly update minipool fee and node timezone

manuelsc:
Fixed an old bug where timezone was constant even though it can change, update it regularly instead of updating it once on detection.

Same applies to minipool fee (commission) starting with Atlas. When a minipool converts from 16 to 8 the commission will be reduced to 14%. Update regularly to reflect those changes.


D13ce (approved):
utACK

PR #1980, opened 2023-03-22
(BIDS-1242) Don't send notifications if it is a machine issue

Eisei24:


Eisei24:
As requested I changed the check using the user ids to instead use the "count of all subscriptions" and the "count of all subscriptions that would be sent" for an event for the calculation.

services/notifications.go
```diff
@@ -2319,6 +2319,13 @@ func collectMonitoringMachine(
 		}
 	}
 
+	// If more than 90% of users would be notified there is an issue and no one should be notified
```
D13ce, line 2322: **Minor:** This comment is inaccurate as it is not "more than 90%" but "at least 90%" :monocle_face: 

Furthermore, "there is an issue" would not let me know what this means if I read this piece of code.

Maybe like this:
```suggestion
	// if at least 90% of users would be notified, we expect an issue on our end and no one will be notified
```

D13ce (approved):
utACK

But I have left some feedback about the comment you have added. Feel free to change or to merge.

services/notifications.go
Eisei24: Sounds good, changed.

PR #1983, opened 2023-03-23
(NOBIDS) check validity of eth2 deposits

peterbitfly:


recy21 (approved):
I would suggest to use "else if", as there is no need to check for all 4 config names, if already a valid name is found. But will merge anyway

PR #1988, opened 2023-03-23
(BIDS-1839) Fix withdrawal estimation for exiting validators

D13ce:


handlers/dashboard.go
```diff
@@ -277,14 +283,23 @@ func getNextWithdrawalRow(queryValidators []uint64) ([][]interface{}, error) {
...
+
 	nextData := make([][]interface{}, 0, 1)
 	nextData = append(nextData, []interface{}{
 		template.HTML(fmt.Sprintf("%v", utils.FormatValidator(nextValidator.Index))),
 		template.HTML(fmt.Sprintf(`<span class="text-muted">~ %s</span>`, utils.FormatEpoch(uint64(utils.TimeToEpoch(timeToWithdrawal))))),
 		template.HTML(fmt.Sprintf(`<span class="text-muted">~ %s</span>`, utils.FormatBlockSlot(utils.TimeToSlot(uint64(timeToWithdrawal.Unix()))))),
 		template.HTML(fmt.Sprintf(`<span class="">~ %s</span>`, utils.FormatTimeFromNow(timeToWithdrawal))),
 		withdrawalCredentialsTemplate,
-		template.HTML(fmt.Sprintf(`<span class="text-muted">~ %s</span>`, utils.FormatAmount(new(big.Int).Mul(new(big.Int).SetUint64(nextValidator.Balance-utils.Config.Chain.Config.MaxEffectiveBalance), big.NewInt(1e9)), "ETH", 6))),
+		template.HTML(fmt.Sprintf(`<span class="text-muted">~ %s</span>`, utils.FormatAmount(new(big.Int).Mul(new(big.Int).SetUint64(withdrawalAmount), big.NewInt(1e9)), "ETH", 6))),
```
Eisei24, line 302: **Unrelated to the bugfix:** Here in the dashboard we show a `~` for the amount but we don't do this for the validator view. Should this be consistent?

D13ce: Good catch, has already been fixed here though: https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2010/files#diff-3ebb39e6ec61a163864798ea1e5af8c31d00e716a2b6cd1625c12336ef8dd281L289

handlers/dashboard.go
```diff
@@ -235,13 +235,19 @@ func getNextWithdrawalRow(queryValidators []uint64) ([][]interface{}, error) {
...
+
+			if (balance[0].Balance > 0 && v.WithdrawableEpoch <= epoch) ||
+				(balance[0].EffectiveBalance == utils.Config.Chain.Config.MaxEffectiveBalance && balance[0].Balance > utils.Config.Chain.Config.MaxEffectiveBalance) {
+				// this validator is eligible for withdrawal, check if it is the next one
+				if nextValidator == nil || v.Index > *stats.LatestValidatorWithdrawalIndex {
+					nextValidator = v
+					nextValidator.Balance = balance[0].Balance
+					if nextValidator.Index > *stats.LatestValidatorWithdrawalIndex {
+						// the first validator after the cursor has to be the next validator
+						break
```
Eisei24, line 252: As discussed this break only leaves the inner for loop and not both so the `nextValidator` still gets overwritten.

D13ce: Good find. Fixed.

Eisei24 (approved):
lgtm

PR #1989, opened 2023-03-24
Add Lodestar to mobile app client monitoring setup guides

nflaig:
Adds lodestar to mobile app client monitoring setup guides.

I tried to align it with how Lighthouse is documented, as technically their setup is the same.

I did not find a quick way to properly test this, so might be worth to check if this renders as expected.

### DIFF
--- a/templates/user/settings.html
+++ b/templates/user/settings.html
@@ -29,6 +29,10 @@
         {
             id: 'teku-graffiti',
             config: "--metrics-publish-endpoint https://beaconcha.in/api/v1/client/metrics?apikey={apiKey}{machineName}"
+        },
+        {
+            id: 'lodestar-graffiti',
+            config: "--monitoring.endpoint 'https://beaconcha.in/api/v1/client/metrics?apikey={apiKey}{machineName}' --metrics"
         }
     ]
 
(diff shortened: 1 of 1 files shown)


nflaig:
Are the quality checks in place to ensure that clients you are listing on your website provide accurate data?

I verified that Lodestar works correctly but would be great if this was additionally confirmed by you.

Buttaa:
> Are the quality checks in place to ensure that clients you are listing on your website provide accurate data?

No, we only display whatever the client sends, so it's totally up to the user/client

manuelsc:
@nflaig Thank you very much for your contribution! I pushed a quick fix to the repo that prevents lighthouse and lodestar to simultaneously open. We used a bad naming scheme there that was easy to misuse, that was on us.

LGTM

manuelsc (approved):

templates/user/settings.html
```diff
@@ -626,6 +630,29 @@ <h2 class="card-header" id="headingFour">
                             </div>
                           </div>
                         </div>
+                        <div class="card" style="box-shadow: none; ">
+                          <h2 class="card-header" id="headingOne">
```
nflaig, line 634: @manuelsc Thanks for the update. Not sure how relevant this is for the id here but `headingOne` is the same as for Lighthouse

manuelsc: Good catch I'll raise a ticket internally to clean this up too. Seems to only have a descriptive impact

PR #2002, opened 2023-03-27
(NOBIDS) fix limit parameter on balancehistory api endpoint

### LINKED ISSUE #2001: [API] The query param `limit` doesn't work as intended `/balancehistory`
**Describe the bug**
The query param `limit` 
https://beaconcha.in/api/v1/validator/__INSERT_KEY__/balancehistory?limit=1

**To Reproduce**
1. Go to `https://beaconcha.in/api/v1/docs/index.html#/Validator/get_api_v1_validator__indexOrPubkey__balancehistory`
2. Set an `indexOrPubkey`
3. Set the query param `limit` to `1`
4. It returns 2 items. Setting the limit to N, the number of results is N + 1

**Expected behavior**
It should return a single item (limit=1)

peterbitfly:
Fixes #2001

### DIFF
--- a/handlers/api.go
+++ b/handlers/api.go
@@ -1952,6 +1952,7 @@ func getBalanceHistoryQueryParameters(q url.Values) (uint64, uint64, error) {
 		if err != nil || limit > defaultLimit || limit < 1 {
 			return 0, 0, fmt.Errorf("invalid limit parameter")
 		}
+		limit = limit - 1
 	}
 
 	return latestEpoch, limit, nil


handlers/api.go
```diff
@@ -1952,6 +1952,7 @@ func getBalanceHistoryQueryParameters(q url.Values) (uint64, uint64, error) {
 		if err != nil || limit > defaultLimit || limit < 1 {
 			return 0, 0, fmt.Errorf("invalid limit parameter")
 		}
+		limit = limit - 1
```
recy21, line 1955: With this fix the function will return 0 as limit, which isn't true, as we want a limit of 1.

I would suggest change
`history, err := db.BigtableClient.GetValidatorBalanceHistory(queryIndices, latestEpoch-limit, latestEpoch)`
to
`history, err := db.BigtableClient.GetValidatorBalanceHistory(queryIndices, latestEpoch-limit+1, latestEpoch)`
instead, as it's the usage of limit which is wrong. 

Can be found here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d0a8c1c3a8b0819932784b2b3358d8721b195e2c/handlers/api.go#L1883

recy21: @peterbitfly fyi

peterbitfly: > With this fix the function will return 0 as limit, which isn't true, as we want a limit of 1.
> 
> I would suggest change `history, err := db.BigtableClient.GetValidatorBalanceHistory(queryIndices, latestEpoch-limit, latestEpoch)` to `history, err := db.BigtableClient.GetValidatorBalanceHistory(queryIndices, latestEpoch-limit+1, latestEpoch)` instead, as it's the usage of limit which is wrong.
> 
> Can be found here:
> 
> https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d0a8c1c3a8b0819932784b2b3358d8721b195e2c/handlers/api.go#L1883

You mean change to `latestEpoch-limit-1`?

recy21: > You mean change to `latestEpoch-limit-1`?

I mean `latestEpoch-(limit-1)`, sorry

peterbitfly: Done

recy21: Hm, which is the same. no I mean `latestEpoch-limit+1`

recy21: `X - (L-1) = X - L + 1` just bad at math early in the morning :)

recy21 (approved):

PR #2003, opened 2023-03-27
(BIDS-1422) return canonical block in case of two blocks at same slot…

peterbitfly:
… height


recy21 (approved):

D13ce:
As discussed with @peterbitfly , this will not be merged. The issue was tackled via a clean-up in the according databases.

PR #2009, opened 2023-03-28
(BIDS-1834) show 0 as expected withdrawal amount during in progress e…

peterbitfly:
…poch


Eisei24 (changes requested):
The dashboard also has to be adjusted.
Currently it still shows the last withdrawal amount instead of 0.

peterbitfly:
Updated the pr to apply the same logic in the dashboard view.

PR #2014, opened 2023-03-29
(NOBIDS) node_jobs: allow only one bls-change-node-job per validator

guybrush:
```sql
CREATE TABLE IF NOT EXISTS
    node_jobs_bls_changes_validators (
        validatorindex INT NOT NULL,
        node_job_id VARCHAR(40) NOT NULL,
        PRIMARY KEY (validatorindex)
        -- we allow only one job per validator
    );
```


peterbitfly:
Why a new table and not just a new column in the `validators` table?

db/node_jobs.go
```diff
@@ -154,21 +163,47 @@ func CreateBLSToExecutionChangesNodeJob(nj *types.NodeJob) (*types.NodeJob, erro
...
-	_, err = WriterDb.Exec(`insert into node_jobs (id, type, status, data, created_time) values ($1, $2, $3, $4, now())`, nj.ID, nj.Type, nj.Status, nj.RawData)
+	tx, err := WriterDb.Beginx()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error starting db transactions: %w", err)
+	}
+	defer tx.Rollback()
+
+	for _, idx := range indicesArr {
+		res, err := tx.Exec(`insert into node_jobs_bls_changes_validators (validatorindex, node_job_id) values($1, $2) on conflict do nothing`, idx, nj.ID)
```
peterbitfly, line 184: This will cause several thousand queries to be executed sequentially. Better to use proper batching here.

peterbitfly (changes requested):

db/node_jobs.go
guybrush: fixed in new e937dab

guybrush:
rebased

PR #2029, opened 2023-03-31
Bids 1850/show link to credentials change guide

Eisei24:


templates/validator/overview.html
```diff
@@ -243,11 +243,15 @@
...
-            <span> {{ .WithdrawalCount }} </span>
+            <i class="fas fa-money-bill"></i>
+            {{ .WithdrawalCount }}
           {{ else }}
-            <span><i class="text-warning fas fa-exclamation-triangle"></i></span>
+            <a href="https://launchpad.ethereum.org/en/withdrawals">
+              <i class="fas fa-money-bill"></i>
+              <i class="text-warning fas fa-exclamation-triangle"></i>
+              <i class="fas fa-external-link-alt"></i>
+            </a>
```
D13ce, line 254: This causes the money-bill to turn blue (as it is a link now). Is that intentional?
Maybe only make the new icon clickable?

Furthermore, I know this has not been part of the spec for this issue, but maybe have a tooltip for the guide itself? Something like `Click here for a guide on this topic` (brainstorm).

Eisei24: Implemented the changes as discussed in the issue.
`class="text-warning"` already forces the colour for the triangle.

D13ce (approved):
lgtm

PR #2039, opened 2023-04-03
(BIDS-1266) Don't load a js for the entire page

Eisei24:


templates/pools/pools.html
```diff
@@ -87,4 +86,5 @@
     </div>
   </section>
   <div id="tempcont" class=""></div>
+  <script src="/js/jquery.min.js"></script>
```
MauserBitfly, line 89: <script src="/js/jquery.min.js"></script> can be removed entirely from this file as it is already included in the layout.html (which wraps this page)

MauserBitfly: The error was caused because the import of the jquery.min overwrote the already loaded jquery instance from the layout.html. 
The version from the layout.html was enhanced with the lib <script src="/js/typeahead.bundle.min.js"></script>.
As this was overwritten, the error ocurred. 

It is just a timing accident that it works when you load the jquery at the end of the pools.html. Pls. remove it!

Eisei24: Removed.

MauserBitfly (approved):

PR #2041, opened 2023-04-03
(BIDS-1911) Fix sync period handling for exiting validators

D13ce:


handlers/api.go
```diff
@@ -1075,15 +1075,13 @@ func getExpectedSyncCommitteeSlots(validators []uint64, epoch uint64) (expectedS
...
+		if vi.ExitEpoch > epoch || vi.ExitEpoch == noEpoch {
 			lastEpoch = epoch
 		}
 
-		{
-			spoe := utils.SyncPeriodOfEpoch(lastEpoch)
-			if _, found := periodInfoMap[spoe]; !found {
-				return 0, fmt.Errorf("required period not found: SyncPeriodOfEpoch(%v) = '%v'", lastEpoch, spoe)
-			}
+		spoe := utils.SyncPeriodOfEpoch(lastEpoch)
```
recy21, line 1082: If you remove the curly braces, spoe will be available outside the scope, for no reason?

D13ce: In this very case we are talking about a harmless number. Protecting it with its own scope seems excessive to me.

Personally I believe that perfectly scoping every single variable leads to unreadable code as more and more intending is added.

recy21: @D13ce in that case it should be:
`
		spoe := utils.SyncPeriodOfEpoch(lastEpoch)
		if entry, found := periodInfoMap[spoe]; found {
			expectedCommitties += entry - periodInfoMap[utils.SyncPeriodOfEpoch(vi.ActivationEpoch)]	
		} else {
			return 0, fmt.Errorf("required period not found: SyncPeriodOfEpoch(%v) = '%v'", lastEpoch, spoe)
		}
	}
`

D13ce: You are right, `spoe` should be used more = Done.

D13ce: Okay, thanks for making me look again. I have improved the code in general now and it should now even be more secure.

Eisei24 (approved):
lgtm

PR #2044, opened 2023-04-04
(NOBIDS) add missing db indices, remove nulls sorting

peterbitfly:
adds missing db indices to improve the performance of the /validators page, removes nulls sorting for lastattestationslot column


D13ce (approved):
utACK

PR #2045, opened 2023-04-04
Bids 1806/add tooltips to rounded values

Eisei24:


utils/eth1.go
```diff
@@ -263,24 +263,25 @@ func FormatAddressLong(address string) template.HTML {
 
 }
 
-func FormatAmountFormated(amount *big.Int, unit string, digits int, maxPreCommaDigitsBeforeTrim int, fullAmountTooltip bool, smallUnit bool, newLineForUnit bool) template.HTML {
+func FormatAmountFormatted(amount *big.Int, unit string, digits int, maxPreCommaDigitsBeforeTrim int, fullAmountTooltip bool, smallUnit bool, newLineForUnit bool) template.HTML {
 	return formatAmount(amount, unit, digits, maxPreCommaDigitsBeforeTrim, fullAmountTooltip, smallUnit, newLineForUnit)
 }
 func FormatAmount(amount *big.Int, unit string, digits int) template.HTML {
-	return formatAmount(amount, unit, digits, 0, false, false, false)
```
Eisei24, line 270: The `fullAmountTooltip` parameter was set to false but not used and the tooltip was always shown.
So now that I actually use it I set the standard to true so that the behaviour stays the same.

D13ce (approved):
lgtm

PR #2051, opened 2023-04-04
(BIDS-1803) search in all tx's for hash

MauserBitfly:


db/bigtable_eth1.go
```diff
@@ -1772,6 +1772,28 @@ func (bigtable *Bigtable) GetAddressesNamesArMetadata(names *map[string]string,
...
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Second*30))
+	defer cancel()
+	key := fmt.Sprintf("%s:TX:%x", bigtable.chainId, txHash)
+	row, err := bigtable.tableData.ReadRow(ctx, key)
+
+	if err != nil {
+		return nil, err
+	}
+	if row == nil {
+		return nil, fmt.Errorf("tx not found %x", txHash)
```
D13ce, line 1785: Why not `return nil, nil` like done, for example, here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/95278d95a21a3c493da5d9c078518bb89b356a76/db/bigtable_eth1.go#L2784

This would make it easier for you to handle the return value.

I honestly would not see this as an error.

handlers/search.go
```diff
@@ -135,11 +135,14 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 				http.Error(w, "Internal server error", http.StatusInternalServerError)
 				return
 			}
-			err = db.ReaderDb.Select(result, `
-				SELECT ENCODE(txhash::bytea, 'hex') AS txhash
-				FROM blocks_transactions
-				WHERE txhash = $1
-				ORDER BY block_slot LIMIT 10`, txHash)
+			tx, dbErr := db.BigtableClient.GetIndexedEth1Transaction(txHash)
```
D13ce, line 138: If you follow my comment for https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2051/files#diff-143a07e5aa31442d87e5dd9584d1525fc98e8b2fef161a08af276d65f60184acR1785, you will not need `dbErr` here. You could use `err` directly.

D13ce (changes requested):
I suggest not treating un-indexed txs as errors, see my comments.

db/bigtable_eth1.go
MauserBitfly: Yea I thought the caller should decide what to do ... but yea let's keep it simple ;-)

handlers/search.go
MauserBitfly: ok

D13ce (approved):
lgtm

PR #2053, opened 2023-04-04
(BIDS-1819) Return CL and EL APR for API

Eisei24:
We round to 16 decimals because the normal APR also gets saved with that precision due to decimals package division.


LuccaBitfly (approved):
lgtm!

PR #2057, opened 2023-04-05
Bids 1695/show fiat value of transactions

Eisei24:
For this issue I don't show the historical values if the transactions is from the current day.
There is also no price info for transactions in the mempool.
The "current price" is not available on testnets so it will always show 0.00.

Please consider possible adaptations.

### DIFF
--- a/db/frontend.go
+++ b/db/frontend.go
@@ -520,35 +520,27 @@ func getMachineStatsGap(resultCount uint64) int {
 	return 1
 }
 
-func GetHistoricPrices(currency string) (map[uint64]float64, error) {
-	data := []struct {
-		Ts       time.Time `db:"ts"`
-		Currency float64   `db:"currency"`
-	}{}
-
+func GetHistoricPrice(currency string, day uint64) (float64, error) {
 	if currency == "ETH" {
 		currency = "USD"
 	}
 	currency = strings.ToLower(currency)
 
 	if currency != "eur" && currency != "usd" && currency != "rub" && currency != "cny" && currency != "cad" && currency != "jpy" && currency != "gbp" && currency != "aud" {
-		return nil, fmt.Errorf("currency %v not supported", currency)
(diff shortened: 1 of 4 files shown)


Mizanur1988 (approved):
I proceed

db/frontend.go
```diff
@@ -539,7 +544,7 @@ func GetHistoricPrices(currency string) (map[uint64]float64, error) {
 	genesisTime := time.Unix(int64(utils.Config.Chain.GenesisTimestamp), 0)
 
 	for _, d := range data {
-		day := uint64(d.Ts.Sub(genesisTime).Hours()) / 24
+		day := uint64(d.Ts.Add(utils.Day).Sub(genesisTime).Hours()) / 24
```
manuelsc, line 547: Slight performance suggestion, rather than retrieving and mapping every historic price as needed, we should either store the map atomically and update it in a service so we don't always have to fetch it from the db.

Or modify the query so we get the historic price directly from the database instead of querying every historic price.

templates/eth1tx.html
```diff
@@ -1,4 +1,19 @@
 {{ define "js" }}
+  <script>
+    function togglePriceContent(currentPrice, historicPrice) {
+      var content = document.getElementById("valuePriceButton")
+      var curValTitle = "Current Value"
+      var estValTitle = "Estimated Value on Day of Txn"
+
+      if (content.getAttribute("data-original-title") === estValTitle) {
```
manuelsc, line 8: Super minor detail I noticed, tooltip won't update until you move the mouse from the element and back to it. Not sure if we can fix this behavior and change the tooltip on click.

db/frontend.go
Eisei24: I rewrote the function/query to only look for a specific day.

templates/eth1tx.html
Eisei24: Fixed.

manuelsc (approved):
lgtm

PR #2071, opened 2023-04-07
(BIDS-1851) Clean up metrics setup IDs

Eisei24:
I couldn't check if everything still looks the same because I can't login on localhost for the "Account Settings" but it should be fine.


D13ce (approved):
lgtm

(FYI, I have also tested it. Hit me up if you want to know how :smiley: )

PR #2079, opened 2023-04-12
Bids 1958/validator performance endpoint

D13ce:


handlers/api.go
```diff
@@ -2020,6 +2020,37 @@ func ApiValidatorPerformance(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// check for recently activated validators that have no performance data yet but already generate income
+	for incomeValidatorIndex := range currentDayIncome {
+		found := false
```
manuelsc, line 2025: I'm not the biggest fan of nested loops. We could improve performance if we optimize the data in `data` a bit.
A first loop that iterates through data and adds the content to a map with index `validatorIndex`.

And then change the already present loop below that iterates over data to iterate over currentDayIncome instead. And retrieve the data from `data` via the new mapped data. That ways it's only one O(n) loop more instead of O(n²).

Alternatively if we want to keep a lower mem footprint, we could also wrap the whole activated validators without performance data in an if. Only execute when len(data) != len(currentDayIncome) so we don't have to do this extra work if we know that all validators are present in both

manuelsc (changes requested):

handlers/api.go
D13ce: Good call. I have chosen something that is similar to your first suggestion.

manuelsc (approved):
lgtm

PR #2088, opened 2023-04-12
(NOBIDS) Don't write an error to log if user uses wrong password / us…

recy21:
…ername

### <samp>🤖 Generated by Copilot at 5888d55</samp>

This pull request improves the error handling and logging of two functions in the `handlers` package: `LoginPost` and `UserUpdatePasswordPost`. It adds checks for the `sql.ErrNoRows` error, which indicates that no user was found with the given email or password.


D13ce (approved):
lgtm

PR #2095, opened 2023-04-13
(BIDS-1960) Add sort field to API

Eisei24:


D13ce (changes requested):
I believe this implementation does not work, here is how to see that:

Use Sepolia and take a look at validator 100: https://sepolia.beaconcha.in/validator/100#blocks
(I use Sepolia as there are very little validators meaning that each validator has proposed many blocks)

Start a local instance and make the following two queries:
```
curl -X 'GET' \
  'http://localhost:8081/api/v1/execution/100/produced?offset=0&limit=10&sort=asc' \
  -H 'accept: application/json'
```
```

curl -X 'GET' \
  'http://localhost:8081/api/v1/execution/100/produced?offset=0&limit=10&sort=asc' \
  -H 'accept: application/json'
```
This will return you the same 10 blocks (the 10 most recent ones), just sorted differently. While this is the expected behavior for "desc", you don't want that to happen for "asc" though:
"asc" should return the 10 oldest blocks sorted from oldest to newest.

In order to achieve that, you will have to change your approach:
Instead of sorting the results you receive from `GetBlocksIndexedMultiple` you have to pass the desired sorting to both `findExecBlockNumbersByFeeRecipient` and `findExecBlockNumbersByProposerIndex` and use it in the Postgres query.

------

Another thing I have noticed while pondering on that:

If you query with a mixture of validator indexes and fee recipient addresses (for example "100,0x670B24610DF99b1685aEAC0dfD5307B92e0cF4d7" on Sepolia) the sorting, even the descending one, should be broken.

Fee recipient addresses are handled first:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/b82c45f5f86ca03548367ff03d0a156d4682c3fe/handlers/api_eth1.go#L171

Validator indexes are handled afterwards:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/b82c45f5f86ca03548367ff03d0a156d4682c3fe/handlers/api_eth1.go#L184

After that, `blockList` contains blocks for fee recipient addresses first and validator indexes last. Since this list is not sorted anywhere anymore, the outcome should be considered "unsorted", right? What I mean by that is, if the fee recipient, for example, has ONLY blocks that are older than the ones for the validator indexes, you would only get blocks for the fee recipients even when sorting descendingly, right? Please confirm this thought.

Here is an example query (again, for Sepolia):
```
curl -X 'GET' \
  'http://localhost:8081/api/v1/execution/100%2C0x670B24610DF99b1685aEAC0dfD5307B92e0cF4d7/produced?offset=0&limit=10&sort=desc' \
  -H 'accept: application/json'
```

If you can confirm that: This topic is not directly related to your changes but would require to take a different approach than mentioned above: Instead of passing the sorting to both `findExecBlockNumbersByFeeRecipient` and `findExecBlockNumbersByProposerIndex` only, you would also have to sort the resulting `blockList`, right?

Eisei24:
The ordering of `blockList` does not matter for our lookup of the data in the bigtable via `GetBlocksIndexedMultiple`.
I can't quite analyze the code of it but the way it functions seems to be that it goes through the bigtable starting with the newest blocks and then checks whether it exists in our `blockList`.
As a result in our `blocks` return value we have the limit-many newest blocks from `blockList` and everything is as intended I guess for descending order.

In your example the blocks for the fee recipient were all older than for the index and so only the blocks for the index are shown, you get the same result as if only searching for the index.

To adjust for ascending order it is not enough to sort `blockList` but you also need to cut it down to the blocks you actually want to search for.

handlers/api_eth1.go
```diff
@@ -192,6 +192,23 @@ func ApiETH1AccountProducedBlocks(w http.ResponseWriter, r *http.Request) {
...
+				allKeys[item] = true
+				list = append(list, item)
+			}
+		}
+		blockList = list
+
+		// Trim to the blocks that are within the limit range
+		sort.Slice(blockList, func(i, j int) bool { return blockList[i] < blockList[j] })
+		blockList = blockList[:limit]
+	}
```
D13ce, line 210: **Minor:** As discussed, personally I think it is confusing to only to this for asc and I would suggest to do it all the time. This seems less error prone in the future, does it not?

D13ce (changes requested):
@Eisei24 , very good analyzes, you are right, this is still problematic.

Here is a summary of our discussion:
We have two sources (validator index and fee recipient address) of blocks to be considered (let's call those sources A and B).
This is problematic for pagination (offset and limit):
- When called with offset 0/limit 10 we just have to query the first 10 blocks from A and B.
- When called with offset 10 (i.e. Page 2)/limit 10 we all of a sudden have to query the first 20 blocks from A and B for the following reason:
It is possible that the first 10 blocks are exclusively coming from A or B. We don't know that though.
- So querying the last 10 entries would mean querying EVERY entry for BOTH A and B.

In all honesty, I would suggest to split up this endpoint into two OR to not allow a mixture of both validator indexes (or addresses) and fee recipient addresses. Both suggestions would break the API though ... which isn't good either.

Please discuss this internally.

D13ce (approved):
lgtm

PR #2106, opened 2023-04-17
(BIDS-440) New Validator Page

LuccaBitfly:
### <samp>🤖 Generated by Copilot at 2fe6d23</samp>

This pull request improves the website's appearance and functionality by modifying the HTML templates, the data types, and the formatting functions. It adds custom styling to some columns and rows, simplifies the layout of the validator page, clarifies the types and comments in `templates.go`, enhances the formatting of hashes, addresses, and other data types, and adds new fields for the validator balance changes in `exporter.go`.


types/templates.go
```diff
@@ -342,6 +342,7 @@ type ValidatorsPageDataValidators struct {
 	Performance7d              int64  `db:"performance7d"`
 }
 
+// ValidatorPageData is a struct to hold data for the validators page
 // ValidatorPageData is a struct to hold data for the validators page
```
Eisei24, line 346: Double comment.

utils/format.go
```diff
@@ -859,87 +843,78 @@ func FormatTimestampTs(ts time.Time) template.HTML {
...
+		return `<span class="text-danger">Slashed</span>`
 	}
-	return fmt.Sprintf(`<span style="font-size: 18px;" class="badge bg-dark text-light mr-1">%s</span>`, tag)
+	return "Unknown"
 }
 
 // FormatValidatorTag will return html formated text of a validator-tag.
 // Depending on the tag it will describe the tag in a tooltip and link to more information regarding the tag.
-func FormatValidatorTag(tag string) template.HTML {
+func formatValidatorTag(tag string) string {
```
Eisei24, line 871: Why change the return type here?
I would rather keep it `template.HTML` here and change the `str` below where it is used also to a `template.HTML`

utils/format.go
```diff
@@ -1277,3 +1252,59 @@ func FormatTimestampUInt64(ts uint64) template.HTML {
 func FormatEth1AddressFull(addr common.Address) template.HTML {
 	return FormatAddress(addr.Bytes(), nil, "", false, false, true)
 }
+
+func FormatHeaderHash(address []byte) template.HTML {
```
Eisei24, line 1256: What is this supposed to do?
Currently it only shows the first 4 and last 4 digits of the validator pubkey in orange (primary).

utils/format.go
```diff
@@ -859,87 +843,78 @@ func FormatTimestampTs(ts time.Time) template.HTML {
...
-	case "ssv":
-		result = `<span style="background-color: rgba(238, 113, 18, .2); font-size: 18px;" class="badge-pill mr-1 font-weight-normal" data-toggle="tooltip" title="Secret Shared Validator"><a style="color: var(--orange);" href="https://github.com/bloxapp/ssv/">SSV</a></span>`
+	switch {
+	case tag == "rocketpool", tag == "pool:rocketpool":
+		result = `<a href="/pools/rocketpool" title="Rocket Pool Validator" data-toggle="tooltip" class="mr-2 badge badge-pill style-bg-neutral-1 style-status-tag-text text-dark">Rocket Pool</a>`
+	case strings.HasPrefix(tag, "pool:"):
+		result = fmt.Sprintf(`<a href="/pools" title="This validator is part of a staking-pool." data-toggle="tooltip" class="mr-2 badge badge-pill style-bg-neutral-1 style-status-tag-text text-dark">Pool: %s</a>`, tag[5:])
+	case strings.HasPrefix(tag, "name:"):
+		result = fmt.Sprintf(`<span title="This name has been set by the owner of this validator." data-toggle="tooltip" class="mr-2 badge badge-pill style-bg-neutral-1 style-status-tag-text text-dark">%s</span>`, tag[5:])
```
Eisei24, line 879: Using `tag[5:]` to trim "pool:" and "name:" is really hard to read and you have to count characters, how about a `strings.TrimPrefix(tag, "pool/name:")`

handlers/validator.go
```diff
@@ -1075,12 +1069,9 @@ func ValidatorProposedBlocks(w http.ResponseWriter, r *http.Request) {
 
 	orderColumn := q.Get("order[0][column]")
 	orderByMap := map[string]string{
-		"0": "epoch",
-		"2": "status",
-		"5": "attestationscount",
-		"6": "depositscount",
-		"8": "voluntaryexitscount",
-		"9": "graffiti",
+		"0": "slot",
```
Eisei24, line 1072: Column 0 is "epoch".
If you would currently click on the arrows to sort by epoch it would sort by slot instead which would coincidentally give the same result I think.

handlers/validator.go
```diff
@@ -1729,32 +1765,60 @@ func ValidatorHistory(w http.ResponseWriter, r *http.Request) {
...
+		} else if syncPenalty > 0 {
+			event.SyncParticipationStatus = 2
+		}
+
+		if incomeDetails[index][i].ProposerSlashingInclusionReward > 0 || incomeDetails[index][i].SlashingReward > 0 {
+			event.SlashingStatus = 1
+		}
+		if incomeDetails[index][i].SlashingPenalty > 0 {
+			event.SlashingStatus = 2
 		}
```
Eisei24, line 1810: Does this work?
I looked at a few validators that got slashed but none showed a `SlashingPenalty` and the icon remained grey.

handlers/validator.go
```diff
@@ -1729,32 +1765,60 @@ func ValidatorHistory(w http.ResponseWriter, r *http.Request) {
 			})
```
Eisei24, line 1765: 1 line above this the `tableData` has 4 entries but shouldn't it be 3 like below?
It probably makes no functional difference.

handlers/validator.go
```diff
@@ -1729,32 +1765,60 @@ func ValidatorHistory(w http.ResponseWriter, r *http.Request) {
...
 			continue
 		}
-		events := template.HTML("")
-		if incomeDetails[index][i].AttestationSourcePenalty > 0 && incomeDetails[index][i].AttestationTargetPenalty > 0 {
-			events += utils.FormatAttestationStatusShort(2)
-		} else {
-			events += utils.FormatAttestationStatusShort(1)
+		event := types.ValidatorHistoryEvent{}
+
+		if incomeDetails[index][i].AttestationSourceReward > 0 {
```
Eisei24, line 1770: Please rename `i` to something more descriptive like `epoch`.

types/templates.go
```diff
@@ -1077,6 +1078,15 @@ type ValidatorHistory struct {
...
 
+type ValidatorDuty struct {
+	SourceAttestationStatus uint64
+	TargetAttestationStatus uint64
+	HeadAttestationStatus   uint64
+	BlockProposalStatus     uint64
+	SyncCommitteeStatus     uint64
+	SlashingStatus          uint64
+}
+
```
Eisei24, line 1089: This is not used and can be deleted.

templates/validator/heading.html
```diff
@@ -1,59 +1,61 @@
...
             {{ if .User.Authenticated }}
-              <span class="mx-1" data-toggle="tooltip" title="Follow this validator and receive email notifications">
-                <button class="btn btn-dark text-white btn-sm" id="follow-button" data-toggle="modal" data-target="#AddValidatorWatchlistModal">
-                  <i class="far fa-bookmark"></i>
-                </button>
-              </span>
+              <div class="dropdown-item px-3 style-paragraph-7" role="button" id="follow-button" data-toggle="modal" data-target="#bookmark-validator-modal">
+                <span class="nav-icon"><i class="fas fa-bookmark"></i></span>
+                Add notification
+              </div>
```
Eisei24, line 37: This does not open the window to add the validator to my watch list.

templates/validator/heading.html
```diff
@@ -1,59 +1,61 @@
...
-            <button class="btn btn-dark text-white btn-sm" type="button" id="copy-button">
-              <i class="fa fa-copy"></i>
-            </button>
-          </span>
-          <span class="mx-1" data-toggle="tooltip" title="Edit validator name">
-            <button class="btn btn-dark text-white btn-sm" type="button" id="edit-button" data-toggle="modal" data-target="#edit-validator-modal">
+          Validator{{ if and (ne .Status "deposited") (ne .Status "deposited_invalid") (ne .Status "deposited_valid") }} {{ .Index }}{{ end }}
+        </h1>
+        {{ if .Tags }}<div>{{ formatValidatorTags .Tags }}</div>{{ end }}
+        {{ if eq .Name "" }}
```
Eisei24, line 10: We only show the edit button if the validator does not have a name yet, is that intentional?
In Figma it is always visible.

Eisei24 (changes requested):

utils/format.go
LuccaBitfly: There is no reason to keep it as template.HTML since that function is not available as template function anymore (it never used as that). In the old code the return value of this function was cast to string anyways so I changed the return type. Won't change unless you insist.

utils/format.go
LuccaBitfly: This function works in tandem with the truncateMiddle function in `layout.js`. In the Figma the public key of the validator gets an ellipsis in the middle when there is not enough space, which is currently not supported by CSS, so I had to write that functionality myself.
Did you see only the first and last 4 digits upon opening the validator page? I would assume you didn't have the right `layout.js` cached, but if it ever happens again please show me.

handlers/validator.go
LuccaBitfly: I already talked about this with @peterbitfly, it seems that CL rewards are calculated here https://github.com/gobitfly/eth-rewards/blob/main/ethrewards.go#L15 but the SlashingReward and SlashindPenalty fields are never set. I think for now it is okay to assume this will work down the line.

handlers/validator.go
```diff
@@ -1729,32 +1765,60 @@ func ValidatorHistory(w http.ResponseWriter, r *http.Request) {
...
 		} else if incomeDetails[index][i].ProposalsMissed > 0 {
-			block := utils.FormatBlockStatusShort(2)
-			events += block
+			event.BlockProposalStatus = 2
+		}
+
+		if syncReward, syncPenalty := incomeDetails[index][i].SyncCommitteeReward, incomeDetails[index][i].SyncCommitteePenalty; syncReward > 0 {
+			if syncPenalty > 0 {
+				event.SyncParticipationStatus = 3
+				balanceChangeForSlot := (syncReward + syncPenalty) / utils.Config.Chain.Config.SlotsPerEpoch
```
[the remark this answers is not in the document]
LuccaBitfly: I did test this somewhat thoroughly, and while the sync rewards and penalties are not always the same, I've found that the floor of the ratio of penalty to reward times the slots per epoch will result in the correct amount of missed slots every time, but really that's just conjecture. The alternative is to fetch the correct amount of missed slots from SQL, but I feel like that could be too much of a performance downgrade. What are your thoughts?

types/templates.go
LuccaBitfly: fixed in new commit, ptal

handlers/validator.go
```diff
@@ -1118,18 +1110,62 @@ func ValidatorProposedBlocks(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		slotToExecBlock = make(map[uint64]*types.Eth1BlockIndexed)
+		for _, execBlock := range execBlocks {
+			slotToExecBlock[utils.TimeToSlot(uint64(execBlock.Time.Seconds))] = execBlock
+		}
+
+		relaysData, err = db.GetRelayDataForIndexedBlocks(execBlocks)
+		if err != nil {
+			logger.Errorf("error retrieving mev bribe data from bigtable: %v", err)
```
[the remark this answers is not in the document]
LuccaBitfly: fixed in new commit, ptal

handlers/validator.go
LuccaBitfly: fixed in new commit, ptal

types/templates.go
LuccaBitfly: oversight sorry, fixed in new commit, ptal

templates/validator/heading.html
LuccaBitfly: fixed in new commit, ptal

templates/validator/heading.html
LuccaBitfly: Yes this is intentional. The name can still be changed in the "More" menu.

utils/format.go
Eisei24: Yes, it was just that, e.g. "0xabcdabcd". Like you said it was probably a cache issue.

handlers/validator.go
Eisei24: We could move the "order arrows" from "Epoch" to "Slot" but we can also just leave it as it is since when your sort by slot you also sort by epoch.

handlers/validator.go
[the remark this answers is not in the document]
Eisei24: As discussed this calculation does not seem to work when there is a large percentage of missed syncs (e.g. 2/32).

I tested it on the prater network with validator 210684:

- Epoch 170305 shows 2/32 but should be 6/32
- Epoch 170309 shows 2/32 but should be 5/32

Eisei24 (commented):

handlers/validator.go
LuccaBitfly: Leaving it as is :)

handlers/validator.go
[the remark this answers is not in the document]
LuccaBitfly: I have now implemented it so that the number of proposed slots get fetched from the database if there are any sync rewards or penalties present in the set of epochs in the queried history page. With these, rewards and penalties are always the same.

Eisei24 (approved):
lgtm!

PR #2113, opened 2023-04-18
(BIDS-1746) Add Withdrawal webhook option

Eisei24:


LuccaBitfly:
The BIDS description says to remove the withdrawal checkbox from the `/user/notifications` view but it is still there.

LuccaBitfly:
Addendum to my last comment: the BIDS description also says to disable this feature for e-mail notifications. Am I correct in assuming that you did not disable the possibility e-mail notifications to include withdrawals? Regardless of whether the checkbox is available for the user to change.

LuccaBitfly (commented):
pls ignore this comment

handlers/user.go
```diff
@@ -2455,7 +2455,7 @@ func NotificationWebhookPage(w http.ResponseWriter, r *http.Request) {
```
LuccaBitfly, line 2455: The BIDS description says to remove the withdrawal checkbox from the /user/notifications view but it is still there. The description also says to disable the feature for e-mail notifications. Am I correct in assuming that you did not remove the possibility for e-mails to include withdrawals? Irregardless of whether the checkbox is available for the user to change.

Eisei24: Right, I at first removed the checkbox but then had to revert is again, see the last comment in the BIDS.

You currently can't disable email notifications for specific notifications just for all in general.
And the validators set for those email notifications are then used for the webhooks.

So we decided to only add the webhooks option for now.

LuccaBitfly (approved):
lgtm

PR #2114, opened 2023-04-18
(BIDS-1978) fix attestations + reward hist. for exiting vali.

MauserBitfly:
### <samp>🤖 Generated by Copilot at dfe889f</samp>

Fix validator history bugs for exited validators. Update `handlers/validator.go` to handle edge cases where the validator's `exit_epoch` is before the current or finalized epoch.

### DIFF
--- a/handlers/validator.go
+++ b/handlers/validator.go
@@ -1628,7 +1628,7 @@ func ValidatorHistory(w http.ResponseWriter, r *http.Request) {
 	// Every validator is scheduled to issue an attestation once per epoch
 	// Hence we can calculate the number of attestations using the current epoch and the activation epoch
 	// Special care needs to be take for exited and pending validators
-	if activationAndExitEpoch.ExitEpoch != 9223372036854775807 && activationAndExitEpoch.ExitEpoch < services.LatestFinalizedEpoch() {
+	if activationAndExitEpoch.ExitEpoch != 9223372036854775807 && activationAndExitEpoch.ExitEpoch <= services.LatestFinalizedEpoch() {
 		totalCount += activationAndExitEpoch.ExitEpoch - activationAndExitEpoch.ActivationEpoch
 	} else {
 		totalCount += services.LatestFinalizedEpoch() - activationAndExitEpoch.ActivationEpoch + 1
--- a/handlers/validator.go
+++ b/handlers/validator.go
@@ -366,7 +366,7 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 		validatorPageData.AttestationsCount = 0
 	}
 
-	if validatorPageData.ExitEpoch != 9223372036854775807 {
+	if validatorPageData.ExitEpoch != 9223372036854775807 && validatorPageData.ExitEpoch < validatorPageData.Epoch {
 		validatorPageData.AttestationsCount = validatorPageData.ExitEpoch - validatorPageData.ActivationEpoch
 	}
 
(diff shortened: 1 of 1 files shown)


handlers/validator.go
```diff
@@ -1177,7 +1177,7 @@ func ValidatorAttestations(w http.ResponseWriter, r *http.Request) {
 		totalCount = 0
 	}
 	lastAttestationEpoch := epoch
-	if ae.ExitEpoch != 9223372036854775807 {
+	if ae.ExitEpoch != 9223372036854775807 && ae.ExitEpoch < epoch {
 		lastAttestationEpoch = ae.ExitEpoch
```
Eisei24, line 1181: As discussed this would include the exit epoch in the attestations table for which no data exists.
As a result the first page for all validators that have already existed show only 9 entries.
Further pages are fine.

Maybe this change is enough
```suggestion
		lastAttestationEpoch = ae.ExitEpoch - 1
```

handlers/validator.go
```diff
@@ -1177,7 +1177,7 @@ func ValidatorAttestations(w http.ResponseWriter, r *http.Request) {
...
 		totalCount = ae.ExitEpoch - ae.ActivationEpoch
 	}
```
Eisei24, line 1183: As discussed this would include the exit epoch in the attestations table for which no data exists from `GetValidatorAttestationHistory`.
As a result the first page for all validators that have already existed show only 9 entries.
Further pages are fine.

```suggestion
	if ae.ExitEpoch != 9223372036854775807 && ae.ExitEpoch <= epoch {
		lastAttestationEpoch = ae.ExitEpoch - 1
		totalCount = lastAttestationEpoch - ae.ActivationEpoch
	}
```

Here I am assuming for the count that you don't have any attestations in the ActivationEpoch.
If you do then +1.

MauserBitfly: Thanks addapted the` <=` and `-1`  ... but I think `totalCount = ae.ExitEpoch - ae.ActivationEpoc` is correct

handlers/validator.go
```diff
@@ -366,7 +366,7 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 		validatorPageData.AttestationsCount = 0
 	}
 
-	if validatorPageData.ExitEpoch != 9223372036854775807 {
+	if validatorPageData.ExitEpoch != 9223372036854775807 && validatorPageData.ExitEpoch < validatorPageData.Epoch {
```
Eisei24, line 369:
```suggestion
	if validatorPageData.ExitEpoch != 9223372036854775807 && validatorPageData.ExitEpoch <= validatorPageData.Epoch {
```

MauserBitfly: yes, makes sense

handlers/validator.go
```diff
@@ -1628,7 +1628,7 @@ func ValidatorHistory(w http.ResponseWriter, r *http.Request) {
 	// Every validator is scheduled to issue an attestation once per epoch
 	// Hence we can calculate the number of attestations using the current epoch and the activation epoch
 	// Special care needs to be take for exited and pending validators
-	if activationAndExitEpoch.ExitEpoch != 9223372036854775807 {
+	if activationAndExitEpoch.ExitEpoch != 9223372036854775807 && activationAndExitEpoch.ExitEpoch < services.LatestFinalizedEpoch() {
```
Eisei24, line 1631: I think it's also `<= ` here.

MauserBitfly: yes, I take it ;-)

Eisei24 (changes requested):
Please have a look at my comments which concern the first attestation page and the case when the exit epoch is equal to the current epoch.

Eisei24 (approved):
lgtm

PR #2120, opened 2023-04-20
Update rocketpool deposit type

manuelsc:
### <samp>🤖 Generated by Copilot at 786f83e</samp>

Refactored minipool deposit type update in `exporter/rocketpool.go`. The code now uses an async contract call to get the correct deposit type for each minipool.


D13ce (approved):
utACK

Issue #1524, opened 2022-09-06
FR: Historical graffiti wall data


benv666:
Currently it is not possible to retrieve data of pixels that have been overwritten.
However, while contemplating making a timelapse video of the Graffiti Wall I realized that this would be needed.
(otherwise it will only show the drawing of the current state, which would skip any "turf wars" that might have happened)

So for this feature request, please implement an API endpoint to retrieve the historical pixeldata.
This could be an option to the current graffitiwall API operation. Bonus point when first implementing https://github.com/gobitfly/eth2-beaconchain-explorer/issues/1523 so we can focus on specific areas.

Buttaa:
Hi ben, currently we don't have any resources for this integration. If you ever open a PR for it, we're happy to merge it

PR #2133, opened 2023-04-23
Add graffitiwall pixel history

RomiRand:
This PR is meant to improve api requests related to the graffitiwall, mainly by implementing #1524.

I think you moved to google's BigTable which I've no access to. So I'm not sure how to properly apply the table definition changes, but you'll find them in the migrations file.

Also I dislike the 10,000 slot limit which has been introduced while implementing #1523. To recall, after this change only the graffiti pixels which have been drawn during the specified 10,000 slot window (and which also have not been overwritten at *any* later slot) are returned. So in order to reproduce the current graffitiwall status from the api, you'd need to query the entire time span of all slots. At the moment that translates to roughly 629 api calls. I understand the slot limit is a way to minimize load, but I think it was better like before (where you could accomplish the same task with just one api call).

Now to fix this I suggest letting the user define a custom slot window (start - end). If you want to keep the window, at least offer a dedicated way to just pull the latest state of the wall as that reduces bandwidth (since overwritten pixels are not sent any more, only the latest ones) and I assume that's what most people like to use anyways.

I've done a combination of both which I liked the most: I added the start field `startSlot`, but also defined a new boolean field called `summarize`. When set to `true` it only returns the end state of all pixels drawn in the specified time window, when set to `false` it returns all pixel updates (=pixel/turf wars).

I could only do some basic mock testing because it takes quite some effort to set everything up properly, hope that's okay.

### DIFF
--- a/db/migrations/20230330131622_init_database.sql
+++ b/db/migrations/20230330131622_init_database.sql
@@ -470,13 +470,15 @@ CREATE TABLE IF NOT EXISTS
 
 CREATE TABLE IF NOT EXISTS
     graffitiwall (
+        slot INT NOT NULL,
         x INT NOT NULL,
         y INT NOT NULL,
         color TEXT NOT NULL,
-        slot INT NOT NULL,
         VALIDATOR INT NOT NULL,
-        PRIMARY KEY (x, y)
+        PRIMARY KEY (slot)
     );
+
+CREATE INDEX IF NOT EXISTS idx_graffitiwall_coordinates ON graffitiwall (x, y);
 
 CREATE TABLE IF NOT EXISTS
     eth1_deposits (
--- a/db/migrations/20230330131622_init_database.sql
+++ b/db/migrations/20230330131622_init_database.sql
@@ -470,15 +470,13 @@ CREATE TABLE IF NOT EXISTS
 
 CREATE TABLE IF NOT EXISTS
     graffitiwall (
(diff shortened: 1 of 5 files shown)


db/migrations/20230330131622_init_database.sql
Eisei24, line 1: When applying changes to the database they must be written in a new file with the current day & time.

RomiRand: Okay checked it out, should be done

Buttaa:
hi @RomiRand , based on a comment by our devs, this PR seems to break the current API integration, which would be a deal breaker. 

> These changes give more option for the /api/v1/graffitiwall API but they also break the current usage.
There are 2 issues:
slot parameter renamed to endSlot.
Before slot was at minimum 10000, now endSlot can be any value (>=0).
This does not break anything but does change the return values.

We'd love to merge it if it does not break the current integration!

RomiRand:
Hey @Buttaa, reasoning for the (semantic) `slot` query parameter changes was that the existence of a minimum doesn't feel very logical to me (at least it could be either documented better and/or an error could be returned if the user requests a slot < 10000).
However I can see how it'd be changing results, so that's rather a separate issue then - applied the suggestions :)

recy21:
@RomiRand thanks, merged! Is now part of Sepolia and will be part of next Release

PR #2134, opened 2023-04-24
(BIDS-1312) Change email notification title and fix type

Eisei24:
### <samp>🤖 Generated by Copilot at f6f2c02</samp>

This pull request fixes several typos in the code that affect the notification system. It corrects the spelling of `event` and `collateral` in various functions and constants, and enhances the email notification subject lines. The changes improve the reliability and clarity of the notification system.

### DIFF
--- a/services/notifications.go
+++ b/services/notifications.go
@@ -727,15 +727,17 @@ func queueEmailNotifications(notificationsByUserID map[uint64]map[types.EventNam
 					metrics.NotificationsQueued.WithLabelValues("email", string(event)).Inc()
 				}
 
-				if len(notificationTitles) > 2 {
-					subject = fmt.Sprintf("%s: %s,... and %d other notifications", utils.Config.Frontend.SiteDomain, notificationTitles[0], len(notificationTitles)-1)
-				} else {
-					subject = fmt.Sprintf("%s: %s and %s", utils.Config.Frontend.SiteDomain, notificationTitles[0], notificationTitles[1])
-				}
-
 				if event == "validator_balance_decreased" {
 					msg.Body += template.HTML("<br>You will not receive any further balance decrease mails for these validators until the balance of a validator is increasing again.<br>")
 				}
+			}
+
+			if len(notificationTitles) > 2 {
+				subject = fmt.Sprintf("%s: %s,... and %d other notifications", utils.Config.Frontend.SiteDomain, notificationTitles[0], len(notificationTitles)-1)
+			} else if len(notificationTitles) == 2 {
+				subject = fmt.Sprintf("%s: %s and %s", utils.Config.Frontend.SiteDomain, notificationTitles[0], notificationTitles[1])
+			} else {
+				subject = fmt.Sprintf("%s: %s", utils.Config.Frontend.SiteDomain, notificationTitles[0])
 			}
 
 			// msg.Body += template.HTML(fmt.Sprintf("<br>Best regards<br>\n%s", utils.Config.Frontend.SiteDomain))
--- a/services/notifications.go
+++ b/services/notifications.go
@@ -301,14 +301,14 @@ func collectNotifications(epoch uint64) (map[uint64]map[types.EventName][]types.
 			}
 			logger.Infof("collecting rocketpool claim round took: %v\n", time.Since(start))
 
-			err = collectRocketpoolRPLCollateralNotifications(notificationsByUserID, types.RocketpoolColleteralMaxReached, epoch)
+			err = collectRocketpoolRPLCollateralNotifications(notificationsByUserID, types.RocketpoolCollateralMaxReached, epoch)
 			if err != nil {
 				metrics.Errors.WithLabelValues("notifications_collect_rocketpool_rpl_collateral_max_reached").Inc()
 				return nil, fmt.Errorf("error collecting rocketpool max collateral: %v", err)
 			}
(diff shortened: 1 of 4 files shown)


types/frontend.go
```diff
@@ -46,8 +46,8 @@ const (
 	TaxReportEventName                               EventName = "user_tax_report"
 	RocketpoolCommissionThresholdEventName           EventName = "rocketpool_commision_threshold"
 	RocketpoolNewClaimRoundStartedEventName          EventName = "rocketpool_new_claimround"
-	RocketpoolColleteralMinReached                   EventName = "rocketpool_colleteral_min"
-	RocketpoolColleteralMaxReached                   EventName = "rocketpool_colleteral_max"
```
Eisei24, line 50: I did not change the event names here because then we would have to adjust the db.

services/notifications.go
```diff
@@ -639,21 +639,6 @@ func queueEmailNotifications(notificationsByUserID map[uint64]map[types.EventNam
...
 		}
 		go func(userEmail string, userNotifications map[types.EventName][]types.Notification) {
-			notification := ""
-			othernotifications := ""
-			i := 0
-			for notificationEvent := range userNotifications {
-				if i == 0 {
-					notification = string(notificationEvent)
-				} else if i == 1 {
-					othernotifications = fmt.Sprintf(" and %s", notificationEvent)
```
Eisei24, line 649: While working on this I also fixed two other small bugs.
The first here, this line is never used since you check for `i > 1` below it and overwrite `othernotifications`.

services/notifications.go
```diff
@@ -639,21 +639,6 @@ func queueEmailNotifications(notificationsByUserID map[uint64]map[types.EventNam
...
-				}
-				i++
-			}
-			if i > 1 {
-				othernotifications = fmt.Sprintf(",... and %d other notifications", i)
```
Eisei24, line 654: The second here, it is weird to say "and 5 others" if it is 5 notifications in total.

Eisei24:
When reviewing this please also check that the linked BIDS-1605 is also fixed.

services/notifications.go
```diff
@@ -731,6 +726,13 @@ func queueEmailNotifications(notificationsByUserID map[uint64]map[types.EventNam
 
 					metrics.NotificationsQueued.WithLabelValues("email", string(event)).Inc()
 				}
+
+				if len(notificationTitles) > 2 {
+					subject = fmt.Sprintf("%s: %s,... and %d other notifications", utils.Config.Frontend.SiteDomain, notificationTitles[0], len(notificationTitles)-1)
+				} else {
+					subject = fmt.Sprintf("%s: %s and %s", utils.Config.Frontend.SiteDomain, notificationTitles[0], notificationTitles[1])
+				}
```
D13ce, line 734: This code does not work if there is only one notification title, does it?

Also, it should not be part of the `for` loop it is currently part of, right? It should be done after the `for`.

services/notifications.go
```diff
@@ -673,6 +661,13 @@ func queueEmailNotifications(notificationsByUserID map[uint64]map[types.EventNam
 				msg.Body += template.HTML(fmt.Sprintf("%s<br>====<br><br>", types.EventLabel[event_title]))
 				unsubURL := "https://" + utils.Config.Frontend.SiteDomain + "/notifications/unsubscribe"
 				for i, n := range ns {
+					// Find all unique notification titles for the subject
+					title := n.GetTitle()
+					if _, ok := notificationTitlesMap[title]; !ok {
+						notificationTitlesMap[title] = true
+						notificationTitles = append(notificationTitles, title)
+					}
```
D13ce, line 669: This triggers slightly different behavior than before:

Since the title for ETH Client updates is unique based on the client
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/f6f2c024da99729047f3724e67f7a7693d52ab06/services/notifications.go#L2111
this would now add 3 notifications to the mail's subject if 3 clients are updated.

Is this intended?

If not, maybe use n.GetEventName() to guarantee uniqueness but use the title within the subject. I did NOT test this thought though and honestly I think it's fine to add all 3 clients to the subject in this scenario as the subject gets shortened anyways.

D13ce (changes requested):

services/notifications.go
Eisei24: Right, fixed.

Also fixed. I thought it already was outside the loop.

services/notifications.go
Eisei24: I compared this with the push notification implementation, see here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/f6f2c024da99729047f3724e67f7a7693d52ab06/services/notifications.go#L517
where we have the same logic and just get every title:
```
for event, ns := range userNotifications {
	for _, n := range ns {
		added := false
		for _, userToken := range userTokens {
			notification := new(messaging.Notification)
			notification.Title = fmt.Sprintf("%s%s", getNetwork(), n.GetTitle())
		}
	}
}
```

Eisei24: I think we pondered the same think. During the implementation I thought about the validator_is_offline EventName. It can also have multiple titles:
- Validator is Offline
- Validator Back Online
so your email subject could be "Validator is Offline and Validator Back Online" which made sense to me.

D13ce (approved):
lgtm

PR #2135, opened 2023-04-24
(NOBIDS) Fix Sigsegv possibilities

recy21:
### <samp>🤖 Generated by Copilot at 2f7566c</samp>

This pull request improves the robustness and clarity of the Bigtable queries for eth1 data. It adds nil checks to prevent potential crashes or incorrect results, and removes unused code from `db/bigtable_eth1.go`.

### DIFF
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -2102,8 +2102,10 @@ func (bigtable *Bigtable) GetAddressInternalTableData(address []byte, search str
 
 	names := make(map[string]string)
 	for _, t := range transactions {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
+		if t != nil {
+			names[string(t.From)] = ""
+			names[string(t.To)] = ""
+		}
 	}
 	names, _, err = BigtableClient.GetAddressesNamesArMetadata(&names, nil)
 	if err != nil {
@@ -2113,8 +2115,12 @@ func (bigtable *Bigtable) GetAddressInternalTableData(address []byte, search str
 	tableData := make([][]interface{}, len(transactions))
 	for i, t := range transactions {
 
-		fromName := names[string(t.From)]
-		toName := names[string(t.To)]
+		fromName := ""
+		toName := ""
+		if t != nil {
+			fromName = names[string(t.From)]
+			toName = names[string(t.To)]
+		}
 
 		from := utils.FormatAddress(t.From, nil, fromName, false, false, !bytes.Equal(t.From, address))
 		to := utils.FormatAddress(t.To, nil, toName, false, false, !bytes.Equal(t.To, address))
@@ -2178,8 +2184,10 @@ func (bigtable *Bigtable) GetInternalTransfersForTransaction(transaction []byte,
 
 	names := make(map[string]string)
 	for _, t := range transfers {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
+		if t != nil {
+			names[string(t.From)] = ""
+			names[string(t.To)] = ""
+		}
 	}
 
 	err = bigtable.GetAddressNames(names)
@@ -2252,9 +2260,11 @@ func (bigtable *Bigtable) GetArbitraryTokenTransfersForTransaction(transaction [
 	tokensToAdd := make(map[string]*types.ERC20Metadata)
 	// init
 	for _, t := range transfers {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
-		tokens[string(t.TokenAddress)] = nil
+		if t != nil {
+			names[string(t.From)] = ""
+			names[string(t.To)] = ""
+			tokens[string(t.TokenAddress)] = nil
+		}
 	}
 	g := new(errgroup.Group)
 	g.SetLimit(25)
@@ -2382,17 +2392,16 @@ func (bigtable *Bigtable) GetAddressErc20TableData(address []byte, search string
 	names := make(map[string]string)
 	tokens := make(map[string]*types.ERC20Metadata)
 	for _, t := range transactions {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
-		tokens[string(t.TokenAddress)] = nil
+		if t != nil {
+			names[string(t.From)] = ""
+			names[string(t.To)] = ""
+			tokens[string(t.TokenAddress)] = nil
+		}
 	}
 	names, tokens, err = BigtableClient.GetAddressesNamesArMetadata(&names, &tokens)
 	if err != nil {
 		return nil, err
 	}
-
-	// fromName := names[string(t.From)]
-	// toName := names[string(t.To)]
 
 	tableData := make([][]interface{}, len(transactions))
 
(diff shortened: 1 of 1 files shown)


db/bigtable_eth1.go
```diff
@@ -2102,8 +2102,10 @@ func (bigtable *Bigtable) GetAddressInternalTableData(address []byte, search str
 
 	names := make(map[string]string)
 	for _, t := range transactions {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
+		if t != nil {
```
D13ce, line 2105: I require some information here:

Ranging over `transactions` coming from `GetEth1ItxForAddress` cannot return a `t` that is `nil`, can it? `GetEth1ItxForAddress` includes the following check:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L2083-L2087

The same check should be done in `GetEth1TxForAddress`, `GetEth1ERC20ForAddress` and `GetEth1TxForToken` too: 
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L1727
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L2375
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L3278

Right?

If you agree, some of the `if t != nil {` within `db/bigtable_eth1.go` should be obsolete (the ones related to the functions mentioned).

Please check and share your thoughts.

db/bigtable_eth1.go
```diff
@@ -2252,9 +2260,11 @@ func (bigtable *Bigtable) GetArbitraryTokenTransfersForTransaction(transaction [
 	tokensToAdd := make(map[string]*types.ERC20Metadata)
 	// init
 	for _, t := range transfers {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
-		tokens[string(t.TokenAddress)] = nil
+		if t != nil {
```
D13ce, line 2263: Here the same comment as for line 2187 applies.

db/bigtable_eth1.go
```diff
@@ -2382,18 +2392,17 @@ func (bigtable *Bigtable) GetAddressErc20TableData(address []byte, search string
 	names := make(map[string]string)
 	tokens := make(map[string]*types.ERC20Metadata)
 	for _, t := range transactions {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
-		tokens[string(t.TokenAddress)] = nil
+		if t != nil {
```
D13ce, line 2395: Here the same comment as for line 2105 applies.

db/bigtable_eth1.go
```diff
@@ -3291,9 +3300,11 @@ func (bigtable *Bigtable) GetTokenTransactionsTableData(token []byte, address []
...
```
D13ce, line 3303: Here the same comment as for line 2105 applies.

D13ce (commented):
I would prefer you to take another approach to fix this. Please take a look at my comments and share your thoughts.

recy21:
@D13ce I might agree with your comments (that the checks might not be necessary, because...), but would be an additional check that bad? When you copy this part to a new position in code, where it's using another function with not precheck the return values, it would crash. So taking care about the integrity of your values isn't wrong. We had some fatals with similar lines during the last month, I just wanna be sure.

D13ce:
@recy21 , okay I will make another review based on that pov.

db/bigtable_eth1.go
D13ce, line 2105: A check similar to this one is also required here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/cf3d27e3ce7d3f94bba0a78e1387afe7fbb0d602/handlers/api_eth1.go#L506
(uses `GetEth1ItxForAddress` too)

Also, here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/cf3d27e3ce7d3f94bba0a78e1387afe7fbb0d602/handlers/api_eth1.go#L424
(uses `GetEth1TxForAddress` which is similar)

db/bigtable_eth1.go
```diff
@@ -2102,8 +2102,10 @@ func (bigtable *Bigtable) GetAddressInternalTableData(address []byte, search str
 
 	names := make(map[string]string)
 	for _, t := range transactions {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
+		if t != nil {
+			names[string(t.From)] = ""
+			names[string(t.To)] = ""
+		}
```
D13ce, line 2108: This check does not suffice. A few lines below, `t.From` and `t.To` are accessed again.

By the way, this also applies to the check here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L1828
This wasn't added by you but has the same problem.

db/bigtable_eth1.go
D13ce, line 3303: This check does not suffice, `t` is accessed a few times below without protection, see here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L3318

db/bigtable_eth1.go
```diff
@@ -2113,8 +2115,12 @@ func (bigtable *Bigtable) GetAddressInternalTableData(address []byte, search str
 	tableData := make([][]interface{}, len(transactions))
 	for i, t := range transactions {
 
-		fromName := names[string(t.From)]
-		toName := names[string(t.To)]
+		fromName := ""
+		toName := ""
+		if t != nil {
```
D13ce, line 2120: This check does not suffice as `t` is accessed below anyways.

https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L1828
has the same issue.

db/bigtable_eth1.go
```diff
@@ -2178,8 +2184,10 @@ func (bigtable *Bigtable) GetInternalTransfersForTransaction(transaction []byte,
 
 	names := make(map[string]string)
 	for _, t := range transfers {
-		names[string(t.From)] = ""
-		names[string(t.To)] = ""
+		if t != nil {
```
D13ce, line 2187: This should not suffice as the same `transfers` are accessed here anyways again:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L2210

db/bigtable_eth1.go
D13ce, line 2263: This should not suffice as the same `transfers` are accessed here anyways again:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L2315

D13ce (changes requested):
First of all, here are some (not all?) parts that are still lacking protection:

https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L2410

https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L3318

---

Overall, still after discussing this with you, I think adding checks that ensure that `nil` is not part of the return value as done here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2f7566c921508e73a435184ee5d86c78258e95f3/db/bigtable_eth1.go#L2084
would be the better approach than checking the return values of similar functions over and over again.

D13ce:
@recy21 , as discussed, we will not merge these changes. A new PR will created with a different approach.

PR #2150, opened 2023-04-27
Bids 1316/page and rows wrong in data table state save

Eisei24:
When doing the review please read all the comments in the BIDS since the original bug description is no longer valid.


D13ce (approved):
lgtm

PR #2154, opened 2023-04-30
(NOBIDS) Update stkaing services

recy21:
### <samp>🤖 Generated by Copilot at 16bf5a6</samp>

This pull request adds a new feature to the staking services page that displays the client diversity of each service. It also updates and expands the data for the existing and new services in the `templates/stakingServices.html` file.


Buttaa (approved):
tested in browser, lgtm

PR #2173, opened 2023-05-04
(BIDS-1981) Add quotation marks around monitoring URL

one-three-three-seven:
The monitoring endpoint URL should be in quotation marks because some terminals might cut off after the special characters "&" or "?".


D13ce (approved):
lgtm

PR #2174, opened 2023-05-04
Bids 1965/validator not found 505

KostasGiannakakis:
### <samp>🤖 Generated by Copilot at 6657f43</samp>

This pull request enhances the error handling and user experience for the validator page and the depositing process. It adds custom templates and messages for the cases when the validator or the eth1 transaction is not found, and links to the knowledge base articles. It also removes an unused dependency from the `go.mod` file.


handlers/validator.go
```diff
@@ -1785,7 +1802,12 @@ func ValidatorStatsTable(w http.ResponseWriter, r *http.Request) {
...
-			http.Error(w, "Internal server error", http.StatusInternalServerError)
+			SetPageDataTitle(data, "Validator not found")
+			data := InitPageData(w, r, "validators", fmt.Sprintf("/validator/%v", vars["index"]), "", validatorNotFoundTemplateFiles)
+
+			if handleTemplateError(w, r, "validator.go", "Validator", "GetValidatorDeposits", validatorNotFoundTemplate.ExecuteTemplate(w, "layout", data)) != nil {
+				return // an error has occurred and was processed
+			}
 			return
 		}
 		index, err = db.GetValidatorIndex(pubKey)
```
MauserBitfly, line 1813: If we get an error here, it still shows the error page, you can try with `0x1234`

handlers/validator.go
```diff
@@ -121,7 +121,12 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 		pubKey, err := hex.DecodeString(strings.Replace(vars["index"], "0x", "", -1))
 		if err != nil {
 			// logger.Errorf("error parsing validator public key %v: %v", vars["index"], err)
-			http.Error(w, "Internal server error", http.StatusInternalServerError)
+			SetPageDataTitle(data, "Validator not found")
```
MauserBitfly, line 124: Please extract the "Validator not found" code in a separate function, to reduce code duplication. 

Including the "validatorNotFoundTemplate" and "validatorNotFoundTemplateFiles" variables.

handlers/validator.go
```diff
@@ -121,7 +121,12 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 		pubKey, err := hex.DecodeString(strings.Replace(vars["index"], "0x", "", -1))
 		if err != nil {
 			// logger.Errorf("error parsing validator public key %v: %v", vars["index"], err)
```
MauserBitfly, line 123: I guess this comment can be removed.

handlers/validator.go
```diff
@@ -130,8 +135,13 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 			var name string
 			err := db.ReaderDb.Get(&name, `SELECT name FROM validator_names WHERE publickey = $1`, pubKey)
 			if err != nil && err != sql.ErrNoRows {
-				logger.Errorf("error getting validator-name from db for pubKey %v: %v", pubKey, err)
-				http.Error(w, "Internal server error", http.StatusInternalServerError)
+				// logger.Errorf("error getting validator-name from db for pubKey %v: %v", pubKey, err)
+				SetPageDataTitle(data, "Validatoooooorr not found")
```
MauserBitfly, line 139: Validatoooooorr   ;-)

handlers/validator.go
```diff
@@ -130,8 +135,13 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 			var name string
 			err := db.ReaderDb.Get(&name, `SELECT name FROM validator_names WHERE publickey = $1`, pubKey)
 			if err != nil && err != sql.ErrNoRows {
-				logger.Errorf("error getting validator-name from db for pubKey %v: %v", pubKey, err)
-				http.Error(w, "Internal server error", http.StatusInternalServerError)
```
MauserBitfly, line 134: We should only come here if the Database has an error (like not reachable or so) and then its ok to show the error.

templates/eth1txnotfound.html
```diff
@@ -21,6 +21,7 @@ <h1 class="h4 mb-1 mb-md-0"><i class="fas fa-history mr-2"></i>Tx not found</h1>
       <div class="card">
         <div class="card-body">
           <div class="d-1">Sorry but we could not find the tx you are looking for</div>
+          <div class="d-1">Stay calm, depending on the gas price of the transaction, <a href="https://kb.beaconcha.in/ethereum-2.0-and-depositing-process">it may still be in the mempool.</a></div>
```
MauserBitfly, line 24: This error belongs to validatornotfound

templates/validator/validatornotfound.html
```diff
@@ -22,7 +22,6 @@ <h1 class="h4 mb-1 mb-md-0">Validator not found</h1>
       <div class="card">
         <div class="card-body">
           <div class="d-1">Sorry, but we could not find the validator you are looking for.</div>
-          <div class="d-1">Stay calm, depending on the gas price of the transaction, <a href="https://kb.beaconcha.in/ethereum-2.0-and-depositing-process">it may still be in the mempool.</a></div>
```
MauserBitfly, line 25: This error belongs here ... check the link - it explains the process of adding a validator

handlers/validator.go
KostasGiannakakis: Wanted to emphasize that the validator was not found :D

templates/validator/validatornotfound.html
KostasGiannakakis: yes ok, it makes sense now. I will add this message back, but I believe it could be phrased better to reflect that the initial validator's transaction has not yet reached the deposit contract. Will look at the message tho later.

handlers/validator.go
```diff
@@ -130,6 +127,7 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 			var name string
 			err := db.ReaderDb.Get(&name, `SELECT name FROM validator_names WHERE publickey = $1`, pubKey)
 			if err != nil && err != sql.ErrNoRows {
+				// logger.Errorf("error getting validator-name from db for pubKey %v: %v", pubKey, err)
```
MauserBitfly, line 130: Pls. remove that comment (is the same as the next line)

handlers/validator.go
```diff
@@ -297,10 +290,7 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
 
 	if err == sql.ErrNoRows {
-		data := InitPageData(w, r, "validators", fmt.Sprintf("/validator/%v", index), "", validatorNotFoundTemplateFiles)
-		if handleTemplateError(w, r, "validator.go", "Validator", "no rows", validatorNotFoundTemplate.ExecuteTemplate(w, "layout", data)) != nil {
-			return // an error has occurred and was processed
-		}
+		validatorNotFound(data, w, r, vars, "")
 		return
 	} else if err != nil {
 		logger.Errorf("error getting validator for %v route: %v", r.URL.String(), err)
```
MauserBitfly, line 296: Hmm maybe we should call validatorNotFound on every error here ... 
like when you navigate to http://localhost:8081/validator/152465465465 you would get the error page, but we want the validatorNotFound

MauserBitfly (approved):
LGTM

MauserBitfly (approved):
LGTM

PR #2176, opened 2023-05-05
(BIDS-1824) show validator index if available

MauserBitfly:
### <samp>🤖 Generated by Copilot at 353c0e8</samp>

Fixed a bug where validator index was not shown for some statuses. Simplified the condition for displaying the index in `templates/validator/heading.html`.

### DIFF
--- a/templates/validator/heading.html
+++ b/templates/validator/heading.html
@@ -4,7 +4,7 @@
       <div class="d-flex mb-1">
         <h1 class="h4 mb-1 mb-md-0 d-flex align-items-center">
           <!-- using nonconventional characters for whitespaces because of linter -->
-          <span class="mr-2">Validator <span class="d-inline">{{ if ne .Index 0 }}{{ .Index }}{{ end }}{{ if ne .Name "" }}&nbsp;({{ formatValidatorName .Name }}){{ end }}</span></span>
+          <span class="mr-2">Validator <span class="d-inline">{{ .Index }}{{ if ne .Name "" }}&nbsp;({{ formatValidatorName .Name }}){{ end }}</span></span>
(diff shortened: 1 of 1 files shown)


D13ce (changes requested):
I am afraid that the Ethereum Foundation will not like the implementation:
Compare https://prater.beaconcha.in/validator/0 with http://localhost:8081/validator/0 (using your branch): The page does not show the correct validator index 0 anymore.

MauserBitfly:
> I am afraid that the Ethereum Foundation will not like the implementation: Compare https://prater.beaconcha.in/validator/0 with http://localhost:8081/validator/0 (using your branch): The page does not show the correct validator index 0 anymore.

Remove the check for the index completely as we navigate to the page only if we find something

D13ce (changes requested):
This approach is wrong. I have tested this with Prater locally and here is how to trigger a bug:

1. Go to https://prater.beaconcha.in/validators/deposits and look for the youngest deposit by sorting by time:
image
2. Click the validator`s public key
3. In my case that lead to https://prater.beaconcha.in/validator/0xb2ad32537d12c3cbbbcbc03f0209ede7d8388f94e99990deb5c43f5aa5c999b34152384a2bfd60361d3d2a623e012d8b#deposits which does not show an index (which is correct).
4. http://localhost:8081/validator/0xb2ad32537d12c3cbbbcbc03f0209ede7d8388f94e99990deb5c43f5aa5c999b34152384a2bfd60361d3d2a623e012d8b#deposits on the other hand shows an index of 0 which is not correct:
image

MauserBitfly:
> This approach is wrong. I have tested this with Prater locally and here is how to trigger a bug:

hope I got it right this time ;-)

PR #2179, opened 2023-05-05
(BIDS-751) missed blocks are not tracked for validators

MauserBitfly:
### <samp>🤖 Generated by Copilot at df86e51</samp>

Fixed a bug in `db/statistics.go` that caused incorrect validator statistics. The bug affected the SQL query that calculated the number of blocks proposed by validators in a given epoch range.


Eisei24 (approved):
lgtm

I also could not find any wrong `missed_attestations`.

PR #2181, opened 2023-05-05
(BIDS-2056) el-rewards are missing in chart after proposal

MauserBitfly:
### <samp>🤖 Generated by Copilot at d2a59db</samp>

Improve dashboard chart rendering and usability. Use a single color for `handlers/dashboard.go` and handle non-finalized blocks and x-axis better.


Eisei24 (commented):
lgtm.

I could only reproduce the bug the first time I checked a validator on prater, every other time it also already worked there.

There is just one issue that I think is unrelated.
The rewards for block proposals are most of the time one day too early in the charts.
See, e.g. https://prater.beaconcha.in/validator/183107#blocks
It proposed a block in epoch 173905 (2023-05-05) but in the chart the rewards show up for 5/4/2023 (Epochs 173568 - 173792).

Eisei24 (approved):
As discussed a new BIDS will be created for the day offset.

PR #2182, opened 2023-05-05
(BIDS-2000) fix wrong tooltip for estimated activation time of validator

KostasGiannakakis:
### <samp>🤖 Generated by Copilot at 68c0253</samp>

Fixed a bug in the validator overview page that caused the screen reader to read the wrong estimated activation date for validators in the activation queue. Changed the `aria-ethereum-date` attribute to use the correct template value in `templates/validator/overview.html`.


D13ce (approved):
utACK (no pending validator on Prater and Sepolia :cry: )

PR #2186, opened 2023-05-05
(BIDS-1944) return the Signature in Deposits page as hex

KostasGiannakakis:
### <samp>🤖 Generated by Copilot at a6fdef4</samp>

Encode deposit signatures as hex in the API response. This improves the consistency and compatibility of the explorer's API with the Eth2.0 API standard.

### DIFF
--- a/handlers/slot.go
+++ b/handlers/slot.go
@@ -511,7 +511,7 @@ func SlotDepositData(w http.ResponseWriter, r *http.Request) {
 			utils.FormatPublicKey(deposit.PublicKey),
 			utils.FormatBalance(deposit.Amount, currency),
 			utils.FormatWithdawalCredentials(deposit.WithdrawalCredentials, true),
-			deposit.Signature,
+			fmt.Sprintf("0x%v", hex.EncodeToString(deposit.Signature)),
 		})
 	}
 
--- a/handlers/slot.go
+++ b/handlers/slot.go
@@ -512,6 +512,7 @@ func SlotDepositData(w http.ResponseWriter, r *http.Request) {
 			utils.FormatBalance(deposit.Amount, currency),
 			utils.FormatWithdawalCredentials(deposit.WithdrawalCredentials, true),
 			fmt.Sprintf("0x%v", hex.EncodeToString(deposit.Signature)),
+			utils.FormatHash(deposit.Signature, true),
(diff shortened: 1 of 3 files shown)


D13ce (changes requested):
This does not fix the issue, does it?
Take a look at the signatures listed under https://prater.beaconcha.in/slot/5592330#deposits
Then start your branch locally and navigate to http://localhost:8081/slot/5592330#deposits
The signatures do not change:
image

Did I miss something?

handlers/slot.go
```diff
@@ -511,7 +511,7 @@ func SlotDepositData(w http.ResponseWriter, r *http.Request) {
 			utils.FormatPublicKey(deposit.PublicKey),
 			utils.FormatBalance(deposit.Amount, currency),
 			utils.FormatWithdawalCredentials(deposit.WithdrawalCredentials, true),
-			deposit.Signature,
+			fmt.Sprintf("0x%v", hex.EncodeToString(deposit.Signature)),
```
D13ce, line 514: Please use `utils.FormatHash` for this as done, for example, here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/20a7a4ee12dec21e06586566257bbd86618a01aa/handlers/eth2Deposits.go#L87

**Reasoning:**
Navigate to, for example, https://prater.beaconcha.in/validator/0xa370ed2029a9a36c6fa4449a254c9c0e68e01824e5a60be5b9bd1ebe78c79e99b2177112d6008b4f2c41864c5445d2b1#deposits and take a look at the deposit's signature there:
image
It shows the first and last 4 characters which is reasonable.

If you run you branch on Prater locally and navigate to http://localhost:8081/slot/5598476#deposits though, the same signature is displayed like this though:
image

---

Furthermore, the font type currently used is wrong (as it isn't monospaced). This should be fixed by using `utils.FormatHash` too.

---

**Note:**
FYI, I have not tested the change I am suggesting, maybe more changes (for example in the according HTML) are required too.

D13ce (changes requested):

handlers/slot.go
KostasGiannakakis: Yeap, there were a couple of more things needed in the html side to be taken of.

PR #2195, opened 2023-05-09
(BIDS-1695) Change button design of hisorical fiat value

Eisei24:
copilot:summary


D13ce (approved):
lgtm

Please note that I could not test this with the explorer set to fiat as this missing on Prater.

PR #2201, opened 2023-05-10
(BIDS-2021) use pool data for charts

MauserBitfly:
### <samp>🤖 Generated by Copilot at 1eb4c80</samp>

Improved pools distribution chart data by using cached pool data and supporting negative balance changes. Modified `services/charts_updater.go`.

### DIFF
--- a/services/charts_updater.go
+++ b/services/charts_updater.go
@@ -969,7 +969,7 @@ func poolsDistributionChartData() (*types.GenericChartData, error) {
 		Drilldown string `json:"drilldown"`
 	}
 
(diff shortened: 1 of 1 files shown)


Eisei24 (commented):
Everything works but the colours of the chart change because I guess the order in `rows` is different. Is that fine?

MauserBitfly:
> Everything works but the colours of the chart change because ETH.STORE is now part of `seriesData` with a `Y = -1` as the first entry. It should probably be removed.

removed it

Eisei24 (approved):
lgtm

PR #2205, opened 2023-05-11
(BIDS-2030) add full and yearly validator performance

MauserBitfly:
copilot:summary

### DIFF
--- a/handlers/api_eth1.go
+++ b/handlers/api_eth1.go
@@ -911,6 +911,7 @@ func getValidatorExecutionPerformance(queryIndices []uint64) ([]types.ExecutionP
 	if latestEpoch < 7200 {
 		monthRange = 0
 	}
+	validatorsPQArray := pq.Array(queryIndices)
 
 	var execBlocks []types.ExecBlockProposer
 	err := db.ReaderDb.Select(&execBlocks,
@@ -922,7 +923,7 @@ func getValidatorExecutionPerformance(queryIndices []uint64) ([]types.ExecutionP
 		AND exec_block_number IS NOT NULL 
 		AND exec_block_number > 0 
 		AND epoch > $2`,
-		pq.Array(queryIndices),
+		validatorsPQArray,
 		monthRange, // 32d range
 	)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


handlers/api_eth1.go
```diff
@@ -973,6 +1009,10 @@ func getValidatorExecutionPerformance(queryIndices []uint64) ([]types.ExecutionP
 			producerReward = mevBribe
 		}
 
+		if block.Time.AsTime().After(firstEpochTime) {
+			result.PerformanceTotal = result.PerformanceTotal.Add(result.PerformanceTotal, producerReward)
+			result.Performance365d = result.Performance365d.Add(result.Performance365d, producerReward)
+		}
```
Eisei24, line 1015: Do we want to add the data from the current day to the results?
The BIDS makes it sound like they should be ignored.

handlers/api_eth1.go
```diff
@@ -973,6 +1009,10 @@ func getValidatorExecutionPerformance(queryIndices []uint64) ([]types.ExecutionP
...
 		if block.Time.AsTime().After(last30dTimestamp) {
```
Eisei24, line 1016: Minor: It slightly annoys me that this is called **last30dTimestamp** when it is for 31 days.

Eisei24 (changes requested):
This looks well structured but please have a look at my comments.

handlers/api_eth1.go
```diff
@@ -940,19 +941,54 @@ func getValidatorExecutionPerformance(queryIndices []uint64) ([]types.ExecutionP
...
+	}
+
+	performanceList := []LongPerformanceResponse{}
+
+	err = db.ReaderDb.Select(&performanceList, `
+		SELECT 
+		validatorindex,
+		CAST(COALESCE(mev_performance_365d, 0) AS bigint) AS el_performance_365d,
+		CAST(COALESCE(mev_performance_total, 0) AS bigint) AS el_performance_total
+		FROM validator_performance WHERE validatorindex = ANY($1)`, validatorsPQArray)
```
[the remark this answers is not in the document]
MauserBitfly: We will leave the monthly logic as it is.

handlers/api_eth1.go
MauserBitfly: You are right, we also need to consider Equal

handlers/api_eth1.go
MauserBitfly: So it shell be renamed!

Eisei24 (approved):
lgtm

PR #2211, opened 2023-05-15
(NOBIDS) check if all epochs are finalized for stats

MauserBitfly:
### <samp>🤖 Generated by Copilot at 6202936</samp>

This pull request improves the statistics export logic for the eth2-beaconchain-explorer by using finality status instead of indexing status, adding `break` statements to optimize the switch cases, and adding a new function to count finalized epochs.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -479,6 +479,18 @@ func GetAllEpochs() ([]uint64, error) {
 	return epochs, nil
 }
 
+// Count finalized epochs in range
+func CountFinalizedEpochs(startEpoch uint64, endEpoch uint64) (uint64, error) {
+	var count uint64
+	err := WriterDb.Get(&count, "SELECT COUNT(*) FROM epochs WHERE epoch >= $1 AND epoch <= $2 AND finalized", startEpoch, endEpoch)
+
+	if err != nil {
+		return 0, fmt.Errorf("error counting finalized epochs [%v -> %v] from DB: %w", startEpoch, endEpoch, err)
+	}
+
+	return count, nil
+}
+
 // GetLastPendingAndProposedBlocks will return all proposed and pending blocks (ignores missed slots) from the database
 func GetLastPendingAndProposedBlocks(startEpoch, endEpoch uint64) ([]*types.MinimalBlock, error) {
 	var blocks []*types.MinimalBlock
--- a/db/db.go
+++ b/db/db.go
@@ -479,7 +479,7 @@ func GetAllEpochs() ([]uint64, error) {
 	return epochs, nil
 }
 
-// Count finalized epochs in range
+// Count finalized epochs in range (including start and end epoch)
 func CountFinalizedEpochs(startEpoch uint64, endEpoch uint64) (uint64, error) {
 	var count uint64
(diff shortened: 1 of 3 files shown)


db/statistics.go
```diff
@@ -28,13 +28,13 @@ func WriteValidatorStatisticsForDay(day uint64) error {
...
-	latestDbEpoch, err := GetLatestEpoch()
+	finalizedCount, err := CountFinalizedEpochs(firstEpoch, lastEpoch)
 	if err != nil {
 		return err
 	}
 
-	if lastEpoch > latestDbEpoch {
-		return fmt.Errorf("delaying statistics export as epoch %v has not yet been indexed. LatestDB: %v", lastEpoch, latestDbEpoch)
+	if finalizedCount < epochsPerDay {
+		return fmt.Errorf("delaying statistics export as not all epochs for this day are finalized. LatestDB: %v", day)
```
D13ce, line 37: **Minor:**
The use of `day` here is wrong, is it not? Maybe the last part can be removed?
```suggestion
		return fmt.Errorf("delaying statistics export as not all epochs for this day are finalized.")
```

Furthermore, you could add something like "(%d of %d)" with the parameters `finalizedCount` and `epochsPerDay`.

db/statistics.go
```diff
@@ -699,13 +699,13 @@ func WriteChartSeriesForDay(day int64) error {
...
-	latestDbEpoch, err := GetLatestEpoch()
+	finalizedCount, err := CountFinalizedEpochs(firstEpoch, uint64(lastEpoch))
 	if err != nil {
 		return err
 	}
 
-	if (uint64(lastSlot) / utils.Config.Chain.Config.SlotsPerEpoch) > latestDbEpoch {
-		return fmt.Errorf("delaying statistics export as epoch %v has not yet been indexed. LatestDB: %v", (uint64(lastSlot) / utils.Config.Chain.Config.SlotsPerEpoch), latestDbEpoch)
+	if finalizedCount < epochsPerDay {
+		return fmt.Errorf("delaying chart series export as not all epochs for this day are finalized. LatestDB: %v", day)
```
D13ce, line 708: **Minor:**
The use of `day` here is wrong, is it not? Maybe the last part can be removed?
```suggestion
		return fmt.Errorf("delaying chart series export as not all epochs for this day are finalized.")
```

Furthermore, you could add something like "(%d of %d)" with the parameters `finalizedCount` and `epochsPerDay`.

db/db.go
```diff
@@ -479,6 +479,18 @@ func GetAllEpochs() ([]uint64, error) {
 	return epochs, nil
 }
 
+// Count finalized epochs in range
```
D13ce, line 482: **Nitpick:** Could state that both `start` and `end` are inclusive (at least `end` is not always inclusive in "similar" functions).

D13ce (changes requested):
Two minor things found, I suggest to change those since the log now does not make perfect sense.

db/statistics.go
MauserBitfly: fixed error

db/statistics.go
MauserBitfly: fixed error

db/db.go
MauserBitfly: added to comment

D13ce (approved):
lgtm

PR #2213, opened 2023-05-15
Bids 1846/gnosis

guybrush:
### <samp>🤖 Generated by Copilot at 521df37</samp>

This pull request adds a new feature to the `misc` command, updates the price and currency formatting functions to use the price service and the config values, and supports the gnosis fork of the mainnet. It also adds a `versionFlag` to the `frontend-data-updater` program, updates the dockerfile to use a newer golang version, and fixes some minor typos and bugs.


db/bigtable_eth1.go
```diff
@@ -2817,20 +2817,11 @@ func (bigtable *Bigtable) GetMetadataForAddress(address []byte) (*types.Eth1Addr
...
-		if string(ret.Balances[j].Metadata.Price) != "" {
-			priceJ, err = decimal.NewFromString(string(ret.Balances[j].Metadata.Price))
-			if err != nil {
-				logger.WithError(err).Errorf("error parsing string price value, price: %s", ret.Balances[j].Metadata.Price)
-			}
+		if len(ret.Balances[j].Metadata.Price) > 0 {
+			priceJ = decimal.NewFromBigInt(new(big.Int).SetBytes(ret.Balances[j].Metadata.Price), 0)
 		}
 
 		mulI := decimal.NewFromFloat(float64(10)).Pow(decimal.NewFromBigInt(new(big.Int).SetBytes(ret.Balances[i].Metadata.Decimals), 0))
```
Eisei24, line 2827: For this 4 line block:
Do you need a precision great than 16?
If so you would need `DivRound`.

handlers/eth1Token.go
```diff
@@ -72,51 +72,35 @@ func Eth1Token(w http.ResponseWriter, r *http.Request) {
...
-		ethUsdRate := decimal.NewFromFloat(price.GetEthPrice("USD"))
-		logger.Infof("usd rate %s", ethUsdRate)
+		ethUsdRate := decimal.NewFromFloat(price.GetPrice(utils.Config.Frontend.ElCurrencySymbol, "USD"))
 		if !ethUsdRate.IsZero() {
 			ethExchangeRate, _ = tokenPrice.Div(ethUsdRate).Float64()
 		}
 	}
 
 	data := InitPageData(w, r, "blockchain", "/token", fmt.Sprintf("Token 0x%x", token), templateFiles)
+	supply := decimal.NewFromBigInt(new(big.Int).SetBytes(metadata.TotalSupply), 0).Div(decimal.NewFromInt(10).Pow(decimal.NewFromBigInt(new(big.Int).SetBytes(metadata.Decimals), 0)))
```
Eisei24, line 89: I assume you need a `DivRound(..., metadata.Decimals)` here. (Pseudocode from me)

rpc/erigon.go
```diff
@@ -636,12 +655,30 @@ func (client *ErigonClient) GetERC20TokenMetadata(token []byte) (*types.ERC20Met
...
+	})
+
 	err = g.Wait()
+	if err != nil {
+		return ret, err
+	}
+
+	num := new(big.Int).Exp(big.NewInt(10), new(big.Int).SetBytes(ret.Decimals), nil)
+	den := new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil)
+	price := new(big.Int).Quo(new(big.Int).Mul(ethRate, num), den)
```
Eisei24, line 680: Is it intended here that you lose all after comma digits like for a normal int division?

rpc/geth.go
```diff
@@ -404,12 +424,30 @@ func (client *GethClient) GetERC20TokenMetadata(token []byte) (*types.ERC20Metad
...
+	})
+
 	err = g.Wait()
+	if err != nil {
+		return ret, err
+	}
+
+	num := new(big.Int).Exp(big.NewInt(10), new(big.Int).SetBytes(ret.Decimals), nil)
+	den := new(big.Int).Exp(big.NewInt(10), big.NewInt(18), nil)
+	price := new(big.Int).Quo(new(big.Int).Mul(ethRate, num), den)
```
Eisei24, line 449: Same question as for Erigon.
Is it intended here that you lose all after comma digits like for a normal int division?

D13ce (changes requested):
First of: Don't be scared by the amount of comments. There are lot of questions and minor things.

Also I have to preface this review with the following information:
I have started your branch locally against Prater. I have NOT adapted my `prater.config.yml` file, I relied on the default values you have added in `ReadConfig`. I did as we will have to merge this PR at some point anyways.
This also means: The currency cookie was unavailable in my test environment as I was testing against Prater and therefore I had no way to select a currency.

Unfortunately, this branch crashes on various pages as many calls of `FormatClCurrency`/`FormatElCurrency` have too few parameters in the according HTML files (I have addressed this in a comment). I have fixed that by adding a few `true`.

Besides the comments that are listed below, here are some more things I'd like to ask:
**1)** Searching for `"ether"` shows a few occurrences that might have to be changed too:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/0c60c08f254e4ea8acd7c7d8e4fa875bc4c85b97/templates/mempoolTx.html#L80 (Gnosis production shows "xDAI" here, I don't know how that happens honestly)
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/e3d144f00fe7e3c6231cba739fb93b0c50e212f8/types/api.go#L216 (not sure how to tackle that)

**2)** The same applies to `"eth"`.
Here are some random examples:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/1312b2c7e94c2ecfe78348b55aaaee7d56504a83/services/charts_updater.go#L670
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/1682b3ad71e34ed064a2555a272cbf4babbac8ee/templates/validatorRewards.html#L70
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/5d30c55d893bd4e2b56d8e21374c7043830a9c9f/templates/calculator.html#L1021

**3)** This PR changes the way currency is shown at some locations. For example:
https://prater.beaconcha.in/validators/withdrawals
https://prater.beaconcha.in/address/0x8f0844fd51e31ff6bf5babe21dccf7328e19fd9f
“ETH” is shown with this branch instead of “Ether”.
Is that intentional? Personally I think that that’s good :)

**4)** The "ERC20 Token Txns" table shows way more digits for the value. Here is prater production:
image
Here is my local instance (as mentioned above, your branch used for Prater):
image
(Please note that this might be caused by my setup.)

**5)** There are several GitHub Actions error within this PR.

**6)** The mobile app most likely should not be advertised on explorers for Gnosis as the app does not support Gnosis.
It is advertised on the landing page
image
and the menu
image

**7)** The `/calculator` needs many adjustments as it is still using ETH. This can be seen on your Gnosis instance too.

PR #2219, opened 2023-05-16
(BIDS-1867) fix for correctly updating next withdrawal line

KostasGiannakakis:
### <samp>🤖 Generated by Copilot at 5998cb1</samp>

This pull request refactors and optimizes the code for the withdrawal table in the dashboard. It removes unused and redundant code, fixes a linting error, and reduces database queries. The affected files are `handlers/dashboard.go`, `templates/dashboard/tables.html`, and `types/templates.go`.

### DIFF
--- a/handlers/dashboard.go
+++ b/handlers/dashboard.go
@@ -553,13 +553,17 @@ func DashboardDataWithdrawals(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var tableData [][]interface{}
+
+	// check if there is a NextWithdrawal and append
 	NextWithdrawalRow, err := getNextWithdrawalRow(validators)
 	if err != nil {
 		logger.WithError(err).WithField("route", r.URL.String()).Error("error calculating next withdrawal row")
+		tableData = make([][]interface{}, 0, len(withdrawals))
 	} else {
 		if NextWithdrawalRow == nil {
 			tableData = make([][]interface{}, 0, len(withdrawals))
 		} else {
+			// make the array +1 larger to append the NextWithdrawal row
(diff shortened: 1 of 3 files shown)


handlers/dashboard.go
```diff
@@ -156,7 +156,7 @@ func Dashboard(w http.ResponseWriter, r *http.Request) {
 	validatorLimit := getUserPremium(r).MaxValidators
 
 	q := r.URL.Query()
-	queryValidators, err := parseValidatorsFromQueryString(q.Get("validators"), validatorLimit)
+	_, err := parseValidatorsFromQueryString(q.Get("validators"), validatorLimit)
```
D13ce, line 159: **Minor:** It seems counterintuitive to call `parseValidatorsFromQueryString` when the return value does not matter.
I guess you do that to get its limit handling i.e. if too many validators are passed you want the `err` to be returned.

While I guess this is fine (even though it includes and unnecessary overhead), you should at least add a comment:
```suggestion
	// Using parseValidatorsFromQueryString to check for validatorLimit
	_, err := parseValidatorsFromQueryString(q.Get("validators"), validatorLimit)
```

KostasGiannakakis: Yes, that sounds reasonable enough to me. I also do not like it that way, but the function introduces some error handling. In the future when a bigger refactor is considered we can optimize that part. For now the `parseValidatorsFromQueryString` does not add a big overhead. The function is O(n), and with a max of 300 validators check, is not something to consider for performance issues.

D13ce (approved):
lgtm

PR #2220, opened 2023-05-16
(BIDS-1694) truncate only tooltip for method names

MauserBitfly:
### <samp>🤖 Generated by Copilot at 18e315d</samp>

Fixed a bug that prevented tooltips from showing up for truncated method names in the UI. Added a custom `truncate-tooltip` attribute and a function in `layout.js` to handle the tooltip logic, and replaced the `data-toggle` attribute in various HTML and Go files.

### DIFF
--- a/handlers/slot.go
+++ b/handlers/slot.go
@@ -718,7 +718,7 @@ func BlockTransactionsData(w http.ResponseWriter, r *http.Request) {
 	for i, v := range transactions.Txs {
 		methodFormatted := `<span class="badge badge-light">Transfer</span>`
 		if len(v.Method) > 0 && v.Method != "Transfer" {
-			methodFormatted = fmt.Sprintf(`<span class="badge badge-light text-truncate mw-100" truncate-tooltip="%v"{>%v</span>`, v.Method, v.Method)
+			methodFormatted = fmt.Sprintf(`<span class="badge badge-light text-truncate mw-100" truncate-tooltip="%v">%v</span>`, v.Method, v.Method)
 		}
 		data[i] = &transactionsData{
 			HashFormatted: v.HashFormatted,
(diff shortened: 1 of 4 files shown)


handlers/slot.go
```diff
@@ -718,7 +718,7 @@ func BlockTransactionsData(w http.ResponseWriter, r *http.Request) {
 	for i, v := range transactions.Txs {
 		methodFormatted := `<span class="badge badge-light">Transfer</span>`
 		if len(v.Method) > 0 && v.Method != "Transfer" {
-			methodFormatted = fmt.Sprintf(`<span class="badge badge-light text-truncate mw-100" data-toggle="tooltip" title="%v"{>%v</span>`, v.Method, v.Method)
+			methodFormatted = fmt.Sprintf(`<span class="badge badge-light text-truncate mw-100" truncate-tooltip="%v"{>%v</span>`, v.Method, v.Method)
```
Eisei24, line 721: The `{` looks like a typo.

static/js/layout.js
```diff
@@ -503,6 +504,23 @@ function formatAriaEthereumDate(elem) {
...
+  let nodes = $("[truncate-tooltip]")
+  nodes.each((_, node) => {
+    console.log(node)
+    let title = ""
+    if (node.scrollWidth > node.offsetWidth) {
+      title = node.attributes["truncate-tooltip"].value
+    }
+    if (node.attributes["title"]?.value != title) {
+      node.setAttribute("title", title)
+      if (title > 0) {
```
Eisei24, line 517: Would this be preferred?

```suggestion
      if (title !== "") {
```

Eisei24 (changes requested):
As discussed the tooltip does not work when the 3 dots at the end when we truncate only replace 2 characters.

We could also consider using our other tooltips with data-toggle-tooltip, see e.g. /address/0xb8ff877ed78ba520ece21b1de7843a8a57ca47cb (prater) where every tooltip other than the method name uses it.

handlers/slot.go
MauserBitfly: removed

static/js/layout.js
MauserBitfly: thats better yes

Eisei24 (approved):
lgtm

PR #2222, opened 2023-05-16
(BIDS-2094) index page: dont show fractional validators

LuccaBitfly:
### <samp>🤖 Generated by Copilot at 97c2210</samp>

Improved the deposit chart UI in the index page and fixed a race condition when loading templates from the embedded filesystem. Moved the template variables and used `templates.GetTemplate` in `handlers/index.go`.

### DIFF
--- a/handlers/index.go
+++ b/handlers/index.go
@@ -7,36 +7,32 @@ import (
 	"eth2-exporter/types"
 	"eth2-exporter/utils"
 	"fmt"
-	"html/template"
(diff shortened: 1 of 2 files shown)


D13ce (approved):
lgtm, also thanks for the minor refactoring.

By the way, your change also removes the decimal places for the ETH line which was not requested. I have discussed this with @Buttaa though and it's fine.

PR #2225, opened 2023-05-17
(BIDS-2081) include genesis day in income history report

MauserBitfly:
### <samp>🤖 Generated by Copilot at a80158b</samp>

Fix income history query bug for validators near genesis time. Adjust query lower bound in `services/rewards.go` to align with genesis timestamp.


Eisei24 (approved):
lgtm

It works for UTC time and near it.
You do get some weird output if you set your local time to something like -11 UTC or +11 UTC but that could be intentional and I don't think there is a point in changing it.

Issue #2227, opened 2023-05-17
API Authorization Issue: Inconsistent Authorization Mechanism Hinders Validator Node Automation


SKlarsen84:
**Describe the bug**
We are currently attempting to automate the process of adding validator nodes to our notifications center through the API. 

The API call for this purpose (https://beaconcha.in/api/v1/user/validator/%7Bpubkey%7D/add) returns an error stating: "invalid_request" and "missing authorization header". 

Upon checking the codebase, it's clear that the API requires an "Authorization" header, which contradicts the swagger docs which say that one should provide a header "apikey" containing  the <API KEY>.

`The API key can be provided in the Header or as a query string parameter.
Key as a query string parameter: curl https://beaconcha.in/api/v1/slot/1?apikey=<your_key>
Key in a request header: curl -H 'apikey: <your_key>' https://beaconcha.in/api/v1/slot/1`

Once I provide a proper Authorization header, I no longer get asked for "missing authorization header" but then encounter a "unauthorized_client" and "authorization failed" error when running the API in postman.

**To Reproduce**

Attempt to make an API call to e.g 'https://beaconcha.in/api/v1/user/validator/123/add'
See the "invalid_request" and "missing authorization header" error
Add an "Authorization" header with a valid user API token to the request
Run the API call again
Encounter the "unauthorized_client" and "authorization failed" error

**Expected behavior**
We expected that adding the "Authorization" header with a valid user API token to the request would resolve the "missing authorization header" error and allow us to make the API call successfully. This would enable us to automate the process of adding validator nodes to our notifications center, as required for our use case.

SKlarsen84:
Looking into the code, it sure does look like the code expects a proper JWT token in BEARER format for it to be able to validate:

`
func stripOffBearerFromToken(tokenString string) (string, error) {
	if len(tokenString) > 6 && strings.ToUpper(tokenString[0:6]) == "BEARER" {
		return tokenString[7:], nil
	}
	return tokenString, nil //"", errors.New("Only bearer tokens are supported, got: " + tokenString)
}

func accessTokenGetClaims(tokenStringFull string, validate bool) (*CustomClaims, error) {
	tokenString, err := stripOffBearerFromToken(tokenStringFull)
	if err != nil {
		return nil, err
	}

	token, err := jwt.ParseWithClaims(tokenString, &CustomClaims{}, func(token *jwt.Token) (interface{}, error) {
		return getSignKey()
	})

[...]

PR #2228, opened 2023-05-17
add support for apikey header to authorize requests to User api endpoints

SKlarsen84:
Suggested fix for  #2227 

 The `oauthutils.go` file has been updated to add support for the `apikey` header to authorize requests. If the `Authorization` header is not present, the middleware will attempt to get the `apikey` header. If both headers are missing, an error response will be sent. If the `apikey` is invalid, an error response will be sent. If there is an error creating the JWT token, an error response will be sent.

This PR tries to address the fact that API-key cannot currently be used to invoke API endpoints for the User, as only JWT tokens are properly authenticated.

🐛 fix(oauthutils.go): handle missing authorization header or apikey header 
🐛 fix(oauthutils.go): handle invalid api key
🐛 fix(oauthutils.go): handle error creating JWT token

### DIFF
--- a/utils/oauthutils.go
+++ b/utils/oauthutils.go
@@ -4,6 +4,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"eth2-exporter/db"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


Buttaa:
See comment in issue :(

PR #2230, opened 2023-05-17
(BIDS-2019) move Time in a template

KostasGiannakakis:
### <samp>🤖 Generated by Copilot at 05f719b</samp>

Added a reusable `timeToggle` component to show and switch the time format of an epoch. Used the component in the `epoch.html` template to replace the original code.


templates/components/timeToggle.html
```diff
@@ -0,0 +1,14 @@
+{{ define "timeToggle" }}
+  <div class="row border-bottom p-3 mx-0">
+    <div class="col-md-2">Time:</div>
```
D13ce, line 3: This breaks the formatting:
image
The time now is placed to much to the left.

templates/components/timeToggle.html
D13ce, line 1: While I really like the idea to put this into a separate file ... I am REALLY confused by the fact that you did not make any use of that refactoring. There are many places where you could use it. Simply search, for example, for `setTs()` and you find about 4 more locations.

D13ce (changes requested):
Left some comments, ptal

templates/components/timeToggle.html
KostasGiannakakis: I am already aware of that. I wanted to first send this one, and make sure it works properly, before proceeding with the rest of the refactoring. I should have explained it with a comment under this PR.

templates/components/timeToggle.html
KostasGiannakakis: will take a look for this one. thnx

D13ce (changes requested):
Not all comments have been solved yet.

templates/components/timeToggle.html
KostasGiannakakis: @D13ce I updated all relevant files, to include the new template for the timeRow

D13ce (changes requested):
Not a single page out of the 4 ones you changed with your latest commits work. Please check.

templates/slot/slotFuture.html
D13ce, line 1: This change breaks the position of the time value, it is not aligned properly anymore (compare it with the slot line above):

image

D13ce (changes requested):
I have found two placement issues, see my comments below.

At this point I believe that at most only the 3 buttons should be part of `timeRow.html` and the file should be renamed accordingly. Thoughts?

KostasGiannakakis:
I decided to revert back to an easier approach. The template for the timeRow I created it seems to have some inconsistencies and it led to require more work than I was originally expecting. With the `autocomplete="off"` approach, the issue is being solved. I will look again in the future whether it is possible to create more reusable components for our frontend. 
cc. @D13ce

D13ce (approved):
lgtm

PR #2234, opened 2023-05-20
(NOBIDS) use redis for rate limiting metrics inserts

peterbitfly:
Switched to redis for rate limiting metric inserts. Uses the SETNX function with the metrics key + the current minute as rate limit key. Redis keys expire after 1 minute

### <samp>🤖 Generated by Copilot at 2969628</samp>

Added Redis caching to the `db` package and updated all commands that use the `InitBigtable` function to pass the Redis address as a parameter. This improves the performance and scalability of the machine metrics feature by reducing the load on Bigtable.


recy21 (approved):

manuelsc (approved):
lgtm

PR #2241, opened 2023-05-23
(BIDS-2108) Remove deposited_valid state from validators

Eisei24:
Please have a look at the comments of the BIDS since the actually implemented solution was pretty much the opposite of what the description says.

### DIFF
--- a/handlers/validator.go
+++ b/handlers/validator.go
@@ -176,17 +176,18 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 				validatorPageData.InclusionDelay = 0
 			}
 
-			for _, deposit := range validatorPageData.Deposits.Eth1Deposits {
+			for _, deposit := range deposits.Eth1Deposits {
 				if deposit.ValidSignature {
 					validatorPageData.Eth1DepositAddress = deposit.FromAddress
 					break
 				}
 			}
 
-			// check if a valid deposit exists
-			for _, d := range deposits.Eth1Deposits {
-				if !d.ValidSignature {
+			// check if an invalid deposit exists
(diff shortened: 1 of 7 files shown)


handlers/validator.go
```diff
@@ -183,21 +183,13 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 				}
 			}
 
-			sumValid := uint64(0)
 			// check if a valid deposit exists
```
D13ce, line 186: **Minor:** This comment is wrong, it should be
```suggestion
			// check if an invalid deposit exists
```

handlers/validator.go
```diff
@@ -183,21 +183,13 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
 			for _, d := range deposits.Eth1Deposits {
-				if d.ValidSignature {
-					sumValid += d.Amount
-				} else {
+				if !d.ValidSignature {
 					validatorPageData.Status = "deposited_invalid"
```
D13ce, line 189: **Nitpick:** You should add a `break` here:
```suggestion
					validatorPageData.Status = "deposited_invalid"
					break
```

D13ce (commented):
lgtm

Left two minor comments though, ptal.

handlers/validator.go
Eisei24: Sure, changed.
I also formatted a bit so that the two for loops look consistent.

handlers/validator.go
Eisei24: Changed.

D13ce (approved):
lgtm

PR #2245, opened 2023-05-24
(BIDS-2143) Fix ordering and page for two syncs

Eisei24:
To test this I would use /validator/198961#sync (prater) which was part of multiple sync committees.
Things that should be tested to work:

- First page
- Last page
- Page with 2 epochs (page 4)
- Page with 2 sync committees (page 820)

all of those in both descending and ascending order.
The bug mentioned in the comments was also fixed.

### DIFF
--- a/handlers/validator.go
+++ b/handlers/validator.go
@@ -1936,6 +1936,13 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
 				return a < b
 			}
 		}
+		diffValue := func(a uint64, b uint64) uint64 {
+			if a >= b {
+				return a - b
+			} else {
+				return b - a
+			}
+		}
 
 		// amount of epochs moved away from start epoch
 		epochOffset := (start / utils.Config.Chain.Config.SlotsPerEpoch)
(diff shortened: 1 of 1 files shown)


handlers/validator.go
```diff
@@ -1956,33 +1963,47 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
...
-		syncDuties, err := db.BigtableClient.GetValidatorSyncDutiesHistoryOrdered(validatorIndex, firstShownEpoch-limit, firstShownEpoch, ascOrdering)
+		var startEpoch, endEpoch uint64
+		if ascOrdering {
+			startEpoch = firstShownEpoch - 1
+			endEpoch = firstShownEpoch + limit - 1
+		} else {
+			startEpoch = firstShownEpoch - limit
+			endEpoch = firstShownEpoch
+		}
+		syncDuties, err := db.BigtableClient.GetValidatorSyncDutiesHistoryOrdered(validatorIndex, startEpoch, endEpoch, ascOrdering)
```
D13ce, line 1991: Shouldn't it be possible to merge this with
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/bf343ea490d488effc0de8a8033fe0e747c53a85/handlers/validator.go#L1999-L2012

What I mean by that: Try to find the right epoch range if two epochs are affected and query big table only once.

D13ce (commented):
Left a suggestion.

Further more, I think I came up with another bug:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/bf343ea490d488effc0de8a8033fe0e747c53a85/handlers/validator.go#L2016
I think this check is not good. What if a sync period recently started and only 12 duties are done in the current epoch? We would not limit the 12 duties to the only 10 we need because of this.
Please share your thoughts.

handlers/validator.go
Eisei24: The first query is for all epochs that are within the same sync committee and therefore next to each other, e.g. the page shows:
- 3x Epoch: 150000 
- 7x Epoch: 150001

The second query is for pages that show two different sync committees, e.g.

- 3x Epoch: 150001
- 7x Epoch: 50001

Between two different sync committees lie many epochs and it is much faster to query twice a single epoch than once 100000 epochs. While testing this I couldn't even load the query with these values, it  probably would have taken an hour.

Eisei24:
I am not sure what you mean with "We would not limit the 12 duties to the only 10 we need because of this." but the intention behind the sync tab view is that it only shows full epochs.
As an example, if a validator is currently participating in syncing epoch 178593 it may show 7 of the slots as **Participated** and 25 as **Scheduled**.
When you reload the page (or change pages back and forth) the **Scheduled** slowly become **Participated** but no new slots are added.
Then at once all 32 slots of epoch 178594 get added with most of them being **Scheduled**.

If the check fails the table would be completely empty.

D13ce (approved):
lgtm

PR #2246, opened 2023-05-24
(BIDS-1695) Change cursor

Eisei24:
A simple change of the cursor if a historic fiat value is available and the button can therefore be pressed.


D13ce (approved):
lgtm

PR #2247, opened 2023-05-24
(BIDS-2120) use no-data-to-display module (v8.2.2) to clear data and …

KostasGiannakakis:
…show correct message when there are no proposals by the validators

### <samp>🤖 Generated by Copilot at 87f8c0a</samp>

This pull request adds a feature to show a message when there is no data to display for the proposed chart in the dashboard page. It uses the `no-data-to-display.js` plugin from Highcharts and modifies the `renderCharts` function in `dashboard.js` and the `dashboard.html` template accordingly.


KostasGiannakakis:
**To the reviewer**
Regarding the failing linting, is because of an issue it encountered with the highcharts `no-data-to-display` module. I believe it is safe for that file only to add a rule that will instruct the linter to not take under consideration the error. I am waiting for your thoughts and I can update the linter (or the file) accordingly

static/js/dashboard.js
```diff
@@ -1094,8 +1094,14 @@ $(document).ready(function () {
         success: function (result) {
           var t1 = Date.now()
           var t2 = Date.now()
+          var chart = $("#proposed-chart").highcharts()
           if (result && result.length) {
             createProposedChart(result)
+          } else {
+            if (chart !== undefined) {
+              chart.showNoData("No proposals have been made from the selected validator(s)")
+              chart.series[0].setData([]) // Clear the series data
```
D13ce, line 1103: This is not enough. The chart has more than one series. Here is how to trigger a bug:

1. With you branch, navigate to the dashboard and add only the following two validators: `4835` and `483555`
FYI: Validator [4835](https://prater.beaconcha.in/validator/4835) is old and contains lots of proposals, [483555](https://prater.beaconcha.in/validator/483555)  is very young and (currently) does not include any proposals. If you test this later, validator `483555` may already include proposals, in that case you have to find another one.
2. View the proposals tab, it might look similar to this:
image
3. Remove validator `4835`. At this point, the chart should not show any data yet it looks like this:
image

static/js/highcharts/no-data-to-display.js
D13ce, line 1: I don't think that this is required. According to the bug description, we simply want to show the same default graphic that is shown when either no validator or only validators without proposals are added:
image

Since we are already able to show said graphic for both scenarios, why do we need to add further highcharts functionality to solve this bug?

One thing that should be considered:
If you trigger the new behavior and then refresh the page, it will default back to the graphic I have attached above. So refreshing the page changes the look of the page which should not happen.
For the sake of simplicity I would suggest to remove the new functionality.

Please share your thoughts.

D13ce (changes requested):
I have found a bug and suggest to take a different approach. ptal

static/js/highcharts/no-data-to-display.js
KostasGiannakakis: The original graphic is shown because is the default (hardcoded) html under the proposed-chart div. To return to the original one, we need to keep the html content in a variable, and reload it in case there are no data on the chart (currently when we want to load new data, we find the div mentioned before and we replace the content with a new Highcharts chart). IMO with the current implementation this is not very clean, to load and unload data from the div. The solution I proposed with the new functionality is more clear, and adheres to the highchart standards of how to handle missing data. In general I am in favor of removing the graphics and add more meaningful information when there is a lack of information.

static/js/dashboard.js
KostasGiannakakis: I replaced the series for all 3 potential states of a proposal.

static/js/highcharts/no-data-to-display.js
D13ce: Thanks for the explanation, I understand where you are coming from now.

I have talked with @MauserBitfly about this and he proposed another approach:
You could use `hidecharts` whenever the chart is shown and a change in the validator set causes the chart to not have any data to show:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/341fbb1783892d2ad0bda7dcf996aeeb21386ff4/static/js/dashboard.js#L1146
This unfortunately would also hide the incomeChart. You could easily solve this by doing something like (pseudocode)
```
hidecharts(){
   hideIncomeChart()
   hideProposedChart()
}

hideIncomeChart(){
   // ...
}

hideProposedChart(){
   // ...
}
```
And then you only call `hideProposedChart` in the scenario described.

This would neglect the need to add a new highcharts file (`no-data-to-display.js`) that introduces Linter issues and use already existing handling.

static/js/dashboard.js
```diff
@@ -1178,8 +1185,13 @@ $(document).ready(function () {
       url: "/dashboard/data/proposals" + qryStr,
       success: function (result) {
         var t1 = Date.now()
+        var chart = $("#proposed-chart").highcharts()
```
D13ce, line 1188: **Minor:** You don't need this here, it would be better placed between lines 1191 and 1192, would it not?
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/709974d262b28421debbd08243ad3a03ca6f575f/static/js/dashboard.js#L1191-L1192

D13ce (commented):
This approach is looking really good, well done (shout-out to @MauserBitfly ). I have left a minor comment though, please reply before I approve this PR. Cheers!

PR #2250, opened 2023-05-25
(BIDS-1835) Don't count the same slot multiple times

Eisei24:
The changes only apply to the frontend-data-updater.


D13ce (approved):
I really like this solution, well done. lgtm

PR #2253, opened 2023-05-26
Bids 2121/overlapping proposal graphics

KostasGiannakakis:
### <samp>🤖 Generated by Copilot at 276a4c1</samp>

Fix UI bug with proposal switch element on dashboard page. Adjust switch position based on `static/js/dashboard.js` file changes.

### DIFF
--- a/static/js/dashboard.js
+++ b/static/js/dashboard.js
@@ -331,28 +331,28 @@ function switchFrom(el1, el2, el3, el4) {
 
 var firstSwitch = true
 
-if($("#proposals-table_wrapper").length === 0) {
-  $(".proposal-switch").css( "position", "static" )
+if ($("#proposals-table_wrapper").length === 0) {
+  $(".proposal-switch").css("position", "static")
 }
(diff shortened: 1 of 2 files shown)


static/js/dashboard.js
```diff
@@ -331,12 +331,40 @@ function switchFrom(el1, el2, el3, el4) {
 
 var firstSwitch = true
 
+if ($("#proposals-table_wrapper").length === 0) {
+  $(".proposal-switch").css("position", "static")
+}
+
+function waitForTable(selector) {
```
MauserBitfly, line 338: Although the solution with the returned Promise looks elegant, it leads to a flickering when you add your first validator (if none was selected before) 

I would suggest a more simple solution: Set the display of the switch initially to 'none' and when we draw the table createProposedChart we also set remove the display none from the switch. 

There is not need to display the switch toggle unless we have data

MauserBitfly (approved):
LGTM

PR #2257, opened 2023-05-30
(BIDS-2030) fix big int on el performance

MauserBitfly:
copilot:summary

### DIFF
--- a/handlers/api_eth1.go
+++ b/handlers/api_eth1.go
@@ -945,8 +945,8 @@ func getValidatorExecutionPerformance(queryIndices []uint64) ([]types.ExecutionP
 	}
 
 	type LongPerformanceResponse struct {
-		Performance365d  int64  `db:"el_performance_365d" json:"performance365d"`
-		PerformanceTotal int64  `db:"el_performance_total" json:"performanceTotal"`
+		Performance365d  string `db:"el_performance_365d" json:"performance365d"`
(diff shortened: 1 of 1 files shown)


handlers/api_eth1.go
```diff
@@ -955,19 +955,21 @@ func getValidatorExecutionPerformance(queryIndices []uint64) ([]types.ExecutionP
...
-		CAST(COALESCE(mev_performance_total, 0) AS bigint) AS el_performance_total
+		CAST(COALESCE(mev_performance_365d, 0) AS text) AS el_performance_365d,
+		CAST(COALESCE(mev_performance_total, 0) AS text) AS el_performance_total
 		FROM validator_performance WHERE validatorindex = ANY($1)`, validatorsPQArray)
 	if err != nil {
 		return nil, fmt.Errorf("error can cl performance from db: %w", err)
 	}
 	for _, val := range performanceList {
+		performance365d, _ := big.NewInt(0).SetString(val.Performance365d, 10)
+		performanceTotal, _ := big.NewInt(0).SetString(val.PerformanceTotal, 10)
```
Eisei24, line 966: There is no need to set the value to 0 first, you can use
```suggestion
		performance365d, _ := new(big.Int).SetString(val.Performance365d, 10)
		performanceTotal, _ := new(big.Int).SetString(val.PerformanceTotal, 10)
```

PR #2262, opened 2023-06-01
(BIDS-2046) Add length button to validators table

Eisei24:
When reviewing this please also check the bugfix mentioned in the comments of the BIDS.

### DIFF
--- a/static/js/dashboard.js
+++ b/static/js/dashboard.js
@@ -419,6 +419,8 @@ $(document).ready(function () {
     },
     pageLength: 10,
     pagingType: "full_numbers",
+    scrollY: "503px",
+    scrollCollapse: true,
     info: false,
     language: {
       search: "",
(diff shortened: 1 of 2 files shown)


D13ce (changes requested):
I have tested the implementation and talked about it with @Buttaa, please implement the following changes:

Make both the table's header (i.e. the line that states the columns' names) and the footer (i.e. the "Show X entries" and the navigation buttons) fixed so that only the table's content (i.e. the lines representing the table's information) scroll.

If it is unclear what I mean, you know where I sit (spoiler, it is right next to you!) :)

(FYI, I have not reviewed your code yet. I will do so after your changes)

static/js/dashboard.js
```diff
@@ -431,6 +440,7 @@ $(document).ready(function () {
       }
     },
     drawCallback: function (settings) {
+      formatTimestamps()
```
Eisei24, line 443: This is the bug fix for the "Last Attestation" display on pages >1.

D13ce (approved):
lgtm

PR #2267, opened 2023-06-02
Bids 1989/validator page replace root hash with cl reward

Eisei24:
### <samp>🤖 Generated by Copilot at 7ba6e40</samp>

No summary available (Limit exceeded: required to process 53350 tokens, but only 50000 are allowed per call)


Eisei24:
Something went wrong, I did not want to merge the branch into master but into feature/newfrontend instead.

PR #2268, opened 2023-06-02
Bids 1989/validator page replace root hash with cl reward

Eisei24:
### <samp>🤖 Generated by Copilot at 7ba6e40</samp>

This pull request improves the performance and accuracy of querying validator income data from bigtable, and displays the proposer reward for each block proposed by a validator in the validator details page. It also removes an unused field and fixes a visual bug in the templates.

### DIFF
--- a/handlers/validator.go
+++ b/handlers/validator.go
@@ -1096,6 +1096,23 @@ func ValidatorProposedBlocks(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	epochCount := make(map[uint64]uint64)
+	for _, slot := range proposedSlots {
+		epochCount[slot.Slot/utils.Config.Chain.Config.SlotsPerEpoch]++
+	}
+
+	epochs := make([]uint64, 0, len(epochCount))
+	for key := range epochCount {
+		epochs = append(epochs, key)
+	}
+
+	incomeData, err := db.BigtableClient.GetValidatorIncomeDetailsIndexedMultiple([]uint64{index}, epochs)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


handlers/validator.go
```diff
@@ -1127,13 +1130,26 @@ func ValidatorProposedBlocks(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, "Internal server error", http.StatusInternalServerError)
 			return
 		}
+
+		epochs := make([]uint64, 0, len(epochCount))
+		for key := range epochCount {
+			epochs = append(epochs, key)
+		}
+
+		incomeData, err = db.BigtableClient.GetValidatorIncomeDetailsIndexedMultiple([]uint64{index}, epochs)
```
LuccaBitfly, line 1139: Why are CL Rewards only shown if EL rewards are present? CL proposer reward for slots exist regardless of whether the chain already has an EL, right?

LuccaBitfly (changes requested):

handlers/validator.go
Eisei24: Oh, those did not work as I expected them to.
I fixed this so now there are 3 possibilities for the CL reward:

- Pending if the epoch is not yet finalized
Also a minute for the latest finalized epoch since the data is not immediately in the bigtable
- The ETH value from the bigtable
0 ETH is possible
- "-" if it is an old epoch and there is no data in the bigtable

I used validator 15500 on prater to check all of these.

LuccaBitfly (approved):
lgtm, don't know why old epochs have no CL reward data in bigtable but I guess it's the best we can do for now

PR #2270, opened 2023-06-05
(BIDS-1695) Don't show historical value on testnets

Eisei24:
### <samp>🤖 Generated by Copilot at 0419587</samp>

Renamed `GetHistoricPrice` to `GetHistoricalPrice` and added `chainId` parameter to support different deposit contract chains. Updated `Eth1TxData` type, `eth1tx.html` template, and `Eth1TransactionTx` handler to use the new function and field names. The purpose was to improve the accuracy and clarity of the eth1 transaction data on the explorer.

### DIFF
--- a/db/frontend.go
+++ b/db/frontend.go
@@ -520,7 +520,11 @@ func getMachineStatsGap(resultCount uint64) int {
 	return 1
 }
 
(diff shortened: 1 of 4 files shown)


D13ce (approved):
lgtm

PR #2272, opened 2023-06-05
(BIDS-2146) fix clipping buttons in small screens

KostasGiannakakis:
### <samp>🤖 Generated by Copilot at 308a45f</samp>

Added responsive option to Highcharts configuration in `genericchart.html` to improve chart usability and appearance on smaller screens. This affects various charts for the eth2-beaconchain-explorer website.


templates/genericchart.html
```diff
@@ -63,6 +63,22 @@
         })
         {{else}}
         Highcharts.stockChart('chart', {
+            responsive: {
+            rules: [{
+                condition: {
+                maxWidth: 400,
```
MauserBitfly, line 69: please indent maxWith

MauserBitfly (commented):
The button is now visible, but I would prefer the buttons and the input to be aligned center with the headlines
image

Not sure if this can be achieved. If you don't find a solution (let's say after 1 hour) then we go with your version.

templates/genericchart.html
KostasGiannakakis: @MauserBitfly I took a look, and we won't be able to make it more dynamic that what currently is. For a better handling of these buttons we would need to use a newer version of the highcharts js library. Imo since it is at least usable now, let's leave it like that, and we can consider an update of the library later.

MauserBitfly (approved):

PR #2273, opened 2023-06-05
(BIDS-2175) fix the pagination and Last page

KostasGiannakakis:
### <samp>🤖 Generated by Copilot at 6701c09</samp>

Fixed a bug in `slots.html` that prevented the search filter from applying correctly. Used the DataTables API to set the search value after the table initialization.

### DIFF
--- a/templates/slots.html
+++ b/templates/slots.html
@@ -60,10 +60,11 @@
       }
       var usp = new URLSearchParams(window.location.search)
       var q = usp.get("q")
+      var t = $("#slots").DataTable(tblOpts)
       if (q) {
-        tblOpts.search = { search: q }
+        t.search(q);
+        t.draw();
       }
-      $("#slots").DataTable(tblOpts)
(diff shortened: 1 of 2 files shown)


templates/slots.html
```diff
@@ -60,10 +60,11 @@
       }
       var usp = new URLSearchParams(window.location.search)
       var q = usp.get("q")
+      var t = $("#slots").DataTable(tblOpts)
```
D13ce, line 63: **Minor:** This should be placed within the if below.

D13ce (commented):
Changes look good, I have left a minor suggestion though. ptal

templates/slots.html
KostasGiannakakis: well, not really. This line actually creates the Datatable the first time. The initial draw. Then we check if we had anything as a query parameter and if yes, we redraw the table with the `t.draw()` using the new search parameter(s)

D13ce (approved):
lgtm

PR #2280, opened 2023-06-06
(BIDS-2117) Implement queue of pending validators in dashboard

Eisei24:
### <samp>🤖 Generated by Copilot at 6a8d449</samp>

This pull request enhances the dashboard with new information and features for pending validators. It adds a new column to the table that shows the queue position and the estimated activation time for each pending validator, based on the validator index, the churn rate, and the latest epoch. It also improves the user interface by fixing some tooltip and table issues. It modifies the files `handlers/dashboard.go` and `static/js/dashboard.js` to implement these changes.

### DIFF
--- a/static/js/dashboard.js
+++ b/static/js/dashboard.js
@@ -321,60 +321,43 @@ var firstSwitch = true
 function initValidatorCountdown(validatorIndex, queueId, ts) {
   var now = Math.round(new Date().getTime() / 1000)
   var secondsLeft = ts - now
-  var seconds = secondsLeft % 60
-  var min = ((secondsLeft - seconds) / 60) % 60
-  var hour = Math.round((secondsLeft - seconds - min * 60) / 3600) % 24
-  var days = Math.round((secondsLeft - seconds - min * 60 - hour * 60 * 60) / 86400)
-  setValidatorCountdown(validatorIndex, queueId, days, hour, min, seconds)
+  setValidatorCountdown(validatorIndex, queueId, secondsLeft)
 
   if (!countdownIntervals.has(validatorIndex)) {
     countdownIntervals.set(
       validatorIndex,
       setInterval(function () {
-        if (hour <= 0 && min <= 0 && seconds <= 0 && days <= 0) {
+        if (secondsLeft <= 0) {
           clearInterval(countdownIntervals.get(validatorIndex))
           return
         }
 
-        if (seconds === 0 && min === 0 && hour === 0 && days > 0) {
-          hour = 24
-          days -= 1
-        }
-
-        if (seconds === 0 && min === 0 && hour > 0) {
-          min = 60
-          hour -= 1
-        }
-
-        if (seconds === 0 && min > 0) {
-          seconds = 60
-          min -= 1
-        }
-
-        seconds -= 1
-
-        setValidatorCountdown(validatorIndex, queueId, days, hour, min, seconds)
+        secondsLeft -= 1
+        setValidatorCountdown(validatorIndex, queueId, secondsLeft)
       }, 1000)
     )
   }
 }
 
-function setValidatorCountdown(validatorIndex, queueId, days, hour, min, second) {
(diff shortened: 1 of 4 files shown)


static/js/dashboard.js
```diff
@@ -420,7 +511,6 @@ $(document).ready(function () {
     pageLength: 10,
     pagingType: "full_numbers",
     scrollY: "503px",
-    scrollCollapse: true,
```
Eisei24, line 423: This fixes an unrelated bug I found:
If you only have one validator in the dashboard the tooltip of "Last Attestation" has no space and automatically toggles on/off every now and then.

static/js/dashboard.js
```diff
@@ -317,6 +318,88 @@ function switchFrom(el1, el2, el3, el4) {
...
+        }
+
+        if (seconds === 0 && min > 0) {
+          seconds = 60
+          min -= 1
+        }
+
+        seconds -= 1
+
+        setValidatorCountdown(validatorIndex, queueId, days, hour, min, seconds)
```
MauserBitfly, line 356: Instead of passing days/hour/... why not simply reduce the secondsLeft by the interval time and pass the secondsLeft. This way you could remove all the complecated -= logic above. 

And only format the secondsLeft (via the luxon duration object) when you need it for creating the tooltip.

static/js/dashboard.js
```diff
@@ -317,6 +318,88 @@ function switchFrom(el1, el2, el3, el4) {
...
+  var hour = Math.round((secondsLeft - seconds - min * 60) / 3600) % 24
+  var days = Math.round((secondsLeft - seconds - min * 60 - hour * 60 * 60) / 86400)
+  setValidatorCountdown(validatorIndex, queueId, days, hour, min, seconds)
+
+  if (!countdownIntervals.has(validatorIndex)) {
+    countdownIntervals.set(
+      validatorIndex,
+      setInterval(function () {
+        if (hour <= 0 && min <= 0 && seconds <= 0 && days <= 0) {
+          clearInterval(countdownIntervals.get(validatorIndex))
```
MauserBitfly, line 335: Not sure if its nice to just remove it (if it was displayed before) from a user experience... maybe write some text "Your Validator should be active any moment now" ... maybe trigger a refresh of the data or add a 'refresh' button to the tooltip?

MauserBitfly (commented):
Works good, just left some comments to reduce some complexity.

static/js/dashboard.js
Eisei24: I don't remove the tooltip I just stop the interval since there is nothing to update if it shows all times as 0.

I implemented this like the validator page does it (except for the rocket). When you reach a time of 0 it will just remain like that. Even if you reload the page it is not immediately active since all it means is that "We have reached the epoch where the validator will be activated". It can still take a few minutes to become active.

But I agree that it looks weird when you approach the end of the time for a few reasons (e.g. a few validators can all be #1 in the queue since they will all be handled in the same epoch) so I will discuss that in the BIDS.

static/js/dashboard.js
```diff
@@ -317,6 +318,88 @@ function switchFrom(el1, el2, el3, el4) {
 
 var firstSwitch = true
 
+function initValidatorCountdown(validatorIndex, queueId, ts) {
+  var now = Math.round(new Date().getTime() / 1000)
+  var secondsLeft = ts - now
+  var seconds = secondsLeft % 60
+  var min = ((secondsLeft - seconds) / 60) % 60
+  var hour = Math.round((secondsLeft - seconds - min * 60) / 3600) % 24
```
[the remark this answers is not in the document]
Eisei24: See the next comment.

static/js/dashboard.js
Eisei24: As written in the BIDS showing “This validator is currently #1 in Queue” with zero time left is fine.

MauserBitfly (approved):
LGTM

PR #2281, opened 2023-06-06
(BIDS-2144) Convert byte slice to string the correct way

Eisei24:
### <samp>🤖 Generated by Copilot at 5e53d81</samp>

The pull request enhances the validator income history API endpoint to include more details about the rewards and penalties that affect a validator's income. It introduces new types in the `types` package and uses them in the `handlers` package. It also removes some unused or obsolete code.

### DIFF
--- a/types/api.go
+++ b/types/api.go
@@ -568,16 +568,30 @@ type ApiValidatorEth1Response struct {
 }
 
 type ApiValidatorIncomeHistoryResponse struct {
-	Income struct {
-		AttestationSourceReward uint64 `json:"attestation_source_reward"`
-		AttestationTargetReward uint64 `json:"attestation_target_reward"`
-		AttestationHeadReward   uint64 `json:"attestation_head_reward"`
-	} `json:"income"`
-	Epoch          uint64    `json:"epoch"`
-	ValidatorIndex uint64    `json:"validatorindex"`
-	Week           uint64    `json:"week"`
-	WeekStart      time.Time `json:"week_start"`
-	WeekEnd        time.Time `json:"week_end"`
+	Income         *ApiValidatorIncomeHistory `json:"income"`
+	Epoch          uint64                     `json:"epoch"`
+	ValidatorIndex uint64                     `json:"validatorindex"`
+	Week           uint64                     `json:"week"`
+	WeekStart      time.Time                  `json:"week_start"`
+	WeekEnd        time.Time                  `json:"week_end"`
+}
+
+type ApiValidatorIncomeHistory struct {
+	AttestationSourceReward            uint64 `json:"attestation_source_reward,omitempty"`
+	AttestationSourcePenalty           uint64 `json:"attestation_source_penalty,omitempty"`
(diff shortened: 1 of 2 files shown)


types/api.go
```diff
@@ -568,16 +568,30 @@ type ApiValidatorEth1Response struct {
...
+	AttestationHeadReward              uint64 `json:"attestation_head_reward,omitempty"`
+	FinalityDelayPenalty               uint64 `json:"finality_delay_penalty,omitempty"`
+	ProposerSlashingInclusionReward    uint64 `json:"proposer_slashing_inclusion_reward,omitempty"`
+	ProposerAttestationInclusionReward uint64 `json:"proposer_attestation_inclusion_reward,omitempty"`
+	ProposerSyncInclusionReward        uint64 `json:"proposer_sync_inclusion_reward,omitempty"`
+	SyncCommitteeReward                uint64 `json:"sync_committee_reward,omitempty"`
+	SyncCommitteePenalty               uint64 `json:"sync_committee_penalty,omitempty"`
+	SlashingReward                     uint64 `json:"slashing_reward,omitempty"`
+	SlashingPenalty                    uint64 `json:"slashing_penalty,omitempty"`
+	TxFeeRewardWei                     string `json:"tx_fee_reward_wei,omitempty"`
```
D13ce, line 593: Why would this be a string and not, for example, a big.Int?
Then again, this would break the API as it was a []byte before. Thoughts?

If required: Could you please discuss this topic internally? We may want to add a new field, for example.

types/api.go
```diff
@@ -568,16 +568,30 @@ type ApiValidatorEth1Response struct {
...
-	WeekEnd        time.Time `json:"week_end"`
+	Income         *ApiValidatorIncomeHistory `json:"income"`
+	Epoch          uint64                     `json:"epoch"`
+	ValidatorIndex uint64                     `json:"validatorindex"`
+	Week           uint64                     `json:"week"`
+	WeekStart      time.Time                  `json:"week_start"`
+	WeekEnd        time.Time                  `json:"week_end"`
+}
+
+type ApiValidatorIncomeHistory struct {
```
D13ce, line 579: I am not sure whether I like duplicating the info here (compared with `itypes "github.com/gobitfly/eth-rewards/types"` where it came before) or not.
Changes in `itypes "github.com/gobitfly/eth-rewards/types"` now will not break the API but it feels redundant.
Could you please share your thoughts?

D13ce (commented):
ptal a look at my comments. I need your thoughts before approving this.

types/api.go
Eisei24: When you display a bigint like 27194623959527110306 as a number it is possible, depending on the way you display it, for it to get converted the wrong way to e.g. 27194623959527110000.
Returning a string prevents this.

D13ce: Now that I think about it, we cannot return a big.Int anyways as this would break the API.

D13ce (approved):
lgtm

PR #2297, opened 2023-06-12
(BIDS-2032) Use saved datatable info for blocks page

Eisei24:
I could not test the database use locally since it requires a normal login otherwise you don't have permission to write to the database for this use.


handlers/common.go
Eisei24, line 1: Changes in this file fix the bug mentioned in the BIDS`s comments.

cmd/explorer/main.go
```diff
@@ -530,7 +530,8 @@ func main() {
 			router.HandleFunc("/tools/broadcast", handlers.BroadcastPost).Methods("POST")
 			router.HandleFunc("/tools/broadcast/status/{jobID}", handlers.BroadcastStatus).Methods("GET")
 
-			router.HandleFunc("/tables/state", handlers.DataTableStateChanges).Methods("POST")
+			router.HandleFunc("/tables/getstate", handlers.GetDataTableStateChanges).Methods("POST")
+			router.HandleFunc("/tables/setstate", handlers.SetDataTableStateChanges).Methods("POST")
```
LuccaBitfly, line 534: The table whose state is stored should be identified in the URI. This would remove the necessity for the key to be in the body of requests, which allows you to use the `GET` method for get requests and remove the error handling for the key inside the handler for both endpoints.
Also, verbs should be avoided in REST API URIs.
Also also, if a resource can be updated multiple times through the same endpoint `PUT` is a more appropriate method than `POST`. 
```suggestion
			router.HandleFunc("/tables/{tableId}/state", handlers.GetDataTableStateChanges).Methods("GET")
			router.HandleFunc("/tables/{tableId}/state", handlers.SetDataTableStateChanges).Methods("PUT")
```

Other parts of the code have to be changed accordingly.

LuccaBitfly (changes requested):

cmd/explorer/main.go
Eisei24: Thank you for the improvements, I will keep them in mind for the future.
Everything implemented.

Eisei24:
As discussed I also adjusted the withdrawals page so that if a query parameter is given in the URI the search for both tables will be set to that.
Keep in mind that if you reload or revisit the normal page /validators/withdrawals the two tables can still have different searches if one of them couldn't finish loading the last search.

LuccaBitfly (changes requested):
I have noticed that if an error happens inside the endpoint handlers the response code is still 200 (to test is you can mock a logged-in user and look at the traffic).
image

image

If an internal error happens the endpoint should return the status code 500.

Eisei24:
I reworked the whole error handling for the 2 methods, please have a look.

Normally we use
`http.Error(w, "Internal server error", http.StatusServiceUnavailable)`
which changes more than just the header but is the implemented
`w.WriteHeader(http.StatusInternalServerError)`
also fine so that we get the `response`?

LuccaBitfly (approved):
Changing it with `w.WriteHeader` is fine, lgtm

PR #2299, opened 2023-06-12
(BIDS-2193) add where clause to avoid null results

KostasGiannakakis:
copilot:summary


KostasGiannakakis:
It seems to me that as of now it is sufficient to filter out any rows that contain a NULL validatorIndex. From my understanding there is some difference between the time we receive the block and when we "connect" the block with the proposer. I searched in the DB and I wasn't able to find a case where we have NULL validator. If, by the reviewers perspective, is better to try and fix the underlying issue I can take a look at it, but from my search I wasn't able to pinpoint the exact location of this discrepancy

MauserBitfly (approved):

MauserBitfly:
LGTM

PR #2301, opened 2023-06-12
(BIDS-180) Send the report when all the data is available

Eisei24:
I could not test whether the changes fix the issue since it only happens once a month.

See the comments in the BIDS for my reasoning for the changes.

### DIFF
--- a/services/notifications.go
+++ b/services/notifications.go
@@ -2554,7 +2554,7 @@ func collectTaxReportNotificationNotifications(notificationsByUserID map[uint64]
 	err = db.FrontendWriterDB.Select(&dbResult, `
 			SELECT us.id, us.user_id, us.created_epoch, us.event_filter, ENCODE(us.unsubscribe_hash, 'hex') as unsubscribe_hash
 			FROM users_subscriptions AS us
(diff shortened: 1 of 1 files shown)


services/notifications.go
```diff
@@ -2528,48 +2528,61 @@ func (n *taxReportNotification) GetInfoMarkdown() string {
...
-		}
+	if tNow.Year() != firstDayOfMonth.Year() || tNow.Month() != firstDayOfMonth.Month() || tNow.Day() != firstDayOfMonth.Day() { // Send the reports on the first day of the month
+		return nil
+	}
 
-		err := db.FrontendWriterDB.Select(&dbResult, `
+	lastStatsDay, err := db.GetLastExportedStatisticDay()
+	if err != nil {
+		return err
+	}
```
D13ce, line 2539: If the statistics export is delayed for a whole day, the user rewards for this month will never be sent because of the if above in line 2532. Is that correct?

If so, this should be addressed.

D13ce (commented):
ptal

services/notifications.go
Eisei24: Yes.

I implemented your requested change. This also "fixes" the weird behaviour that if the subscription was created on the first of the month the user would immediately get sent the report of the previous month.
Assuming nothing further along in the code stops that.

PR #2303, opened 2023-06-12
(BIDS-2191) add activationeligibilityepoch to validator_queue_

MauserBitfly:
copilot:summary

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -1818,7 +1818,12 @@ func updateQueueDeposits() error {
 	// then we add any new ones that are queued
 	_, err = WriterDb.Exec(`
 		INSERT INTO validator_queue_deposits
-		SELECT validatorindex FROM validators WHERE activationepoch=9223372036854775807 and status='pending' ON CONFLICT DO NOTHING`)
+		SELECT validatorindex, activationeligibilityepoch 
+		FROM validators 
+		WHERE activationepoch=9223372036854775807 and status='pending' 
+		ON CONFLICT (validatorindex) DO UPDATE SET
+			activationeligibilityepoch = EXCLUDED.activationeligibilityepoch
+		`)
 	if err != nil {
 		logger.Errorf("error adding queued publickeys to validator_queue_deposits: %v", err)
 		return err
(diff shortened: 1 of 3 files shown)


handlers/validator.go
```diff
@@ -544,7 +544,8 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 	})
 
 	g.Go(func() error {
-		if validatorPageData.ActivationEpoch > 100_000_000 {
+		// we only need to get the queue information if the action epoch is in the future and we have an eligibility epoch
```
Eisei24, line 547: **nitpick:** The first part of the comment could be confusing. I would write it as something like
"// we only need to get the queue information if we don't have an activation epoch but we have an eligibility epoch"

db/migrations/tbd_activationeligibilityepoch_for_validator_queue_deposits.sql
```diff
@@ -0,0 +1,19 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'add column activationeligibilityepoch to validator_queue_deposits';
+ALTER TABLE validator_queue_deposits ADD COLUMN IF NOT EXISTS activationeligibilityepoch BIGINT NOT NULL;
+
+SELECT 'pupulate activationeligibilityepoch data on validator_queue_deposits from validators table';
```
Eisei24, line 6: **typo:** pupulate => populate

db/db.go
```diff
@@ -1818,7 +1818,12 @@ func updateQueueDeposits() error {
...
-		SELECT validatorindex FROM validators WHERE activationepoch=9223372036854775807 and status='pending' ON CONFLICT DO NOTHING`)
+		SELECT validatorindex, activationeligibilityepoch 
+		FROM validators 
+		WHERE activationepoch=9223372036854775807 and status='pending' 
+		ON CONFLICT (validatorindex) DO UPDATE SET
+			activationeligibilityepoch = EXCLUDED.activationeligibilityepoch
+		`)
 	if err != nil {
 		logger.Errorf("error adding queued publickeys to validator_queue_deposits: %v", err)
 		return err
```
Eisei24, line 1829: Don't you also need to set the **activationeligibilityepoch** when you set the **block_slot** and **block_index** 5 lines below this?

db/db.go
```diff
@@ -1866,23 +1871,16 @@ func updateQueueDeposits() error {
 func GetQueueAheadOfValidator(validatorIndex uint64) (uint64, error) {
```
Eisei24, line 1871: **praise:** This is a good restructuring to not only reduce the load time but also make it much more readable.

Eisei24 (changes requested):
Please fix the query typo which breaks it.

handlers/validator.go
MauserBitfly: yes yours sounds better

db/migrations/tbd_activationeligibilityepoch_for_validator_queue_deposits.sql
MauserBitfly: fixed

db/db.go
```diff
@@ -1866,23 +1871,16 @@ func updateQueueDeposits() error {
...
 	var res uint64
 	err := ReaderDb.Get(&res, `
-	with SelectedValidator as (
-		select block_slot, block_index, validators.activationeligibilityepoch from validator_queue_deposits vqd
-		inner join validators on validators.validatorindex = $1
-		where vqd.validatorindex = validators.validatorindex 
+	WITH SelectedValidator AS (
+		SELECT block_slot, block_index, activationeligibilityepoch FROM validator_queue_deposits
+		WHERE validatorindex = %1
```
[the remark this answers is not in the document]
MauserBitfly: good catch!

db/db.go
MauserBitfly: thanks ;-)

Eisei24 (approved):
utACK

I don't have permissions to adjust the db for testing but everything looks fine to me.

guybrush (approved):

PR #2314, opened 2023-06-15
(NOBIDS) use redis to cache the last block in the block / data table

peterbitfly:
After merging the eth1indexer needs to be updated alongside the frontend

### <samp>🤖 Generated by Copilot at 28d33e5</samp>

This pull request adds a redis cache for the last block numbers in the eth1 blocks and data tables, and updates it after indexing new data from the node. This reduces the need to query the bigtable, which improves the performance and cost of the eth1 indexer. The changes affect the files `db/bigtable_eth1.go` and `cmd/eth1indexer/main.go`.

### DIFF
--- a/cmd/eth1indexer/main.go
+++ b/cmd/eth1indexer/main.go
@@ -716,7 +716,25 @@ func IndexFromNode(bt *db.Bigtable, client *rpc.ErigonClient, start, end, concur
 
 	}
 
-	return g.Wait()
+	err := g.Wait()
+
+	if err != nil {
+		return err
+	}
+
+	lastBlockInCache, err := bt.GetLastBlockInBlocksTable()
+	if err != nil {
+		return err
+	}
+
+	if end > int64(lastBlockInCache) {
+		err := bt.SetLastBlockInBlocksTable(end)
+
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func ImportMainnetERC20TokenMetadataFromTokenDirectory(bt *db.Bigtable) {
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -230,8 +230,10 @@ func (bigtable *Bigtable) GetLastBlockInBlocksTable() (int, error) {
 	res, err := bigtable.redisCache.Get(ctx, redisKey).Result()
 
 	if err != nil {
+		// key is not yet set, get data from bigtable
+		// key will be set via the eth1indexer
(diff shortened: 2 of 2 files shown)


cmd/eth1indexer/main.go
```diff
@@ -716,7 +716,25 @@ func IndexFromNode(bt *db.Bigtable, client *rpc.ErigonClient, start, end, concur
...
+	if err != nil {
+		return err
+	}
+
+	lastBlockInCache, err := bt.GetLastBlockInBlocksTable()
+	if err != nil {
+		return err
+	}
+
+	if end > int64(lastBlockInCache) {
```
MauserBitfly, line 730: This will not work if redis has not yet been set.
As GetLastBlockInBlocksTable will return the same block that has just been added to to bt by the importer so end will be the same as lastBlockInCache

D13ce (changes requested):

MauserBitfly (approved):
LGTM

db/bigtable_eth1.go
```diff
@@ -222,32 +223,43 @@ func (bigtable *Bigtable) CheckForGapsInBlocksTable(lookback int) (gapFound bool
...
+	res, err := bigtable.redisCache.Get(ctx, redisKey).Result()
 
-		if err != nil {
-			logger.Errorf("error parsing block number from key %v: %v", r.Key(), err)
-			return false
-		}
-		c = max_block_number - c
+	if err != nil {
+		// key is not yet set, get data from bigtable
+		// key will be set via the eth1indexer
```
D13ce, line 234: This comment became obsolete via the last commit. Please remove it.
Also applies here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/e4fd81eed01e0ca45a6caaea63a9b62ced109053/db/bigtable_eth1.go#L309-310

cmd/eth1indexer/main.go
```diff
@@ -716,7 +716,25 @@ func IndexFromNode(bt *db.Bigtable, client *rpc.ErigonClient, start, end, concur
...
 	}
 
-	return g.Wait()
+	err := g.Wait()
+
+	if err != nil {
+		return err
+	}
+
+	lastBlockInCache, err := bt.GetLastBlockInBlocksTable()
```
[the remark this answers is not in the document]
D13ce: Since with the latest comment the return value is automatically cached, the name technically is correct now. Resolving this conversation.

D13ce (changes requested):

PR #2318, opened 2023-06-16
(BIDS-2204) Fix null issue for last attestation slot

Eisei24:
The query threw an error because it tried to save NULL in int64.

See the linked NOBIDS in the comments which probably caused this bug.
To me it looked like we saved 0 instead of NULL before these changes.


D13ce (approved):
lgtm

PR #2320, opened 2023-06-16
(BIDS-2142) fix event value parsing

MauserBitfly:
### <samp>🤖 Generated by Copilot at de48be7</samp>

Fix eth1 transaction value display for non-string types. Use `fmt.Sprintf` to convert values with non-string type prefixes in `eth1data/eth1data.go`.

### DIFF
--- a/eth1data/eth1data.go
+++ b/eth1data/eth1data.go
@@ -217,6 +217,10 @@ func GetEth1Transaction(hash common.Hash) (*types.Eth1TxData, error) {
 							if strings.HasPrefix(b, "byte") {
 								a.Value = a.Raw
 							}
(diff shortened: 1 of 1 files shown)


eth1data/eth1data.go
Eisei24, line 1: **Suggestion:** Can't we just replace the `%s` with an `%v` in line 211 instead?

Eisei24 (changes requested):
Please have a look at my suggestion.

eth1data/eth1data.go
MauserBitfly: You are right, that's easier and I did not find a case where it has a bad effect

Eisei24 (approved):
lgtm

PR #2329, opened 2023-06-19
(BIDS-2123) vali page: fix income chart dates

LuccaBitfly:
### <samp>🤖 Generated by Copilot at 225a62c</samp>

This pull request enhances the validator income chart by removing data grouping, changing tooltip text, and using `epochToTime` to display timestamps.

### DIFF
--- a/templates/validator/charts.html
+++ b/templates/validator/charts.html
@@ -190,12 +190,6 @@
                                     enabled: false
                                 },
                                 pointInterval: 24 * 3600 * 1000,
-                                dataGrouping: {
-                                    forced: true,
-                                    units: [
-                                        ['day', [1]]
-                                    ]
-                                }
                             }
                         },
                         xAxis: {
@@ -242,7 +236,8 @@
                                 const startEpoch = timeToEpoch(tooltip.chart.hoverPoint.x)
                                 const timeForOneDay = 24 * 60* 60 * 1000
                                 const endEpoch = timeToEpoch(tooltip.chart.hoverPoint.x + timeForOneDay) - 1
-                                text += `${new Date(tooltip.chart.hoverPoints[0].x).toLocaleDateString()} (Epochs ${startEpoch} - ${endEpoch})<br/>`;
+                                const endTs = epochToTime(endEpoch)
+                                text += `${new Date(tooltip.chart.hoverPoints[0].x).toLocaleDateString()} - ${new Date(endTs).toLocaleDateString()} <br> Epochs ${startEpoch} - ${endEpoch}<br/>`;
 
                                 // income
                                 for (var i = 0; i < tooltip.chart.hoverPoints.length; i++) {
--- a/templates/validator/charts.html
+++ b/templates/validator/charts.html
@@ -237,9 +237,8 @@
                                 const timeForOneDay = 24 * 60* 60 * 1000
                                 const endEpoch = timeToEpoch(tooltip.chart.hoverPoint.x + timeForOneDay) - 1
                                 const startDate = luxon.DateTime.fromMillis(tooltip.chart.hoverPoints[0].x)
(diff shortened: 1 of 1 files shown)


templates/validator/charts.html
```diff
@@ -242,7 +236,8 @@
                                 const startEpoch = timeToEpoch(tooltip.chart.hoverPoint.x)
                                 const timeForOneDay = 24 * 60* 60 * 1000
                                 const endEpoch = timeToEpoch(tooltip.chart.hoverPoint.x + timeForOneDay) - 1
-                                text += `${new Date(tooltip.chart.hoverPoints[0].x).toLocaleDateString()} (Epochs ${startEpoch} - ${endEpoch})<br/>`;
+                                const endTs = epochToTime(endEpoch)
+                                text += `${new Date(tooltip.chart.hoverPoints[0].x).toLocaleDateString()} - ${new Date(endTs).toLocaleDateString()} <br> Epochs ${startEpoch} - ${endEpoch}<br/>`;
```
Eisei24, line 240: **issue:** According to the description/comments we do not just want to show the date in the tooltip but the exact time for the network.

templates/validator/charts.html
```diff
@@ -190,12 +190,6 @@
                                     enabled: false
                                 },
                                 pointInterval: 24 * 3600 * 1000,
-                                dataGrouping: {
```
Eisei24, line 193: **issue:** I don't think this leads to an x-axis labeling as desired.
Now the marks on the x-axis are placed when a new calendar day starts and don't match with the bars.
The epochs are also still the same with the first entry on the x-axis on prater is "24. Mar Epoch 84".

However maybe this design is acceptable, if you want it can be discussed.

Eisei24 (changes requested):
Please have a look at my issues.

templates/validator/charts.html
LuccaBitfly: fixed

templates/validator/charts.html
```diff
@@ -236,8 +236,10 @@
                                 const startEpoch = timeToEpoch(tooltip.chart.hoverPoint.x)
                                 const timeForOneDay = 24 * 60* 60 * 1000
                                 const endEpoch = timeToEpoch(tooltip.chart.hoverPoint.x + timeForOneDay) - 1
-                                const endTs = epochToTime(endEpoch)
-                                text += `${new Date(tooltip.chart.hoverPoints[0].x).toLocaleDateString()} - ${new Date(endTs).toLocaleDateString()} <br> Epochs ${startEpoch} - ${endEpoch}<br/>`;
+                                const startDate = luxon.DateTime.fromMillis(tooltip.chart.hoverPoints[0].x)
+                                const endDate = luxon.DateTime.fromMillis(epochToTime(endEpoch))
+                                const endTime = luxon.DateTime.fromMillis(epochToTime(endEpoch+1))
+                                text += `${startDate.toFormat("MMM-dd-yyyy HH:mm:ss")} - ${endDate.toFormat("MMM-dd-yyyy")} ${endTime.toFormat("HH:mm:ss")} <br> Epochs ${startEpoch} - ${endEpoch}<br/>`;
```
Eisei24, line 242: Why do we have `endDate` and `endTime`?
Isn't 
```suggestion
                                const startDate = luxon.DateTime.fromMillis(tooltip.chart.hoverPoints[0].x)
                                const endDate = luxon.DateTime.fromMillis(epochToTime(endEpoch+1))
                                text += `${startDate.toFormat("MMM-dd-yyyy HH:mm:ss")} - ${endDate.toFormat("MMM-dd-yyyy HH:mm:ss")} <br> Epochs ${startEpoch} - ${endEpoch}<br/>`;
```
always correct?

LuccaBitfly: fixed

Eisei24 (approved):
lgtm

PR #2331, opened 2023-06-19
(BIDS-2184) Fix search on withdrawals page

Eisei24:
### <samp>🤖 Generated by Copilot at 9e87cca</samp>

This pull request simplifies and improves the code for searching withdrawals and BLS changes in the `eth2-beaconchain-explorer` database. It extracts common logic into the `db` package and removes unnecessary prefix removal in the `handlers` package. It also cleans up some unused imports and code.


db/db.go
```diff
@@ -2267,41 +2269,46 @@ func GetWithdrawals(query string, length, start uint64, orderBy, orderDir string
...
+				WHERE w.validatorindex = $3
+					OR block_slot = $3
+					OR (block_slot / $5) = $3
+					OR ENCODE(address, 'hex') LIKE ($4 || '%')`
+
+			err := ReaderDb.Select(&withdrawals, fmt.Sprintf(withdrawalsQuery, searchQuery, orderBy, orderDir),
+				length, start, uiQuery, trimmedQuery, utils.Config.Chain.Config.SlotsPerEpoch)
+			if err != nil {
+				return nil, err
+			}
```
D13ce, line 2300: **Minor:** You should "centralize" the error handling by moving it down below the if starting at line `2287`. No need for it to be part of every possible if path when there is no unique error handling based on the path.

This also applies to `func GetBLSChanges`: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/9e87cca798e6d7d590ab4b459d9f5acbe1db6842/db/db.go#L2874

db/db.go
```diff
@@ -2267,41 +2269,46 @@ func GetWithdrawals(query string, length, start uint64, orderBy, orderDir string
...
+		INNER JOIN blocks b ON w.block_root = b.blockroot AND b.status = '1'
+		%s
+		ORDER BY %s %s
+		LIMIT $1
+		OFFSET $2`
+
+	trimmedQuery := strings.ToLower(strings.TrimPrefix(query, "0x"))
+	if trimmedQuery != "" {
+		searchQuery := ""
+		if uiQuery, err := strconv.ParseUint(query, 10, 64); err == nil {
```
D13ce, line 2289: **Minor:** This if could use a comment. The `else if` is fine as it utilizes a variable named `addressLikeRE` which is self-explanatory.

This also applies to `func GetBLSChanges`: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/9e87cca798e6d7d590ab4b459d9f5acbe1db6842/db/db.go#L2874

db/db.go
```diff
@@ -2267,41 +2269,46 @@ func GetWithdrawals(query string, length, start uint64, orderBy, orderDir string
...
+		FROM blocks_withdrawals w
+		INNER JOIN blocks b ON w.block_root = b.blockroot AND b.status = '1'
+		%s
+		ORDER BY %s %s
+		LIMIT $1
+		OFFSET $2`
+
+	trimmedQuery := strings.ToLower(strings.TrimPrefix(query, "0x"))
+	if trimmedQuery != "" {
+		searchQuery := ""
```
D13ce, line 2288: **Minor:** There is no need for this line, just use `:=` below.

This also applies to `func GetBLSChanges`: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/9e87cca798e6d7d590ab4b459d9f5acbe1db6842/db/db.go#L2874

D13ce (commented):
Changes work fine, I have left some minor code improvement suggestions though.

Furthermore, not everything that can be searched for is indexed in the according data tables. I suggest checking this to speed up the search. Maybe talk with @recy21 about that who may have an open PR adding various indexes.
If you see this outside the scope of this PR, please let me know.

db/db.go
Eisei24: Right, became obsolete through changes.

db/db.go
Eisei24: Added.

Eisei24:
As the BIDS comment says the `blocks_withdrawals` table is the relevant for the search since it is much bigger, millions of rows, but all the changes are also applied to the `blocks_bls_change` table.
The longest search you can have is if there are no results, e.g. **9951** or **25c45** on Sepolia.

I discussed it a bit with @recy21 and he suggested to add a new column that already has the byte value converted to a string.
This improves the query time somewhat.

Searching for a number like **9951** takes much longer since there are 4 WHERE in the query.
Without any indices it takes ~5m. Strangely what breaks the query is the ORDER/LIMIT/OFFSET at the end.
Using just one of those 3 still gives you a reasonable time of about 15s but having ORDER and LIMIT already takes 5m even though there are not even any resulting rows.

Searching for a hex value like **25c45** takes ~30s.

I added all three possible indices that help in the search

- address_text
- block_slot
- validatorindex, block_slot, address_text

Arithmetic operations are not possible for an index so there is no block_slot/32.

In the end I could improve the time for a search like **9951** to ~2m.

Please also consider whether it now takes too long to write the `blocks_withdrawals` table since we constantly add data to it.

db/migrations/20230710145000_add_withdrawals_indices.sql
D13ce, line 1: Don't add the timestamp when you create the file but when the PR is about to be merged. Otherwise we might introduce ordering problems if code reviews are delayed (like in this instance because of my vacation, sorry).

For now I suggest to keep the name as it is and before we merge, when this comment is the last one left, we rename the file accordingly.

db/db.go
D13ce: I have added a commit to centralize this a bit more. ptal

D13ce (changes requested):

db/migrations/20230710145000_add_withdrawals_indices.sql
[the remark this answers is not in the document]
Eisei24: Comments added.

db/db.go
```diff
@@ -2898,49 +2902,56 @@ func GetBLSChanges(query string, length, start uint64, orderBy, orderDir string)
...
-				OR CAST((block_slot / $5) as varchar) LIKE $3 || '%%'
-				OR val.deposit_adress LIKE $4::bytea || '%%'::bytea
-			ORDER BY bls.%s %s
-			LIMIT $1
-			OFFSET $2`, orderBy, orderDir), length, start, strings.ToLower(query), bquery, utils.Config.Chain.Config.SlotsPerEpoch)
+				validators.validatorindex as validatorindex,
+				eth1_deposits.from_address as deposit_adress
+			FROM validators 
+			INNER JOIN eth1_deposits ON validators.pubkey = eth1_deposits.publickey
+		) AS val ON val.validatorindex = bls.validatorindex
```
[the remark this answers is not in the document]
Eisei24: Oh yeah, it isn't needed. Changed.

D13ce (changes requested):

db/migrations/20230710145000_add_withdrawals_indices.sql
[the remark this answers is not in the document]
Eisei24: I also changed a few more words to be more consistent with other migration files.

D13ce (approved):
lgtm

peterbitfly:
LGTM

PR #2334, opened 2023-06-20
(BIDS-1826) vali history: remove pending for pre activation

LuccaBitfly:
Fixes the validator page history so that epochs before the validators activation don't show "pending" as their reward anymore. Also made a misc. fix for the validators page, as templating inside JS literals throws an error now.


Eisei24 (approved):
lgtm

Thanks for renaming `i` to `epoch`.

PR #2335, opened 2023-06-20
(BIDS-1838) dynamic change the offline/online detection limit

KostasGiannakakis:
### <samp>🤖 Generated by Copilot at 6fec741</samp>

Added configurable limits for online and offline validators notifications in `services/notifications.go` and `types/config.go`. This allows users to adjust the thresholds for triggering notifications based on their preferences.

### DIFF
--- a/services/notifications.go
+++ b/services/notifications.go
@@ -1526,25 +1526,21 @@ func collectAttestationAndOfflineValidatorNotifications(notificationsByUserID ma
 		}
 	}
 
-	var offlineValidatorsLimit int
-	if utils.Config.Notifications.OfflineDetectionLimit != nil {
-		offlineValidatorsLimit = *utils.Config.Notifications.OfflineDetectionLimit
-	} else {
-		offlineValidatorsLimit = 5000
+	offlineValidatorsLimit := 5000
+	if utils.Config.Notifications.OfflineDetectionLimit != 0 {
+		offlineValidatorsLimit = utils.Config.Notifications.OfflineDetectionLimit
 	}
 
-	var onlineValidatorsLimit int
-	if utils.Config.Notifications.OnlineDetectionLimit != nil {
-		onlineValidatorsLimit = *utils.Config.Notifications.OnlineDetectionLimit
-	} else {
-		onlineValidatorsLimit = 5000
+	onlineValidatorsLimit := 5000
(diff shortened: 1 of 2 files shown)


types/config.go
```diff
@@ -167,6 +167,8 @@ type Config struct {
 		FirebaseCredentialsPath                       string `yaml:"firebaseCredentialsPath" envconfig:"NOTIFICATIONS_FIREBASE_CRED_PATH"`
 		ValidatorBalanceDecreasedNotificationsEnabled bool   `yaml:"validatorBalanceDecreasedNotificationsEnabled" envconfig:"VALIDATOR_BALANCE_DECREASED_NOTIFICATIONS_ENABLED"`
 		PubkeyCachePath                               string `yaml:"pubkeyCachePath" envconfig:"NOTIFICATIONS_PUBKEY_CACHE_PATH"`
+		OnlineDetectionLimit                          *int   `yaml:"onlineDetectionLimit" envconfig:"ONLINE_DETECTION_LIMIT"`
```
recy21, line 170: Even if I agree that pointer be a better choice in the first place. I would not mix the current solution with pointers.

services/notifications.go
```diff
@@ -1526,12 +1526,26 @@ func collectAttestationAndOfflineValidatorNotifications(notificationsByUserID ma
 		}
 	}
 
-	if len(offlineValidators) > 5000 {
-		return fmt.Errorf("retrieved more than 5000 offline validators notifications: %v, exiting", len(offlineValidators))
+	var offlineValidatorsLimit int
```
recy21, line 1529: It's common to init it directly with the default.
So 
```
offlineValidatorsLimit := 5000
if ... { offlineValidatorsLimit = value }
```
is the more convenient way

services/notifications.go
```diff
@@ -1526,12 +1526,26 @@ func collectAttestationAndOfflineValidatorNotifications(notificationsByUserID ma
...
+	}
+
+	var onlineValidatorsLimit int
+	if utils.Config.Notifications.OnlineDetectionLimit != nil {
+		onlineValidatorsLimit = *utils.Config.Notifications.OnlineDetectionLimit
+	} else {
+		onlineValidatorsLimit = 5000
+	}
+
+	if len(offlineValidators) > int(offlineValidatorsLimit) {
```
recy21, line 1543: Why casting an int to an int?

PR #2341, opened 2023-06-21
add MarsProtocol to stakingservice page

Jodragon:
Adding MarsProtocol to the Beacon Chain Explorer [Staking Services Page](https://beaconcha.in/stakingServices).
Name Of Service- MarsProtocol
Website- www.marsprotocol.com

Our users are institutions, so the registration needs to be reviewed, and the fee is also charged in the form of a bill, so if you want to register, please give me your account number?


Jodragon:
@Buttaa Hello, what is the reason for the rejection? How should I apply correctly?

Buttaa:
Hey there! I noticed that the website and product don't seem very reliable, and there are some missing SSL credentials. We don't have a formal application process, but we have the right to decline projects that we feel don't meet our standards.

Jodragon:
@Buttaa It seems that the certificate has expired. Can we resubmit after the fix, or open this again?

Jodragon:
@Buttaa https://www.marsprotocol.com/ Please take a look again, https is fine

PR #2344, opened 2023-06-21
Bids 2218/cache validatorincomehistory

guybrush:
### <samp>🤖 Generated by Copilot at 9c73861</samp>

This pull request improves the performance and scalability of the validator income history API endpoint by adding a cache layer, removing duplicate inputs, and creating missing database indices. It modifies the `db/statistics.go`, `utils/utils.go`, and `db/migrations/20230612084855_add_missing_indices.sql` files.

### DIFF
--- a/db/statistics.go
+++ b/db/statistics.go
@@ -1160,7 +1160,7 @@ func GetValidatorIncomeHistory(validator_indices []uint64, lowerBoundDay uint64,
 	}
 	queryValidatorsStr = utils.UniqueStrings(queryValidatorsStr)
 	sort.Strings(queryValidatorsStr)
-	cacheDur := time.Second * time.Duration(utils.Config.Chain.Config.SecondsPerSlot*utils.Config.Chain.Config.SlotsPerEpoch) // updates every epoch
+	cacheDur := time.Second*time.Duration(utils.Config.Chain.Config.SecondsPerSlot*utils.Config.Chain.Config.SlotsPerEpoch) + 10 // updates every epoch, keep 10sec longer
 	cacheKey := fmt.Sprintf("%d:validatorIncomeHistory:%d:%d:%d:%s", utils.Config.Chain.Config.DepositChainID, lowerBoundDay, upperBoundDay, lastFinalizedEpoch, strings.Join(queryValidatorsStr, ","))
 	if cached, err := cache.TieredCache.GetWithLocalTimeout(cacheKey, cacheDur, []types.ValidatorIncomeHistory{}); err == nil {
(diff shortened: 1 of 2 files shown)


db/statistics.go
```diff
@@ -625,11 +626,28 @@ func GetValidatorIncomeHistoryChart(validator_indices []uint64, currency string,
...
+
+	validatorIndices = utils.SortedUniqueUint64(validatorIndices)
+	validatorIndicesStr := make([]string, len(validatorIndices))
+	for i, v := range validatorIndices {
+		validatorIndicesStr[i] = fmt.Sprintf("%d", v)
+	}
+
+	validatorIndicesPqArr := pq.Array(validatorIndices)
+
+	cacheDur := time.Second * time.Duration(utils.Config.Chain.Config.SecondsPerSlot*utils.Config.Chain.Config.SlotsPerEpoch+10) // updates every epoch, keep 10sec longer
```
recy21, line 646: Should be const

db/statistics.go
```diff
@@ -625,11 +626,28 @@ func GetValidatorIncomeHistoryChart(validator_indices []uint64, currency string,
...
+	if len(validatorIndices) == 0 {
+		return []types.ValidatorIncomeHistory{}, nil
+	}
+
 	if upperBoundDay == 0 {
 		upperBoundDay = 65536
 	}
-	queryValidatorsArr := pq.Array(validator_indices)
+
+	validatorIndices = utils.SortedUniqueUint64(validatorIndices)
```
recy21, line 638: As Golang doesn't support templates (afaik), I would suggest to have a `SortedUniqueUint64ToString` function as well, which do the same thing like this, just convert it to string in the append step.

recy21 (changes requested):

db/statistics.go
guybrush: I am using the sorted uniq `validatorIndices` to create the `validatorIndicesPqArr` aswell, which I think cant handle strings.

db/statistics.go
guybrush: Making this a const is not possible since variable values are involved to calculate it. I could calculate it outside the function, but then need to make sure config has been initialized already. Also I think it is good to have this cacheDuration-calculation close to where its used.

guybrush:
rebased

db/statistics.go
recy21: Okay

db/statistics.go
recy21: Pretty sure strings should be possible here, but Okay, let us keep it for now.

recy21 (approved):

PR #2349, opened 2023-06-26
(BIDS-2176) Reexport historic values

Eisei24:
### <samp>🤖 Generated by Copilot at 2880fa2</samp>

Added a new feature and improved an existing one for handling historical ETH prices. The `misc` tool can now export prices to the database using a command-line argument. The `services` package was refactored to optimize the fetching and writing of prices from a third-party API.


services/historic_prices.go
```diff
@@ -17,6 +17,53 @@ func StartHistoricPriceService() {
 		time.Sleep(time.Hour)
 	}
 }
+
+func WriteHistoricPricesForDay(day int64) error {
```
D13ce, line 21: Why isn't this function also used here: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/2880fa2d8ce2fb31c1dadd7445eb457f96272d01/services/historic_prices.go#L90-L118

The only difference seems to be the addition of the `ON CONFLICT` clause. This could be a parameter though. That way you would reduce duplicated code.

services/historic_prices.go
```diff
@@ -41,12 +88,18 @@ func updateHistoricPrices() error {
...
 				logger.Errorf("error retrieving historic eth prices for day %v: %v", currentDayTrunc, err)
 				currentDay = currentDay.Add(time.Hour * 24)
 				continue
 			}
+
+			if historicPrice.MarketData.CurrentPrice.Usd == 0.0 {
+				logger.Warnf("no historic eth prices for day %v", currentDayTrunc)
+				currentDay = currentDay.Add(time.Hour * 24)
+				continue
+			}
```
D13ce, line 101: While I like the inclusion of this check, I am not sure whether we should only check USD. Why not check all currencies?
By the way, on Prater it looks like many rows include an `AUD` value of 0. This would not be found by this check.

D13ce (changes requested):

services/historic_prices.go
Eisei24: The `ON CONFLICT` clause doesn't really matter, the `updateHistoricPrices` could also use it.
The issue is the update of `currentDay` which is different but it can actually be refactored.

cmd/misc/main.go
```diff
@@ -162,6 +163,17 @@ func main() {
 		IndexOldEth1Blocks(opts.StartBlock, opts.EndBlock, opts.BatchSize, opts.DataConcurrency, opts.Transformers, bt)
 	case "update-aggregation-bits":
 		updateAggreationBits(rpcClient, opts.StartEpoch, opts.EndEpoch, opts.DataConcurrency)
+	case "historic-prices-export":
```
D13ce, line 166: This does not include any logging. Starting the tool, especially now that after each day a delay of 5s has been added, will not give any feedback which might seem odd.

Consider adding stuff like
* Exporting day X
* Finished exporting day X, took Ys

Since this would add more code, I suggest to also move whole block to a separate function (like done for `updateAggreationBits` for example).

D13ce (changes requested):

services/historic_prices.go
```diff
@@ -18,18 +18,24 @@ func StartHistoricPriceService() {
...
+	if historicPrice.MarketData.CurrentPrice.Eth == 0.0 ||
+		historicPrice.MarketData.CurrentPrice.Eur == 0.0 ||
+		historicPrice.MarketData.CurrentPrice.Usd == 0.0 ||
+		historicPrice.MarketData.CurrentPrice.Rub == 0.0 ||
+		historicPrice.MarketData.CurrentPrice.Cny == 0.0 ||
+		historicPrice.MarketData.CurrentPrice.Cad == 0.0 ||
+		historicPrice.MarketData.CurrentPrice.Jpy == 0.0 ||
+		historicPrice.MarketData.CurrentPrice.Gbp == 0.0 ||
+		historicPrice.MarketData.CurrentPrice.Aud == 0.0 {
+		logger.Warnf("no historic eth prices for day %v", ts)
```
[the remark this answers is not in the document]
Eisei24: **1)** Changed
**2)** I think that would be a lot of useless information added. I though about just adding the ones that were 0 to the log but normally if one of them is 0 all of them should be.

cmd/misc/main.go
Eisei24: You currently have logs like 
msg="fetching historic prices for day 2021-03-24 00:00:00 +0000 UTC"
for each day.
I added a general "exporting days X - X" at the start and a "finished export" after each day.

I also changed the "incomplete prices" to an error return so that we stop the export.

D13ce (approved):
lgtm

Please note that I have added two commits on my own. They are very small but ptal before you merge. Cheers!

PR #2353, opened 2023-06-26
(BIDS-2212) get block rewards based on config epoch

MauserBitfly:
### <samp>🤖 Generated by Copilot at 4a910cb</samp>

This pull request updates the code to use the `ChainConfig` type for the Eth1 fork epochs, which are needed to calculate the Eth1 block reward and the Eth2 genesis time. It affects the files `types/chain.go` and `utils/eth1.go`.


D13ce (approved):
lgtm

PR #2355, opened 2023-06-26
Bids 2245/split notification process

peterbitfly:
Split the notification process into two binaries, one responsible for queuing new notifications and one for sending notifications out. Improves marking webhook notifications as sent and harmonizes logging message in the notification module.

### <samp>🤖 Generated by Copilot at a8e1d31</samp>

This pull request refactors the notification service of the explorer into two new binaries: `notification-sender` and `notification-collector`. It also improves the logging, error handling, and database connection management of the notification system. It updates the `Makefile` and the `cmd/explorer/main.go` file to support the new binaries.

### DIFF
--- a/cmd/notification-sender/main.go
+++ b/cmd/notification-sender/main.go
@@ -1,0 +1,162 @@
+package main
+
+import (
+	"eth2-exporter/cache"
+	"eth2-exporter/db"
+	"eth2-exporter/metrics"
+	"eth2-exporter/services"
+	"eth2-exporter/types"
+	"eth2-exporter/utils"
+	"eth2-exporter/version"
+	"flag"
+	"fmt"
+	"net/http"
+	"strings"
+	"sync"
+
+	"github.com/sirupsen/logrus"
+
+	_ "eth2-exporter/docs"
+	_ "net/http/pprof"
+
+	_ "github.com/jackc/pgx/v4/stdlib"
+)
+
+func main() {
+	configPath := flag.String("config", "", "Path to the config file, if empty string defaults will be used")
+
+	flag.Parse()
+
+	cfg := &types.Config{}
+	err := utils.ReadConfig(cfg, *configPath)
+	if err != nil {
+		logrus.Fatalf("error reading config file: %v", err)
+	}
+	utils.Config = cfg
+	logrus.WithFields(logrus.Fields{
+		"config":    *configPath,
+		"version":   version.Version,
+		"chainName": utils.Config.Chain.Config.ConfigName}).Printf("starting")
+
+	if utils.Config.Chain.Config.SlotsPerEpoch == 0 || utils.Config.Chain.Config.SecondsPerSlot == 0 {
+		utils.LogFatal(err, "invalid chain configuration specified, you must specify the slots per epoch, seconds per slot and genesis timestamp in the config file", 0)
+	}
+
+	if utils.Config.Pprof.Enabled {
+		go func() {
+			logrus.Infof("starting pprof http server on port %s", utils.Config.Pprof.Port)
+			logrus.Info(http.ListenAndServe(fmt.Sprintf("0.0.0.0:%s", utils.Config.Pprof.Port), nil))
+		}()
+	}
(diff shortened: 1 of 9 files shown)


cmd/notification-sender/main.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	wg.Wait()
+	if utils.Config.TieredCacheProvider == "bigtable" && len(utils.Config.RedisCacheEndpoint) == 0 {
+		cache.MustInitTieredCacheBigtable(db.BigtableClient.GetClient(), fmt.Sprintf("%d", utils.Config.Chain.Config.DepositChainID))
+		logrus.Infof("tiered Cache initialized, latest finalized epoch: %v", services.LatestFinalizedEpoch())
+	}
+
+	if utils.Config.TieredCacheProvider != "bigtable" && utils.Config.TieredCacheProvider != "redis" {
+		logrus.Fatalf("no cache provider set, pease set TierdCacheProvider (example redis, bigtable)")
```
recy21, line 124: typo in please (missing L)

cmd/notification-sender/main.go
```diff
@@ -0,0 +1,161 @@
...
+			cfg.Frontend.WriterDatabase.Name}
+		frontendDBStr := strings.Join(frontendDBInfo, "-")
+		if DBStr != frontendDBStr {
+			go metrics.MonitorDB(db.FrontendWriterDB)
+		}
+	}
+
+	logrus.Infof("database connection established")
+
+	services.InitNotificationSender()
```
recy21, line 156: 1:1 same function as in notifications-collector, should be handled in a function. The different line can be called after, or via flag / if

services/notifications.go
```diff
@@ -1011,13 +1012,13 @@ func sendWebhookNotifications(useDB *sqlx.DB) error {
...
 
-			if resp != nil && resp.StatusCode < 400 {
-				_, err := useDB.Exec(`UPDATE notification_queue SET sent = now();`)
-				if err != nil {
-					logger.WithError(err).Errorf("error updating notification_queue table")
-					return
-				}
+			_, err = useDB.Exec(`UPDATE notification_queue SET sent = now() where id = $1`, n.Id)
+			if err != nil {
+				logger.WithError(err).Errorf("error updating notification_queue table")
```
recy21, line 1017: Should use `utils.LogError(err,"error updating notification_queue table",0)`

recy21 (changes requested):
As `Config.Notifications.Enabled`is obsolete now, it should be removed from config

services/notifications.go
peterbitfly: we should rework all errors in a separate PR and not in a single location in this pr

cmd/notification-sender/main.go
peterbitfly: don't understand

cmd/notification-sender/main.go
peterbitfly: fixed

cmd/notification-sender/main.go
recy21: `cmd/notification-sender/main.go > func main()`
and 
`cmd/notification-collector/main.go > func main()`
differ only by 1 line.

So I suggest to have only 1 function for this, where the 1 different line is called outside or called via an flag.

types/config.go
```diff
@@ -169,7 +169,6 @@ type Config struct {
 		Pprof   bool   `yaml:"pprof" envconfig:"METRICS_PPROF"`
 	} `yaml:"metrics"`
 	Notifications struct {
-		Enabled                                       bool   `yaml:"enabled" envconfig:"NOTIFICATIONS_ENABLED"`
 		Sender                                        bool   `yaml:"sender" envconfig:"NOTIFICATIONS_SENDER"`
```
recy21, line 172: Please remove Sender too, isn't used anymore

recy21 (changes requested):

recy21 (approved):

PR #2356, opened 2023-06-26
(NOBIDS) handle error more casual and rem bt keys on batch

MauserBitfly:
### <samp>🤖 Generated by Copilot at 988e565</samp>

This pull request enhances the ENS update process by fixing some error handling and performance issues in `db/ens.go`. It avoids unnecessary mutations, checks for write errors, and lowers the error level for name removals.

### DIFF
--- a/db/ens.go
+++ b/db/ens.go
@@ -413,13 +413,13 @@ func (bigtable *Bigtable) ImportEnsUpdates(client *ethclient.Client) error {
 		if err := g.Wait(); err != nil {
 			return err
 		}
+		// After processing a batch keys we remove them from bigtable
 		err = bigtable.WriteBulk(mutsDelete, bigtable.tableData)
 		if err != nil {
 			return err
 		}
 	}
 	logger.Info("ens key indexing completed")
-	// After processing the keys we remove them from bigtable
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


db/ens.go
```diff
@@ -413,10 +413,14 @@ func (bigtable *Bigtable) ImportEnsUpdates(client *ethclient.Client) error {
...
 		}
+		err = bigtable.WriteBulk(mutsDelete, bigtable.tableData)
+		if err != nil {
+			return err
+		}
 	}
 	logger.Info("ens key indexing completed")
 	// After processing the keys we remove them from bigtable
-	return bigtable.WriteBulk(mutsDelete, bigtable.tableData)
+	return err
```
D13ce, line 423: **Minor:** Should return `nil` to show that this path means everything's fine.
Also, the comment above now is misplaced, it should be moved to the `WriteBulk` code again.

db/ens.go
```diff
@@ -541,8 +545,8 @@ func removeEnsName(client *ethclient.Client, name string) error {
 		ens_name = $1
 	;`, name)
 	if err != nil {
-		utils.LogError(err, fmt.Errorf("error deleting ens name [%v]", name), 0)
-		return err
+		logger.Warnf("error deleting ens name [%v]: %v", name, err)
+		return nil
```
D13ce, line 549: I suggest to keep the error and only log a warning for the error in question (i.e. `ERROR: invalid byte sequence for encoding "UTF8": 0x00 (SQLSTATE 22021)` . This would be a less intrusive change.

D13ce (changes requested):

D13ce (approved):
utACK

PR #2357, opened 2023-06-26
(BIDS-2032) Fix slots page url query

Eisei24:
When testing this please also play around with the "Show only empty graffiti" checkbox which should be mutually exclusive with respect to the search text.

### DIFF
--- a/templates/slots.html
+++ b/templates/slots.html
@@ -42,6 +42,9 @@
                 console.error("error loading table state, err:", data, "status: ", status)
               }
               var response = data !== "" ? data : JSON.parse(localStorage.getItem("DataTables_" + settings.sInstance))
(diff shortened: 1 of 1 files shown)


templates/slots.html
```diff
@@ -42,6 +42,9 @@
                 console.error("error loading table state, err:", data, "status: ", status)
               }
               var response = data !== "" ? data : JSON.parse(localStorage.getItem("DataTables_" + settings.sInstance))
+              if (q) {
+                response.search.search = q
+              }
```
Eisei24, line 47: This serves to ensure that use the query from the URL if there is one instead of info from the db or local storage.

LuccaBitfly (approved):
lgtm

PR #2366, opened 2023-06-27
(BIDS-2249) count withdrawals bls for filter

MauserBitfly:
### <samp>🤖 Generated by Copilot at 8aca6af</samp>

This pull request enhances the backend functionality and user interface for displaying withdrawals and BLS changes on the eth2-beaconchain-explorer website. It adds new database functions to count the matching records, optimizes the query execution with concurrency and timeouts, and fixes the data table pagination and filtering. It affects the files `db/db.go` and `handlers/withdrawals.go`.


db/db.go
```diff
@@ -2273,6 +2273,25 @@ func GetTotalWithdrawals() (total uint64, err error) {
 	return
 }
 
+func GetWithdrawalsCountForQuery(query string) (uint64, error) {
+	bquery, _ := hex.DecodeString(strings.TrimPrefix(query, "0x"))
+	count := uint64(0)
+	err := ReaderDb.Get(&count, `
+			SELECT count(*)
```
Eisei24, line 2280: **issue:** I see this is essentially the same query as in GetWithdrawals but that query doesn't really work which is something that will be fixed in BIDS-2184. Whichever one gets merged first, the other one has to fix this too.

db/db.go
```diff
@@ -2888,6 +2907,33 @@ func GetTotalBLSChanges() (uint64, error) {
 	return count, nil
 }
 
+func GetBLSChangesCountForQuery(query string) (uint64, error) {
+	bquery, _ := hex.DecodeString(strings.TrimPrefix(query, "0x"))
+	count := uint64(0)
+	err := ReaderDb.Get(&count, `
+		SELECT count(*)
```
Eisei24, line 2914: **issue:** Same issue as with the withdrawals.

Eisei24 (changes requested):
This fix was also required for the epochs page, see the edit in the BIDS' description.

db/db.go
MauserBitfly: applied the new logic also for counting

db/db.go
MauserBitfly: applied the new logic also for counting

Eisei24 (approved):
lgtm

db/db.go
```diff
@@ -2892,6 +2926,53 @@ func GetTotalBLSChanges() (uint64, error) {
...
+		) AS val ON val.validatorindex = bls.validatorindex`
+
+	// Check whether the query can be used for a validator, slot or epoch search
+	if uiQuery, parseErr := strconv.ParseUint(query, 10, 64); parseErr == nil {
+		searchQuery := `
+			WHERE bls.validatorindex = $1			
+				OR block_slot = $1
+				OR (block_slot / $3) = $1
+				OR pubkey_text LIKE ($2 || '%')
+				OR ENCODE(val.deposit_adress, 'hex') LIKE ($2 || '%')`
```
recy21, line 2958: That's not how we spell address in English, even if you just copied it from GetBLSChanges, I would correct it.

db/db.go
```diff
@@ -2892,6 +2926,53 @@ func GetTotalBLSChanges() (uint64, error) {
...
+
+		err = ReaderDb.Get(&count, fmt.Sprintf(blsQuery, joinQuery, searchQuery),
+			uiQuery, trimmedQuery, utils.Config.Chain.Config.SlotsPerEpoch)
+	} else if blsLikeRE.MatchString(query) {
+		searchQuery := `
+				WHERE pubkey_text LIKE ($1 || '%')
+				OR ENCODE(val.deposit_adress, 'hex') LIKE ($1 || '%')`
```
recy21, line 2965: Even there is an index for eth1_deposits.from_address, I don't think it can do the index thing, when used with ENCODE, so a full-table-scan will be very likely. I would suggest to create another field, which contain already the encoded value and use an index there.

recy21 (changes requested):

db/db.go
MauserBitfly: renamed all occurrences of deposit_adress

db/db.go
MauserBitfly: added a from_address_text column to the eth1_deposits table

db/migrations/tbd_add_from_address_to_text_eth1_deposits.sql
```diff
@@ -0,0 +1,42 @@
...
+SELECT 'populate new eth1_deposits column from_address_text, 1000 at a time'; 
+DO
+$do$
+DECLARE
+    count INTEGER default 0;
+BEGIN
+LOOP
+    count := (count + 1);
+    RAISE NOTICE 'count = %', count;
+    EXIT WHEN (SELECT count(*) from (SELECT from_address_text FROM eth1_deposits WHERE from_address_text = '' LIMIT(1)) as sub) = 0;
```
recy21, line 18: FROM and AS please uppercase

exporter/eth1.go
```diff
@@ -267,13 +268,14 @@ func saveEth1Deposits(depositsToSave []*types.Eth1Deposit) error {
 			removed,
 			valid_signature
 		)
-		VALUES ($1, $2, $3, $4, TO_TIMESTAMP($5), $6, $7, $8, $9, $10, $11, $12, $13)
+		VALUES ($1, $2, $3, $4, TO_TIMESTAMP($5), $6, ENCODE($7.from_address, 'hex'), $8, $9, $10, $11, $12, $13, %14)
```
recy21, line 271: %14 looks wrong, shouldn't it be $14 ?

handlers/search.go
```diff
@@ -417,51 +417,34 @@ func FindValidatorIndicesByEth1Address(search string) (types.SearchValidatorsByE
...
-					ENCODE(from_address::bytea, 'hex') as from_address,
-					DENSE_RANK() OVER (PARTITION BY from_address ORDER BY validatorindex) AS validatorrow,
-					DENSE_RANK() OVER (ORDER BY from_address) AS addressrow
+					from_address_text,
+					DENSE_RANK() OVER (PARTITION BY from_address_text ORDER BY validatorindex) AS validatorrow,
+					DENSE_RANK() OVER (ORDER BY from_address_text) AS addressrow
 				FROM eth1_deposits
 				INNER JOIN validators ON validators.pubkey = eth1_deposits.publickey
-				WHERE from_address LIKE $1 || '%'::bytea
+				WHERE from_address_text = $1
```
recy21, line 428: You removed the %, don't get why?
But you added it in line 443.
Is this cause by the rework?

exporter/eth1.go
MauserBitfly: Good catch, yes $14 is correct

db/migrations/tbd_add_from_address_to_text_eth1_deposits.sql
MauserBitfly: is now upper case

handlers/search.go
MauserBitfly: No its the git highlighting that is misleading.
the `from_address_text = $1` is from the first if statement where we do an exact match search if the eth1AddressHash has the exact length. 

I remove the second if statement as it is no longer needed with the switch to the new from_address_text column.

And in the else there is the wildcard search.

recy21 (approved):

peterbitfly (approved):

PR #2368, opened 2023-06-27
(BIDS-1414) let's not start with a full chart day

MauserBitfly:
### <samp>🤖 Generated by Copilot at b880fa1</samp>

Fixed a bug in the chart series data calculation for the first and last slot of a day, and added a helper function to find the first slot of an epoch given a timestamp. The changes affect the files `db/statistics.go` and `utils/utils.go`.


Eisei24 (approved):
lgtm

PR #2372, opened 2023-06-28
(BIDS-2252) remove errors from search input

LuccaBitfly:
### <samp>🤖 Generated by Copilot at 345b7d1</samp>

Fix overflow bugs in slot and validator search functions. Use `strconv.ParseInt` instead of `strconv.Atoi` in `handlers/search.go`.

### DIFF
--- a/handlers/search.go
+++ b/handlers/search.go
@@ -81,7 +81,7 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 		}
 		result = &types.SearchAheadSlotsResult{}
 		if searchLikeRE.MatchString(search) {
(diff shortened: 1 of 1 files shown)


D13ce (changes requested):
While this cannot be triggered with the central search bar, searching for `59408965940896` in the dashboard triggers the same error here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/345b7d17a23d71e280feb5510da7925973c53d9f/handlers/search.go#L228

I would argue that it is in the scope of this PR to fix that too.

D13ce (approved):
lgtm

PR #2374, opened 2023-06-28
(BIDS-2178) lets not finalize too early

MauserBitfly:
### <samp>🤖 Generated by Copilot at b8c65ab</samp>

This pull request fixes two bugs related to the display and calculation of finalized epochs and slots in the slot page and the Lighthouse RPC client. It adjusts the join condition in `handlers/slot.go` and the epoch parsing in `rpc/lighthouse.go` to account for the different ways that finalized checkpoints are represented by the database and the Lighthouse API.

### DIFF
--- a/handlers/slot.go
+++ b/handlers/slot.go
@@ -241,7 +241,7 @@ func GetSlotPageData(blockSlot uint64) (*types.BlockPageData, error) {
 		LEFT JOIN validator_names ON validators.pubkey = validator_names.publickey
 		LEFT JOIN blocks_tags ON blocks.slot = blocks_tags.slot and blocks.blockroot = blocks_tags.blockroot
 		LEFT JOIN tags ON blocks_tags.tag_id = tags.id
-		LEFT JOIN epochs ON (GREATEST(blocks.slot,1)-1)/32 = epochs.epoch
+		LEFT JOIN epochs ON (GREATEST(blocks.slot,1)-1)/$2 = epochs.epoch
 		WHERE blocks.slot = $1 
 		group by
 			blocks.epoch,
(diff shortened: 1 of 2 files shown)


handlers/slot.go
```diff
@@ -240,7 +241,7 @@ func GetSlotPageData(blockSlot uint64) (*types.BlockPageData, error) {
 		LEFT JOIN validator_names ON validators.pubkey = validator_names.publickey
 		LEFT JOIN blocks_tags ON blocks.slot = blocks_tags.slot and blocks.blockroot = blocks_tags.blockroot
 		LEFT JOIN tags ON blocks_tags.tag_id = tags.id
-		LEFT JOIN epochs ON blocks.epoch = epochs.epoch
+		LEFT JOIN epochs ON (GREATEST(blocks.slot,1)-1)/32 = epochs.epoch
```
LuccaBitfly, line 244: The epoch in this query is also used for the participation rate. If an epoch is justified, but the following one isn't, the first slot of the latter epoch would be shown as "Not Finalized (Safe)" i.e. justified with this change, which is wrong.

You could alternatively use `services.LatestFinalizedEpoch()` to determine a slot's finality.

rpc/lighthouse.go
```diff
@@ -149,12 +149,18 @@ func (lc *LighthouseClient) GetChainHead() (*types.ChainHead, error) {
 		return nil, fmt.Errorf("error parsing finality checkpoints of head: %v", err)
 	}
 
+	// The epoch in the Finalized Object is not the finalized epoch, but the epoch for the checkpoint - the 'real' finalized epoch is the one before
```
LuccaBitfly, line 152: Imo this comment is misleading. I would argue that 'real' finalization means that only an epoch's first slot is finalized. We mark an epoch finalized if all of its slots are finalized instead, which is not the 'real' definition of finalization but a more practical one.

LuccaBitfly (changes requested):
Please have a look at my issues.

LuccaBitfly (approved):
lgtm

peterbitfly (changes requested):

handlers/slot.go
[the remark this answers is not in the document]
MauserBitfly: Yes of course - changed it

peterbitfly:
LGTM

peterbitfly (approved):

PR #2375, opened 2023-06-28
Bids 2250/improve last attestation slot handling 2

peterbitfly:
Move storing and retrieving the last attestation slot of a validator to bigtable. Data is still written to postgres in parallel to support rolling this PR back in case of any issues. Disabled sorting the validators table after the last attestation slot.

Requires a cleanup pr to remove exporting the last attestation slot to postgres and delete the lastattestation slot column in the validators table.

Removes the leveldb based LastAttestationCache, obsolescence the LastAttestationCachePath config flag.

### <samp>🤖 Generated by Copilot at 01a6b55</samp>

This pull request implements a new feature that uses an in-memory cache and Bigtable storage to store and retrieve the last attestation slot data for validators. It removes the dependency on the Postgres database and the file-based cache for this data, and updates the relevant packages and templates to use the new system. It improves the performance, scalability, and reliability of the attestation data queries and displays.


db/bigtable.go
```diff
@@ -844,6 +885,63 @@ func (bigtable *Bigtable) GetValidatorAttestationHistory(validators []uint64, st
...
+	)
+
+	if len(columnFilters) == 1 { // special case to retrieve data for one validators
+		filter = gcp_bigtable.ChainFilters(
+			gcp_bigtable.FamilyFilter(ATTESTATIONS_FAMILY),
+			columnFilters[0],
+			gcp_bigtable.LatestNFilter(1),
+		)
+	}
+	if len(columnFilters) == 0 { // special case to retrieve data for all validators
```
D13ce, line 917: **nitpick:** Should be `else if`.

D13ce (changes requested):
Changes look good, I have added a few comments though.

Furthermore: `.gitignore` contains a line for the now deleted `lastAttestationCache.go` which should be removed: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/1259ed79e75450071e6e1a49c8fdd099dd3d271f/.gitignore#L28

peterbitfly:
> Changes look good, I have added a few comments though.
> 
> Furthermore: `.gitignore` contains a line for the now deleted `lastAttestationCache.go` which should be removed:
> 
> https://github.com/gobitfly/eth2-beaconchain-explorer/blob/1259ed79e75450071e6e1a49c8fdd099dd3d271f/.gitignore#L28

I'd suggest keeping the entry in the `.gitignore` as there might be an old directory in local folders that will then get committed.

types/templates.go
```diff
@@ -346,11 +346,11 @@ type ValidatorPageData struct {
 	SlashedBy                                uint64
 	SlashedAt                                uint64
 	SlashedFor                               string
-	ActivationEligibilityEpoch               uint64         `db:"activationeligibilityepoch"`
-	ActivationEpoch                          uint64         `db:"activationepoch"`
-	ExitEpoch                                uint64         `db:"exitepoch"`
-	Index                                    uint64         `db:"index"`
-	LastAttestationSlot                      uint64         `db:"lastattestationslot"`
```
D13ce: I did not suggest to remove the `db` tag here. Now that you did, I checked again why I did not suggest it and I realized that I forgot to mention this in my previous review:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/4c13158c1df89dc46db427cdac0117c1ac58d446/services/rewards.go#L297-L301
This is still using `lastattestationslot` coming from Postgres. I guess this should be adapted so it takes the info from bigtable too? If so, removing the `db` tag from `LastAttestationSlot` in `ValidatorPageData` in fine.

D13ce (changes requested):
Your changes lgtm. Still, one thing came up I forgot to mention before. Sorry for that.

D13ce (approved):
utACK :+1:

Bumdue2233 (commented):
All assets to be sent to Samuel Boyers coin base wallets.

- [](url)https://www.coinbase.com/portfolio/f6a4964d-f7a3-44d5-80b1-863a525b9684?utm_campaign=rt_i_m_w_m_acq_ugc_soc_0_portfolio&utm_source=ugc&utm_platform=Android

PR #2381, opened 2023-06-29
(BIDS-2115) import previously missed block

MauserBitfly:
### <samp>🤖 Generated by Copilot at a8057af</samp>

This pull request enhances the block processing logic in `cmd/misc/main.go` by adding a block status check and simplifying the attestation queries.

### DIFF
--- a/cmd/misc/main.go
+++ b/cmd/misc/main.go
@@ -221,6 +221,32 @@ func updateAggreationBits(rpcClient *rpc.LighthouseClient, startEpoch uint64, en
 					})
 					continue
 				}
+
+				status := uint64(0)
+				err := db.ReaderDb.Get(&status, `
+				SELECT status
+				FROM blocks WHERE 
+					slot=$1`, block.Slot)
+				if err != nil {
+					utils.LogError(err, fmt.Errorf("error getting Slot [%v] status", block.Slot), 0)
+					return
+				}
+
+				if status != block.Status {
+					logrus.Infof("Slot[%v] has the wrong status [%v], but should be [%v]", block.Slot, status, block.Status)
+					if block.Status == 1 {
+						err := db.SaveBlock(block)
+						if err != nil {
+							utils.LogError(err, fmt.Errorf("error saving Slot [%v]", block.Slot), 0)
+							return
+						}
+						continue
+					} else {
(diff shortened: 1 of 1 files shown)


cmd/misc/main.go
```diff
@@ -221,6 +221,32 @@ func updateAggreationBits(rpcClient *rpc.LighthouseClient, startEpoch uint64, en
...
+
+				if status != block.Status {
+					logrus.Infof("Slot[%v] has the wrong status [%v], but should be [%v]", block.Slot, status, block.Status)
+					if block.Status == 1 {
+						err := db.SaveBlock(block)
+						if err != nil {
+							utils.LogError(err, fmt.Errorf("error saving Slot [%v]", block.Slot), 0)
+							return
+						}
+						continue
```
Eisei24, line 243: **question:** Why do we continue here to the next block after updating the slot?

Eisei24 (commented):
Please have a look at my question.

cmd/misc/main.go
MauserBitfly: Because all the relevant information is set in the db.SaveBlock(block), so we don't have to check the aggregation bits explicitly for this one.

Eisei24: Are you sure this works?
db.SaveBlock(block) does save the aggregation bits here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/11ccd3cd738899a84b800a5009ec7d6849df3a19/db/db.go#L1515
but if the entry already exists we have an `ON CONFLICT ... DO NOTHING`

MauserBitfly: But we only get here for blocks with status 0 in our db - which means we have not yet saved the block data for this block.   So there should not be any conflicts.

Katarzynamm:
❤️❤️❤️

Wysłano z programu Outlook dla systemu Android<https://aka.ms/AAb9ysg>
________________________________

cmd/misc/main.go
Eisei24: I see. Thank you for the explanation.

Eisei24 (approved):
lgtm

peterbitfly (approved):

PR #2382, opened 2023-06-29
(BIDS-2274) Fix the status of the validators in the dashboard

Eisei24:
### <samp>🤖 Generated by Copilot at 4f253a9</samp>

Fix validator state counts on dashboard by using state name instead of array. Update `static/js/dashboard.js` to access the correct element of `v[3]`.


D13ce (approved):
lgtm

PR #2384, opened 2023-07-02
(BIDS-2031) add sanity check to lighthouse module to restart on error…

peterbitfly:
… with event source block stream

### <samp>🤖 Generated by Copilot at b3a0ae8</samp>

Replace SSE stream with polling loop and add health check for Lighthouse API in `rpc/lighthouse.go`. This improves data reliability and accuracy for the explorer.

### DIFF
--- a/rpc/lighthouse.go
+++ b/rpc/lighthouse.go
@@ -60,6 +60,7 @@ func (lc *LighthouseClient) GetNewBlockChan() chan *types.Block {
 		for ; ; time.Sleep(time.Second) {
 			lc.lastBlockSeenMux.Lock()
 			if time.Since(lc.lastBlockSeen) > time.Minute*2 {
+				lc.lastBlockSeenMux.Unlock()
 				// fatal if no new block was received for more than two minutes
 				logger.Fatalf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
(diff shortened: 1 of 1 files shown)


rpc/lighthouse.go
```diff
@@ -42,13 +44,29 @@ func NewLighthouseClient(endpoint string, chainID *big.Int) (*LighthouseClient,
...
+	// setup health check & exit if the new block chan does no longer receive new blocks
+	lc.lastBlockSeenMux.Lock()
+	lc.lastBlockSeen = time.Now()
+	lc.lastBlockSeenMux.Unlock()
+	go func() {
+		for ; ; time.Sleep(time.Second) {
+			lc.lastBlockSeenMux.Lock()
+			if time.Since(lc.lastBlockSeen) > time.Minute*2 {
+				// fatal if no new block was received for more than two minutes
+				logger.Fatalf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
```
recy21, line 64: You should unlock the mutex before the fatal.
`lc.lastBlockSeenMux.Unlock()`
As it's not guaranteed that the program will end by the fatal, it can be caught (in the future) and the we end in an deadlock.

recy21 (changes requested):

recy21 (approved):

PR #2393, opened 2023-07-10
(BIDS-2292) Fix data indexes used in dashboard.js

D13ce:
This PR adds comments in dashboard.js regarding the data indexes and adds a block for the income (that would not be required if the right indexes would have been used but now everything is cleaner and more obvious).

### DIFF
--- a/static/js/dashboard.js
+++ b/static/js/dashboard.js
@@ -518,6 +518,7 @@ $(document).ready(function () {
     },
     order: [[1, "asc"]],
     columnDefs: [
+      // Pubkey
       {
         targets: 0,
         data: "0",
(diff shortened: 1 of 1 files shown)


static/js/dashboard.js
```diff
@@ -618,14 +625,22 @@ $(document).ready(function () {
...
-        // hidden column for filtering by DepositAddress
         targets: 9,
+        data: "9",
+        render: function (data, type, row, meta) {
+          return data
+        },
+      },
+      // Deposit address
+      {
+        targets: 10,
```
D13ce, line 638: `data: "9"` stands for the income and there was no code block for that before as the content for data[9] could be used as it is. Adding the block ranging from line 628 to 635 does not really do anything besides making everything more obvious.

LuccaBitfly (approved):
lgtm

static/js/dashboard.js
LuccaBitfly: paprika

PR #2397, opened 2023-07-12
(BIDS-2294) Only check for proposed blocks

D13ce:
This PR causes the `/execution/block/` API endpoint to only check for proposed blocks in `func findExecBlockNumbersByExecBlockNumber`.


Eisei24 (approved):
lgtm, good find.

PR #2401, opened 2023-07-14
(BIDS-2279) Add head parameter to slot endpoint

D13ce:
This PR adds the possibility to use "head" as parameter for the API V1 slot endpoint. When used, it will return the slot containing the chain's head block.
Swagger has been updated accordingly.

### DIFF
--- a/handlers/api.go
+++ b/handlers/api.go
@@ -372,11 +372,11 @@ func ApiEpochSlots(w http.ResponseWriter, r *http.Request) {
 }
 
 // ApiSlots godoc
-// @Summary Get a slot by its slot number or root hash
+// @Summary Get a slot by its slot number or root hash. Alternatively get the latest slot or the slot containing the head block.
 // @Tags Slot
-// @Description Returns a slot by its slot number or root hash or the latest slot with string latest
+// @Description Returns a slot by its slot number or root hash, the latest slot with string latest or the slot containing the head block with string head
 // @Produce  json
-// @Param  slotOrHash path string true "Slot or root hash or the string latest"
+// @Param  slotOrHash path string true "Slot or root hash or the string latest or head"
 // @Success 200 {object} types.ApiResponse{data=types.APISlotResponse}
 // @Failure 400 {object} types.ApiResponse
 // @Router /api/v1/slot/{slotOrHash} [get]
@@ -384,25 +384,42 @@ func ApiSlots(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 
 	vars := mux.Vars(r)
-
 	slotOrHash := strings.Replace(vars["slotOrHash"], "0x", "", -1)
+
 	blockSlot := int64(-1)
-	blockRootHash, err := hex.DecodeString(slotOrHash)
-	if slotOrHash != "latest" && (err != nil || len(slotOrHash) != 64) {
-		blockRootHash = []byte{}
-		blockSlot, err = strconv.ParseInt(vars["slotOrHash"], 10, 64)
-		if err != nil {
-			sendErrorResponse(w, r.URL.String(), "could not parse slot number")
+	blockRootHash := []byte{}
+
+	if slotOrHash == "latest" {
+		// simply check the latest slot (might be empty which causes an error)
+		blockSlot = int64(services.LatestSlot())
+	} else if slotOrHash == "head" {
+		// retrieve the slot containing the head block of the chain (cannot return an empty slot unless there is no block in the chain yet)
+		err := db.ReaderDb.Get(&blockRootHash, `
+			SELECT blockroot
+			FROM blocks
+			WHERE status = '1'
+			ORDER BY slot DESC
+			LIMIT 1`)
+
+		if err != nil || len(blockRootHash) != 32 {
+			sendErrorResponse(w, r.URL.String(), "could not retrieve db results")
 			return
+		}
+	} else {
+		var err error
+		blockRootHash, err = hex.DecodeString(slotOrHash)
+		if err != nil || len(slotOrHash) != 64 {
+			// not a valid root hash, try to parse as slot number instead
+			blockRootHash = []byte{}
+			blockSlot, err = strconv.ParseInt(vars["slotOrHash"], 10, 64)
+			if err != nil {
+				sendErrorResponse(w, r.URL.String(), "could not parse slot number")
+				return
+			}
 		}
 	}
 
-	if slotOrHash == "latest" {
-		blockSlot = int64(services.LatestSlot())
-	}
-
 	if len(blockRootHash) != 32 {
-		// blockRootHash is required for the SQL statement below, if none has passed we retrieve it manually
 		err := db.ReaderDb.Get(&blockRootHash, `SELECT blockroot FROM blocks WHERE slot = $1`, blockSlot)
 
 		if err != nil || len(blockRootHash) != 32 {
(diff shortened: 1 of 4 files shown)


Eisei24 (approved):
lgtm

I could not reproduce the bug before or after the changes.

handlers/api.go
```diff
@@ -372,37 +372,54 @@ func ApiEpochSlots(w http.ResponseWriter, r *http.Request) {
...
-		}
-	}
+	blockRootHash := []byte{}
 
 	if slotOrHash == "latest" {
+		// simply check the latest slot (might be empty which causes an error)
 		blockSlot = int64(services.LatestSlot())
+	} else if slotOrHash == "head" {
+		// retrieve the slot containing the head block of the chain (cannot return an empty slot unless there is no block in the chain yet)
+		err := db.ReaderDb.Get(&blockRootHash, `
```
peterbitfly, line 397: this should not hit the database, instead create a new property in the services module that gets updated via the FDU

peterbitfly (changes requested):

Eisei24 (approved):
lgtm

services/execution_layer.go
```diff
@@ -34,14 +36,60 @@ func latestBlockUpdater(wg *sync.WaitGroup) {
...
+		blockRootHash := []byte{}
+		err := db.ReaderDb.Get(&blockRootHash, `
+		SELECT blockroot
+		FROM blocks
+		WHERE status = '1'
+		ORDER BY slot DESC
+		LIMIT 1`)
+
+		if err != nil {
+			logger.WithError(err).Error("error getting blockrroot hash for chain head")
```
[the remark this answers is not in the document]
D13ce: Fixed.

services/execution_layer.go
```diff
@@ -34,14 +36,60 @@ func latestBlockUpdater(wg *sync.WaitGroup) {
...
+// headBlockRootHashUpdater updates the hash of the current chain head block
+func headBlockRootHashUpdater(wg *sync.WaitGroup) {
+	firstRun := true
+
+	for {
+		blockRootHash := []byte{}
+		err := db.ReaderDb.Get(&blockRootHash, `
+		SELECT blockroot
+		FROM blocks
+		WHERE status = '1'
```
recy21, line 62: There is no index for just `status` we should use one, or proof that a "multi-index" will cover this case.

services/execution_layer.go
```diff
@@ -34,14 +36,60 @@ func latestBlockUpdater(wg *sync.WaitGroup) {
...
+		ORDER BY slot DESC
```
recy21, line 63: There is no index for just `slot` we should use one, or proof that a "multi-index" will cover this case.

services/execution_layer.go
```diff
@@ -34,14 +36,60 @@ func latestBlockUpdater(wg *sync.WaitGroup) {
...
+		LIMIT 1`)
+
+		if err != nil {
+			logger.WithError(err).Error("error getting blockroot hash for chain head")
+		}
+		cacheKey := fmt.Sprintf("%d:frontend:%s", utils.Config.Chain.Config.DepositChainID, latestBlockHashRootCacheKey)
+		err = cache.TieredCache.SetString(cacheKey, string(blockRootHash), time.Hour*24)
+		if err != nil {
+			logger.Errorf("error caching %s: %v", latestBlockHashRootCacheKey, err)
```
recy21, line 72: Should use utils.LogError

services/execution_layer.go
```diff
@@ -34,14 +36,60 @@ func latestBlockUpdater(wg *sync.WaitGroup) {
...
+		blockRootHash := []byte{}
+		err := db.ReaderDb.Get(&blockRootHash, `
+		SELECT blockroot
+		FROM blocks
+		WHERE status = '1'
+		ORDER BY slot DESC
+		LIMIT 1`)
+
+		if err != nil {
+			logger.WithError(err).Error("error getting blockroot hash for chain head")
```
recy21, line 67: Should use utils.LogError

services/execution_layer.go
```diff
@@ -18,10 +20,10 @@ func latestBlockUpdater(wg *sync.WaitGroup) {
 		if err != nil {
 			logger.WithError(err).Error("error getting most recent eth1 block")
 		}
-		cacheKey := fmt.Sprintf("%d:frontend:latestEth1BlockNumber", utils.Config.Chain.Config.DepositChainID)
+		cacheKey := fmt.Sprintf("%d:frontend:%s", utils.Config.Chain.Config.DepositChainID, latestBlockNumberCacheKey)
 		err = cache.TieredCache.SetUint64(cacheKey, recent.GetNumber(), time.Hour*24)
 		if err != nil {
-			logger.Errorf("error caching latestBlockNumber: %v", err)
+			logger.Errorf("error caching %s: %v", latestBlockNumberCacheKey, err)
```
recy21, line 26: Should use utils.LogError

recy21 (changes requested):

services/execution_layer.go
D13ce: Fixed

services/execution_layer.go
```diff
@@ -34,14 +36,60 @@ func latestBlockUpdater(wg *sync.WaitGroup) {
...
+// LatestEth1BlockNumber will return most recent eth1 block number
 func LatestEth1BlockNumber() uint64 {
-	cacheKey := fmt.Sprintf("%d:frontend:latestEth1BlockNumber", utils.Config.Chain.Config.DepositChainID)
+	cacheKey := fmt.Sprintf("%d:frontend:%s", utils.Config.Chain.Config.DepositChainID, latestBlockNumberCacheKey)
 
 	if wanted, err := cache.TieredCache.GetUint64WithLocalTimeout(cacheKey, time.Second*5); err == nil {
 		return wanted
 	} else {
-		logger.Errorf("error retrieving latestEth1BlockNumber from cache: %v", err)
+		logger.Errorf("error retrieving %s from cache: %v", latestBlockNumberCacheKey, err)
```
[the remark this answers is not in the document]
D13ce: Fixed

services/execution_layer.go
```diff
@@ -9,6 +9,8 @@ import (
 	"time"
 )
 
+const latestBlockNumberCacheKey = "latestEth1BlockNumber"
```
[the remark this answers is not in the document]
D13ce: Moved

handlers/api.go
D13ce: Added a new service accordingly

services/execution_layer.go
D13ce: Here is your proof that the composite index `idx_blocks_status_slot_epoch` is used:

```
EXPLAIN ANALYZE
	SELECT blockroot
	FROM blocks
	WHERE status = '1'
	ORDER BY slot DESC
	LIMIT 1;
```
=>
```
... // some output
   ->  Index Scan Backward using idx_blocks_status_slot_epoch on blocks 
... // some output
```
This works as `idx_blocks_status_slot_epoch` has `status` as its first part, see here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/44141630c110fa8284724f18d5e6970fb92ab819/db/migrations/20230612084855_add_missing_indices.sql#L49

services/execution_layer.go
D13ce: Similar to my answer here (https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2401#discussion_r1291084623), the already existing index `idx_blocks_status_slot_epoch` covers this as its second part is `slot`.
So this query ONLY uses `idx_blocks_status_slot_epoch` for both status and slot which works as the composite index has both columns as first and second part.

services/execution_layer.go
D13ce: Fixed

services/execution_layer.go
D13ce: Fixed

services/execution_layer.go
recy21: good

services/execution_layer.go
recy21: good

recy21 (approved):

peterbitfly (approved):

PR #2405, opened 2023-07-17
(BIDS-2250) remove last attestation slot from the exporter process

peterbitfly:
this is a cleanup PR that removes writing the last attestation slot to the postgres db and will significantly improve the epoch export speed

includes a one time migration script that writes the last attestation slot that is currently in postgres to bigtable. this can safely be done for active validators as bigtable will only keep the most recent last attestation slot. the misc tool needs to be run with the `migrate-last-attestation-slot-bigtable` command.

### <samp>🤖 Generated by Copilot at 5189fe6</samp>

This pull request improves the efficiency and reliability of the validator attestation queries by using Bigtable as a cache layer. It refactors the code in `./db/db.go` to replace the PostgreSQL queries with Bigtable calls.

### DIFF
--- a/cmd/misc/main.go
+++ b/cmd/misc/main.go
@@ -166,8 +166,33 @@ func main() {
 		updateAggreationBits(rpcClient, opts.StartEpoch, opts.EndEpoch, opts.DataConcurrency)
 	case "historic-prices-export":
 		exportHistoricPrices(opts.StartDay, opts.EndDay)
+	case "migrate-last-attestation-slot-bigtable":
+		migrateLastAttestationSlotToBigtable()
 	default:
 		utils.LogFatal(nil, "unknown command", 0)
+	}
+}
+
+// one time migration of the last attestation slot values from postgres to bigtable
+// will write the last attestation slot that is currently in postgres to bigtable
+// this can safely be done for active validators as bigtable will only keep the most recent
+// last attestation slot
+func migrateLastAttestationSlotToBigtable() {
+	validators := []types.Validator{}
+
+	err := db.WriterDb.Select(&validators, "SELECT validatorindex, lastattestationslot FROM validators WHERE lastattestationslot IS NOT NULL ORDER BY validatorindex")
+
+	if err != nil {
+		utils.LogFatal(err, "error retrieving last attestation slot", 0)
+	}
+
+	for _, validator := range validators {
+		logrus.Infof("setting last attestation slot %v for validator %v", validator.Index, validator.LastAttestationSlot)
+
+		err := db.BigtableClient.SetLastAttestationSlot(validator.Index, uint64(validator.LastAttestationSlot.Int64))
+		if err != nil {
+			utils.LogFatal(err, "error setting last attestation slot", 0)
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


cmd/misc/main.go
```diff
@@ -166,11 +166,36 @@ func main() {
...
+	validators := []types.Validator{}
+
+	err := db.WriterDb.Select(&validators, "SELECT validatorindex, lastattestationslot FROM validators WHERE lastattestationslot IS NOT NULL ORDER BY validatorindex")
+
+	if err != nil {
+		utils.LogFatal(err, "error retrieving last attestation slot", 0)
+	}
+
+	for _, validator := range validators {
+		logrus.Infof("setting last attestation slot %v for validator %v", validator.Index, validator.LastAttestationSlot)
```
D13ce, line 190: Wrong order of the parameters.

db/db.go
D13ce, line 1: Now that `lastattestationslot` is not used anymore, there should be a migration file that deletes the column from the according table. Said file cannot be part of this PR of course (as it is required for the migration script), but we should not forget to add it at some point.

db/db.go
```diff
@@ -1053,18 +1053,23 @@ func saveValidators(data *types.EpochData, tx *sqlx.Tx, client rpc.Client) error
 	}
 
 	var currentState []*types.Validator
-	err := tx.Select(&currentState, "SELECT validatorindex, withdrawableepoch, withdrawalcredentials, slashed, activationeligibilityepoch, activationepoch, exitepoch, lastattestationslot, status FROM validators;")
+	err := tx.Select(&currentState, "SELECT validatorindex, withdrawableepoch, withdrawalcredentials, slashed, activationeligibilityepoch, activationepoch, exitepoch, status FROM validators;")
```
D13ce, line 1056: Now that `lastattestationslot` is not part of the table anymore, consider removing the `db` tag and changing the datatype here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/36a95eb96c69036c42bc6891d98cd5d177b592af/types/exporter.go#L87

D13ce (changes requested):

db/bigtable.go
```diff
@@ -612,6 +612,20 @@ func (bigtable *Bigtable) SaveAttestations(blocks map[uint64]map[string]*types.B
...
+	mutLastAttestationSlot := gcp_bigtable.NewMutation()
+	mutLastAttestationSlot.Set(ATTESTATIONS_FAMILY, fmt.Sprintf("%d", validator), gcp_bigtable.Timestamp(lastAttestationSlot*1000), []byte{})
+	err := bigtable.tableValidators.Apply(ctx, fmt.Sprintf("%s:lastAttestationSlot", bigtable.chainId), mutLastAttestationSlot)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
```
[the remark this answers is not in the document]
peterbitfly: currently `tableValidators` is a private field of the `bigtable` struct. as it is now, its not possible to access that field outside of the db package. hence the function. I've added a comment that this method is not for production use.

db/db.go
peterbitfly: that needs to be done in a separate bids, as the migration script still uses that mapping

db/db.go
peterbitfly: ok, should be done in a separate bids

cmd/misc/main.go
peterbitfly: fixed

db/db.go
D13ce: Right :facepalm:

PR #2409, opened 2023-07-18
(BIDS-2228) limit vali page income to 30 days

LuccaBitfly:
### <samp>🤖 Generated by Copilot at 89ac788</samp>

This pull request improves the performance, accuracy, and user interface of the income calculation and display for validators. It adds a day filter and a status filter to the execution income query, uses a cached value for the last exported statistic day, adds a background service to update the statistics, and adds a button to load the full income history data. It also fixes some minor issues with the error handling, variable scope, and layout of the dashboard and validator pages. It modifies several files in the `handlers`, `services`, `db`, `static`, and `templates` directories, and deletes the unused `pools_updater.go` file.


templates/validator/charts.html
```diff
@@ -154,111 +154,149 @@
...
+                alert('Error loading data. Please try again.');
+                incomeChartLoaded = false
+              });
+          });
+        })
+      </script>
     {{ end }}
   {{ end }}
 {{ end }}
+001058219
```
D13ce, line 302: I guess this should be removed.

templates/validator/charts.html
```diff
@@ -154,111 +154,149 @@
...
+          $("#load-income-btn").on("click", () => {
+            if (incomeChartLoaded) {
+              return
+            }
+            incomeChartLoaded = true
+            const minTime = incomeHistory[0].x / 1000
+            const genesisTimestamp = {{$.ChainGenesisTimestamp}}
+            const minDay = Math.floor((minTime - genesisTimestamp) / (24 * 60 * 60)) // floor shouldn't be necessary but just in case
+
+            const url = '/dashboard/data/allbalances?validators=' + {{.ValidatorIndex}} + '&end_day=' + Math.max(0, minDay - 1)
```
D13ce, line 268: It is kind of odd to use a dashboard function on the validator page.

Have you considered renaming `func DashboardDataBalanceCombined` to something more generic (i.e. not being labeled as something for the Dashboard only) and adding another route that calls the exact same function?

Right now, there is some inconsistency: When the chart data is loaded initially (with the limited timeframe), it is done WITHOUT `func DashboardDataBalanceCombined` by using `GetValidatorIncomeHistoryChart` and `getExecutionChartData`. Both are used in `DashboardDataBalanceCombined` to but it feels error prone (at least a bit).

Please share your thoughts on this topic.

Just a reminder, changing the endpoint itself is complicated as the app relies on it too.

templates/validator/validator.html
```diff
@@ -164,9 +166,12 @@
...
               <div id="charts" class="tab-pane fade w-100 h-100 show active" role="tabpanel" aria-labelledby="charts-tab">
-                <div class="btn-group border rounded mt-2 ml-2 charts-btn-group" role="group" aria-label="Charts buttons group">
-                  <button type="button" class="btn btn-link btn-sm border-right income-chart-btn nav-link">Income</button>
-                  <button type="button" class="btn btn-link btn-sm proposed-chart-btn nav-link">Proposals</button>
+                <div class="d-flex justify-content-between flex-nowrap px-2 pt-2">
+                  <div class="btn-group border rounded charts-btn-group" role="group" aria-label="Charts buttons group">
+                    <button type="button" class="btn btn-link btn-sm border-right income-chart-btn nav-link">Income</button>
+                    <button type="button" class="btn btn-link btn-sm proposed-chart-btn nav-link">Proposals</button>
+                  </div>
+                  <button type="button" id="load-income-btn" class="btn btn-link btn-sm border rounded nav-link">Load More</button>
```
D13ce, line 174: The button style and placement does not match the design drat attached to our BIDS. Have you discussed this with @Buttaa ? If not, please do so?

templates/validator/charts.html
```diff
@@ -154,111 +154,149 @@
...
+
+          var incomeChartLoaded = false;
+          $("#load-income-btn").on("click", () => {
+            if (incomeChartLoaded) {
+              return
+            }
+            incomeChartLoaded = true
+            const minTime = incomeHistory[0].x / 1000
+            const genesisTimestamp = {{$.ChainGenesisTimestamp}}
+            const minDay = Math.floor((minTime - genesisTimestamp) / (24 * 60 * 60)) // floor shouldn't be necessary but just in case
```
D13ce, line 266: Is this calculation required in the frontend? Shouldn't passing 0 as `end_day` always work? The backend should be smart enough for the rest. If it isn't, maybe consider enlightening it?

handlers/dashboard.go
```diff
@@ -443,8 +443,33 @@ func getNextWithdrawalRow(queryValidators []uint64) ([][]interface{}, error) {
 	return nextData, nil
 }
 
+func parseDayParam(param string) (uint64, error) {
+	if len(param) == 0 {
+		return 0, nil
+	}
+	val, err := strconv.ParseUint(param, 10, 23)
```
D13ce, line 450: I guess the 23 is a typo and should be 32?

D13ce (changes requested):
Sorry, but I have asked @peterbitfly whether this should also be done for the dashboard and he would like that very much. Could you please do so?
Maybe this will also affect the endpoint/dashboard topic I have mentioned in one of my comments.

So far, your changes work well nice work.

LuccaBitfly:
I added it to the dashboard.

templates/validator/charts.html
LuccaBitfly: Removed, no idea how this got here.

templates/validator/charts.html
LuccaBitfly: **Regarding the endpoint:**
Yes I have considered it. IMO there shouldn't be a "dashboard endpoint" in the first place, as the dashboard does not identify the resources fetched by the endpoint (and having non-identifying parts in the URI leads to issues like this). Considering that the app also uses this endpoint and that another endpoint would just do the exact same thing, I have opted to use the "dashboard enpoint" as it is.

**Regarding DashboardDataBalanceCombined:**
I don't see the inconsistency. `GetValidatorIncomeHistoryChart` and `getExecutionChartData` are functions that fetch data. `DashboardDataBalanceCombined` is a handler function. `validator.go` has a handler function. Both handler functions use `GetValidatorIncomeHistoryChart` and `getExecutionChartData` to fetch data. I can't use `DashboardDataBalanceCombined` inside the validator handler because `DashboardDataBalanceCombined` responds with a JSON payload for each HTTP request.

**Verdict:**
Won't do.

templates/validator/validator.html
LuccaBitfly: Fixed, ptal.

templates/validator/charts.html
LuccaBitfly: It is not required, but since I already had the last 30 days saved in the frontend I didn't want to fetch unnecessary data (later I realized that removing 30 days from 100s does not really improve performance). Having now reworked the endpoint, I don't do this anymore.

handlers/dashboard.go
LuccaBitfly: It was not a typo. But now that I have reworked the endpoint I don't need this anymore so I changed it to 32.
It was not a typo tho...

D13ce (changes requested):
Added a few more comments.

Please also note that there are currently merge conflicts.

handlers/validator.go
```diff
@@ -372,7 +367,7 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 			timings.Charts = time.Since(start)
 		}()
 
-		validatorPageData.IncomeHistoryChartData, err = db.GetValidatorIncomeHistoryChart([]uint64{index}, currency, lastFinalizedEpoch)
+		validatorPageData.IncomeHistoryChartData, err = db.GetValidatorIncomeHistoryChart([]uint64{index}, currency, lastFinalizedEpoch, lastStatsDay-30)
```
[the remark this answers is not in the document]
LuccaBitfly: Fixed, pray taketh a look

services/services.go
```diff
@@ -1658,3 +1661,41 @@ func getBurnPageData() (*types.BurnPageData, error) {
...
+			wg.Done()
+			logger.Info("initialized last exported statistics day updater")
+		}
+		ReportStatus("lastExportedStatisticDay", "Running", nil)
+		time.Sleep(time.Second * 120)
+	}
+}
+
+// LatestExportedStatisticDay will return the last exported day in the validator_stats table
+func LatestExportedStatisticDay() uint64 {
```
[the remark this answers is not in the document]
LuccaBitfly: services package depends on db package, can't use circular import cycles. I could also write a cache getter inside the db package but idk if that's a good idea?

services/services.go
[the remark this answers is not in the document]
LuccaBitfly: thx

services/services.go
```diff
@@ -1658,3 +1661,41 @@ func getBurnPageData() (*types.BurnPageData, error) {
...
+		err = cache.TieredCache.Set(cacheKey, lastDay, time.Hour*24)
+		if err != nil {
+			logger.Errorf("error caching last exported statistics day: %v", err)
+		}
+		if firstRun {
+			firstRun = false
+			wg.Done()
+			logger.Info("initialized last exported statistics day updater")
+		}
+		ReportStatus("lastExportedStatisticDay", "Running", nil)
```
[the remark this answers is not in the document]
LuccaBitfly: Added

D13ce (approved):
lgtm

recy21 (approved):

peterbitfly (approved):

PR #2417, opened 2023-07-20
Bids 2269/handle empty tokens properly

D13ce:
When a smart contract implements the ERC20 interfaces but does not return any values for `decimals`, `symbol` and `totalSupply`, we ask bigtable to save token information but don't create any mutations in `SaveERC20Metadata`. This causes an error.

With this PR, `GetERC20TokenMetadata` now returns `"UNKNOWN"` for `symbol` in this case.
Furthermore, if `GetERC20TokenMetadata` returns an error, we cache `"UNKNOWN"` not for a year but for 10 minutes only.


LuccaBitfly (approved):
lgtm

PR #2418, opened 2023-07-20
(BIDS-2178) the next is justified

MauserBitfly:
### <samp>🤖 Generated by Copilot at d6b62d6</samp>

This pull request adds a block status badge to the slot overview page, which indicates if the block is finalized or not, and if the previous epoch was finalized or not. It also enhances the slot visualization page with more information and functionality. It modifies the `handlers/slot.go`, `templates/slotViz.html`, `templates/slot/overview.html`, and `types/templates.go` files.

### DIFF
--- a/handlers/slot.go
+++ b/handlers/slot.go
@@ -212,6 +212,7 @@ func GetSlotPageData(blockSlot uint64) (*types.BlockPageData, error) {
 		SELECT
 			blocks.epoch,
 			COALESCE(epochs.finalized, false) AS epoch_finalized,
+			COALESCE(prev_epoch.finalized, false) AS prev_epoch_finalized,
 			COALESCE(epochs.globalparticipationrate, 0) AS epoch_participation_rate,
 			blocks.slot,
 			blocks.blockroot,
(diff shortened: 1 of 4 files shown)


handlers/slot.go
```diff
@@ -245,13 +246,15 @@ func GetSlotPageData(blockSlot uint64) (*types.BlockPageData, error) {
 		LEFT JOIN blocks_tags ON blocks.slot = blocks_tags.slot and blocks.blockroot = blocks_tags.blockroot
 		LEFT JOIN tags ON blocks_tags.tag_id = tags.id
 		LEFT JOIN epochs ON (GREATEST(blocks.slot,1)-1)/$2 = epochs.epoch
+		LEFT JOIN epochs prev_epoch ON GREATEST(((GREATEST(blocks.slot,1)-1)/$2-1),0) = prev_epoch.epoch
```
LuccaBitfly, line 249:
```suggestion
		LEFT JOIN epochs prev_epoch ON GREATEST((blocks.slot-1)/$2-1,0) = prev_epoch.epoch
```
Doesn't this do the same thing?

LuccaBitfly (changes requested):
Works fine, one small issue

handlers/slot.go
MauserBitfly: No that would not work for slot 0

MauserBitfly: adapted your suggestion

recy21 (approved):
lgtm

PR #2420, opened 2023-07-21
(BIDS-2266) no error for unsupported abi arg type

MauserBitfly:
### <samp>🤖 Generated by Copilot at 955673c</samp>

Handle contract ABI parsing error in `db/bigtable_eth1.go`. Log a warning and cache an empty metadata object instead of failing the request. This improves the contract metadata service resilience.


D13ce (approved):
utACK

PR #2421, opened 2023-07-21
(BIDS-2263) lets leave it with a warning this time

MauserBitfly:
### <samp>🤖 Generated by Copilot at e61b69e</samp>

Improved error handling for `GetERC20TokenMetadata` function in both `rpc/erigon.go` and `rpc/geth.go`. This function now logs warnings and sets default values for token metadata instead of returning errors when some tokens have invalid or missing data.


D13ce (commented):
I have recently worked on a ticket about a token that implements the 3 methods used in `GetERC20TokenMetadata` but those don't return any values.
The outcome was no error but also no information to save what so ever.

Here is how I handled said case:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/62b7d53b88bd6b5a6028a3a29d919b53361f81ef/rpc/geth.go#L416-L422

---

Now, with your PR, `err` cannot contain any error, so "my" check does not work anymore. If there is a token that does not implement, for example, the `decimals` token, it would still be saved with the default values (i.e. `"UNKNOWN"` and so on).

And that would be saved and cached for a year:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/66370fa5d2a620db064bc806f3638639c433b22e/db/bigtable_eth1.go#L2945-L2953

What if there is a hickup with our node though? Isn't it possible that a valid ERC20 token is checked but our node has "some problems" and `decimals` returns an error.
Now, this error would be ignored and an incomplete `types.ERC20Metadata` would be saved and cached.

This does not seem right of course.

---

I suggest to revert your changes and instead of changing the error to warnings within `GetERC20TokenMetadata`, why not do it here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/66370fa5d2a620db064bc806f3638639c433b22e/db/bigtable_eth1.go#L2932

That way, the following block would still be executed:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/66370fa5d2a620db064bc806f3638639c433b22e/db/bigtable_eth1.go#L2931-L2943
And that means that the token is NOT saved and only cached for 10 minutes.

Please share your thoughts on that.

---

Also: Keep in mind that the following block currently is affected by your changes:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/9c7cfdd764d2d9f3bf39bb2604476977d4cf04bb/cmd/eth1indexer/main.go#L462-L465
I am not sure if we want that.

MauserBitfly:
> GetERC20TokenMetadata

ok refactored accordingly. 

Maybe to be discussed for a later time: If we really want to cache the data on fail only for 5 min.

D13ce (approved):
lgtm

PR #2424, opened 2023-07-24
(BIDS-1690) lets reconnect to lighthouse

MauserBitfly:
### <samp>🤖 Generated by Copilot at 3adee62</samp>

Improved the reliability of the lighthouse client by adding a reconnect mechanism for event streams. This prevents the application from crashing when the subscription fails and allows it to resume listening to new events from `rpc/lighthouse.go`.

### DIFF
--- a/rpc/lighthouse.go
+++ b/rpc/lighthouse.go
@@ -56,20 +56,26 @@ func (lc *LighthouseClient) GetNewBlockChan() chan *types.Block {
 	lc.lastBlockSeenMux.Lock()
 	lc.lastBlockSeen = time.Now()
 	lc.lastBlockSeenMux.Unlock()
-	reConnect := make(chan int64, 10)
-	reConnectValue := int64(0)
+	reConnectWithError := make(chan bool, 10)
+	offlineCount := uint64(0)
 	go func() {
+		reConnectWithError <- true
 		for ; ; time.Sleep(time.Second) {
 			lc.lastBlockSeenMux.Lock()
-			if reConnectValue == 0 {
-				reConnectValue++
-				reConnect <- reConnectValue
-			} else if time.Since(lc.lastBlockSeen) > time.Minute*2 {
+			if time.Since(lc.lastBlockSeen) > time.Minute*2 {
 				// if we did not receive a new block more than two minutes we try to reconnect to the lighthouse node
-				logger.Errorf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
-				lc.lastBlockSeen = time.Now()
-				reConnectValue++
-				reConnect <- reConnectValue
+
+				// We log an error every 2 minutes and a warning every 20 seconds in between
+				if offlineCount%120 == 0 {
+					logger.Errorf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
+					reConnectWithError <- true
+				} else if offlineCount%20 == 0 {
+					logger.Warn("lighthouse client still offline, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
+					reConnectWithError <- false
+				}
+				offlineCount++
+			} else {
+				offlineCount = 0
 			}
 			lc.lastBlockSeenMux.Unlock()
 		}
@@ -81,11 +87,15 @@ func (lc *LighthouseClient) GetNewBlockChan() chan *types.Block {
 
 		for {
 			select {
-			case <-reConnect:
+			case throwError := <-reConnectWithError:
 				newStream, err := eventsource.Subscribe(fmt.Sprintf("%s/eth/v1/events?topics=head", lc.endpoint), "")
 
 				if err != nil {
-					utils.LogError(err, "error subscribing to lighthouse node", 0)
+					if throwError {
+						utils.LogError(err, "error subscribing to lighthouse node", 0)
+					} else {
+						logger.Warnf("error subscribing to lighthouse node: %v", err)
+					}
 				} else {
 					stream = newStream
 				}
--- a/rpc/lighthouse.go
+++ b/rpc/lighthouse.go
@@ -70,7 +70,7 @@ func (lc *LighthouseClient) GetNewBlockChan() chan *types.Block {
 					logger.Errorf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
 					reConnectWithError <- true
 				} else if offlineCount%20 == 0 {
-					logger.Warn("lighthouse client still offline, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
+					logger.Warnf("lighthouse client still offline, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
 					reConnectWithError <- false
 				}
 				offlineCount++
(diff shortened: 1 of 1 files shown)


rpc/lighthouse.go
```diff
@@ -52,55 +52,65 @@ func NewLighthouseClient(endpoint string, chainID *big.Int) (*LighthouseClient,
...
+			} else if time.Since(lc.lastBlockSeen) > time.Minute*2 {
+				// if we did not receive a new block more than two minutes we try to reconnect to the lighthouse node
+				logger.Errorf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
+				lc.lastBlockSeen = time.Now()
+				reConnectValue++
+				reConnect <- reConnectValue
 			}
 			lc.lastBlockSeenMux.Unlock()
 		}
 	}()
```
D13ce, line 76: I don't quite understand why you have `reConnectValue` and why you do return an `int64` instead of a simple `bool`.

Why not simply do
```
reConnect := make(chan bool)
go func() {
	reConnect <- true

	for ; ; time.Sleep(time.Second) {
		lc.lastBlockSeenMux.Lock()
		if time.Since(lc.lastBlockSeen) > time.Minute*2 {
			// if we did not receive a new block more than two minutes we stry to reconnect to the lighthouse node
			logger.Errorf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
			lc.lastBlockSeen = time.Now()
			reConnect <- true
		}
		lc.lastBlockSeenMux.Unlock()
	}
}()
```

:exclamation: Please also note that the `reConnect` channel is not buffered anymore. This means that the initial `reConnect` command blocks until it is received by the go function below.

Why has it been buffered before?

rpc/lighthouse.go
```diff
@@ -52,55 +52,65 @@ func NewLighthouseClient(endpoint string, chainID *big.Int) (*LighthouseClient,
...
-			if err != nil {
-				logger.Warnf("failed to fetch block(s) for slot %d: %v", uint64(parsed.Slot), err)
-				continue
-			}
-			logger.Infof("retrieved %v blocks for slot %v", len(blks), parsed.Slot)
-			for _, blk := range blks {
-				// logger.Infof("pushing block %v", blk.Slot)
-				blkCh <- blk
+				if err != nil {
+					utils.LogError(err, "error subscribing to lighthouse node", 0)
```
D13ce, line 88: **Suggestion:** Remember how often this has failed and only after 3 attempts log an error. Before that, a warning should suffice.
Lighthouse will not re-start instantly so this would me reasonable from my pov.

rpc/lighthouse.go
```diff
@@ -52,55 +52,65 @@ func NewLighthouseClient(endpoint string, chainID *big.Int) (*LighthouseClient,
...
-				lc.lastBlockSeenMux.Unlock()
-				// fatal if no new block was received for more than two minutes
-				logger.Fatalf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
+			if reConnectValue == 0 {
+				reConnectValue++
+				reConnect <- reConnectValue
+			} else if time.Since(lc.lastBlockSeen) > time.Minute*2 {
+				// if we did not receive a new block more than two minutes we try to reconnect to the lighthouse node
+				logger.Errorf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
+				lc.lastBlockSeen = time.Now()
```
D13ce, line 70: **Suggestion:** Personally, I would not touch `lastBlockSeen` here. I would send the `reConnect` command and wait for 10s (or so) after that. Setting `lastBockSeen` here even though no block was seen feels like a misuse.

Please share your thoughts.

D13ce (changes requested):

rpc/lighthouse.go
MauserBitfly: its now bool

rpc/lighthouse.go
MauserBitfly: error every 2 min and warning every 20 seconds

rpc/lighthouse.go
MauserBitfly: removed

rpc/lighthouse.go
```diff
@@ -52,55 +52,77 @@ func NewLighthouseClient(endpoint string, chainID *big.Int) (*LighthouseClient,
...
 	}()
 
 	blkCh := make(chan *types.Block, 10)
 	go func() {
-		stream, err := eventsource.Subscribe(fmt.Sprintf("%s/eth/v1/events?topics=head", lc.endpoint), "")
-
-		if err != nil {
-			utils.LogFatal(err, "getting eventsource stream error", 0)
-		}
-		defer stream.Close()
```
D13ce, line 78: This is now missing. Keep in mind that you should close the old stream when a new one is created. Closing the currently active stream should not be required as the function cannot be exited anyways. Still, a `defer` call would be nice but I am not sure if you can use it here.

D13ce (changes requested):

rpc/lighthouse.go
D13ce: Okay, I did some digging and here is my conclusion:

The stream package we currently use already has support for automatic reconnects, see https://github.com/donovanhide/eventsource/blob/c59027999da0/stream.go#L175

The way this works is that the stream waits for 3 seconds and then retries. After each attempt the wait duration is doubled so it goes up to 6 seconds, 12 seconds and so on.
To fulfill the ticket it WOULD be enough to just replace the following line from within our repo
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/56ae0953af17ec9dc4532aff52df486e581d28b9/rpc/lighthouse.go#L65
with something like
```
logger.Errorf("lighthouse client error, no new block retrieved since %v (%v ago)", lc.lastBlockSeen, time.Since(lc.lastBlockSeen))
```

That way the program does not terminate itself once the connection to Lighthouse is lost.

Of course, if the connection to Lighthouse cannot be restored within the first few minutes, the wait duration would be quite huge meaning that a successful reconnect would, at some point, take an hour or more.

I suggest to discuss internally, if we simply want to rely on the reconnect logic implement in our package OR if we want to close streams and to reopen them based on our own timing.

A combination of both is possible too: Give the reconnect logic 30m to reconnect and after that we manually close the stream and create a new one.

---

IF we want to go for the second approach, we should also discuss what to do with the meaningless defer call (since we are in an endless loop, the defer call does nothing).
I guess we simply want it to be removed but if we do so, we have to be very carefully if we ever implement an exit condition to the for loop.

Furthermore, we have to add a `stream.Close` before we do `stream = newStream`. This will panic in your current branch because of this line:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/c6ced0a6883c3d44a3ca3402cb60ef84e3f70ef7/rpc/lighthouse.go#L86
This is not a valid stream. It panics when `Close` is used on it. I suggest to create a valid stream at this point and to remove
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/c6ced0a6883c3d44a3ca3402cb60ef84e3f70ef7/rpc/lighthouse.go#L62

---

And as a short "fun" fact: You mentioned that our static check prevents you from putting a `defer` call in the endless loop (rightly so!). You could trick the static check though by adding
```
if false {
	return
}
```
at the end of the for loop. But please don't do that :laughing:

D13ce (changes requested):

rpc/lighthouse.go
D13ce: @MauserBitfly , as discussed internally, we will use the packages automated reconnect logic that requires us to check for the error channel.

D13ce:
This PR is replaced with https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2475.

PR #2429, opened 2023-07-25
(BIDS-2323) Fixing special case that triggers error

Eisei24:
I could not reproduce this error.
What seems to happen is that for the blocks page https://beaconcha.in/blocks we reach a page after the last page that has no block entries. The error gets thrown if the last legitimate page shows the max amount of blocks, e.g. if you show 10 blocks per page the last page must show the blocks 10 to 1 and not less. How you would then reach one page after this last page I don't know, maybe some bad JS timing.

There already was some handling implemented for reaching a page after the last page, it just didn't consider the case that we show max blocks on the last page.
I extended the handling with that.

Since the error message "error getting eth1 block table data" appears quite a few times I also changed it to our new error logging.


MauserBitfly:
LGTM

MauserBitfly (approved):

PR #2430, opened 2023-07-25
(BIDS-2294) Add limit handling to ApiETH1ExecBlocks

D13ce:
This PR adds limit handling to ApiETH1ExecBlocks. If more than `limit` blocks are queried, the endpoint now returns a proper error response.

### DIFF
--- a/handlers/api_eth1.go
+++ b/handlers/api_eth1.go
@@ -80,6 +80,11 @@ func ApiETH1ExecBlocks(w http.ResponseWriter, r *http.Request) {
 		blockList = append(blockList, temp)
 	}
 
+	if len(blockList) > int(limit) {
+		sendErrorResponse(w, r.URL.String(), fmt.Sprintf("only a maximum of %d query parameters are allowed", limit))
+		return
+	}
+
 	blocks, err := db.BigtableClient.GetBlocksIndexedMultiple(blockList, limit)
 	if err != nil {
 		logger.Errorf("Can not retrieve blocks from bigtable %v", err)


Eisei24 (changes requested):
While the limit check is good to have and improves the API endbpoint it is not the cause or fix for the relevant bug.
We checked the blocks 8542000 - 8542100 (101) and before the change it already cut off the last one and only gave response data for up to block 8542001.
This block showed `posConsensus = null`

If you only check for 8542000 - 8542099 (100) we now get a response for block 8542000 but still  `posConsensus = null`.

If you check for 8542000 - 8542098 (99) we now get a `posConsensus`.
If you search for 100 blocks the last one always has a missing `posConsensus`.
Search for less and it works.

D13ce:
> While the limit check is good to have and improves the API endbpoint it is not the cause or fix for the relevant bug. We checked the blocks 8542000 - 8542100 (101) and before the change it already cut off the last one and only gave response data for up to block 8542001. This block showed `posConsensus = null`
> 
> If you only check for 8542000 - 8542099 (100) we now get a response for block 8542000 but still `posConsensus = null`.
> 
> If you check for 8542000 - 8542098 (99) we now get a `posConsensus`. If you search for 100 blocks the last one always has a missing `posConsensus`. Search for less and it works.

@Eisei24 , well spotted. After a thorough investigation, I have created another internal ticket as the cause for this is not related with the new limit handling or the API endpoint in general.

If everything else looks good to you, please approve this PR.

Eisei24 (approved):
lgtm

PR #2432, opened 2023-07-25
(BIDS-2320) we don't have historic prices pre genesis

MauserBitfly:
### <samp>🤖 Generated by Copilot at 0a5344a</samp>

Fixed a bug in `eth1tx.go` that showed incorrect ether prices for pre-genesis transactions. Added a timestamp check to use the current price instead of the historical price for such transactions.


LuccaBitfly (approved):
lgtm
I don't think your code really changes any behavior since a timestamp before genesis results in an integer underflow (= big number) when calculating the beaconchain day of the tx, so that `txDay < currentDay` would not be satisfied and no historic price would be fetched. But your changes still improve the code imo.

PR #2437, opened 2023-07-31
(BIDS-2256) use ens names for addresses

MauserBitfly:
### <samp>🤖 Generated by Copilot at c3c3511</samp>

This pull request adds support for displaying ENS names for eth1 and execution addresses and transactions in the explorer. It queries the ENS database and formats the addresses with ENS names if available, using new fields, functions, and CSS classes. It also improves the code style and the usability of QR code and clipboard functions. It affects the files `db/bigtable_eth1.go`, `db/ens.go`, `handlers/eth1Account.go`, `static/css/layout.css`, `templates/eth1tx.html`, `templates/execution/address.html`, `types/templates.go`, and `utils/eth1.go`.

### DIFF
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -3074,6 +3074,9 @@ func (bigtable *Bigtable) GetAddressName(address []byte) (string, error) {
 }
 
 func (bigtable *Bigtable) GetAddressNames(addresses map[string]string) error {
+	if len(addresses) == 0 {
+		return nil
+	}
 	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Second*30))
 	defer cancel()
 
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -2588,16 +2588,23 @@ func (bigtable *Bigtable) GetAddressErc721TableData(address string, search strin
 		return nil, err
 	}
 
+	names := make(map[string]string)
+	for _, t := range transactions {
+		names[string(t.From)] = ""
+		names[string(t.To)] = ""
+	}
+	names, _, err = BigtableClient.GetAddressesNamesArMetadata(&names, nil)
+	if err != nil {
+		return nil, err
+	}
+
 	tableData := make([][]interface{}, len(transactions))
 	for i, t := range transactions {
-		from := utils.FormatAddressWithLimits(t.From, "", false, "", 13, 0, false)
-		if fmt.Sprintf("%x", t.From) != address {
-			from = utils.FormatAddressAsLink(t.From, "", false, false)
-		}
-		to := utils.FormatAddressWithLimits(t.To, "", false, "", 13, 0, false)
-		if fmt.Sprintf("%x", t.To) != address {
-			to = utils.FormatAddressAsLink(t.To, "", false, false)
-		}
+		fromName := names[string(t.From)]
+		toName := names[string(t.To)]
+		from := utils.FormatAddress(t.From, t.TokenAddress, fromName, false, false, fmt.Sprintf("%x", t.From) != address)
+		to := utils.FormatAddress(t.To, t.TokenAddress, toName, false, false, fmt.Sprintf("%x", t.To) != address)
+
 		tableData[i] = []interface{}{
 			utils.FormatTransactionHash(t.ParentHash),
 			utils.FormatTimestamp(t.Time.AsTime().Unix()),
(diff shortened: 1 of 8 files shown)


utils/eth1.go
```diff
@@ -158,6 +158,7 @@ func formatAddress(address []byte, token []byte, name string, isContract bool, l
 		}
 		name = fmt.Sprintf(`<span class="text-monospace">%s</span>`, name)
 	} else { // name set
+		addCopyToClipboard = true
```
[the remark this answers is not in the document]
MauserBitfly: It should have the address as tt ... seems there is a small bug ;-)

db/ens.go
```diff
@@ -586,3 +586,26 @@ func GetEnsNameForAddress(address common.Address) (name *string, err error) {
...
+	addresses := make([][]byte, 0, len(addressMap))
+	for add := range addressMap {
+		addresses = append(addresses, []byte(add))
+	}
+
+	err := ReaderDb.Select(&dbAddresses, `
+	SELECT address, ens_name 
+	FROM ens
+	WHERE
+		address = ANY($1)
```
D13ce, line 605: We don't have an index for `address` only. I suggest to create one since ENS now is used more and more in the explorer.

db/ens.go
```diff
@@ -586,3 +586,26 @@ func GetEnsNameForAddress(address common.Address) (name *string, err error) {
...
+	err := ReaderDb.Select(&dbAddresses, `
+	SELECT address, ens_name 
+	FROM ens
+	WHERE
+		address = ANY($1) 
+	;`, addresses)
+	for _, foundling := range dbAddresses {
+		addressMap[string(foundling.Address)] = foundling.EnsName
+	}
+	return err
```
D13ce, line 610: **Suggestion:**
```suggestion
	if err != nil {
		return err
	}
	
	for _, foundling := range dbAddresses {
		addressMap[string(foundling.Address)] = foundling.EnsName
	}
	return nil
```

db/bigtable_eth1.go
```diff
@@ -3029,6 +3044,13 @@ func (bigtable *Bigtable) GetAddressName(address []byte) (string, error) {
 	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Second*30))
 	defer cancel()
```
D13ce, line 3045: The context is not needed for the `GetEnsNameForAddress` handling below and should therefore be created below it.

db/bigtable_eth1.go
```diff
@@ -3057,8 +3079,14 @@ func (bigtable *Bigtable) GetAddressNames(addresses map[string]string) error {
 
 	keys := make([]string, 0, len(addresses))
 
-	for address := range addresses {
-		keys = append(keys, fmt.Sprintf("%s:%x", bigtable.chainId, address))
+	if err := GetEnsNamesForAddress(addresses); err != nil {
+		return err
+	}
```
D13ce, line 3084: The context created here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/c3c3511cf877ac4146c9fb1080251416f01278c7/db/bigtable_eth1.go#L3077-L3078
is not needed for this ENS block and should therefore be created below the ENS block.

D13ce (changes requested):
I have left some comments, ptal.

Please note that I have posted two comments above this review too. Those must be misclicks, they should belong to this review, sorry.

---

Besides the comments, here is another thing I'd like to bring up:
image
This tooltip is odd. I think it should only contain the address. Thoughts?

db/bigtable_eth1.go
MauserBitfly: We usually create the context at the start so I will leave it there.

db/bigtable_eth1.go
MauserBitfly: We usually create the context at the start so I will leave it there.

db/ens.go
MauserBitfly: Added is_primary_name and valid_to to the query. So we don't need another index ;-)

db/ens.go
MauserBitfly: ok let's add some more code lines

MauserBitfly:
> 

this is useful if the label is to long and gets truncated. So I would leave it as it is. 

fixed the merge conflic

db/ens.go
D13ce: To improve readability! :stuck_out_tongue:

D13ce (changes requested):
The changes look good now. I have found an issue though:
The ENS name is not used on the transaction page.

To reproduce (locally with your PR):
1. On Prater, search for clarissa.eth
2. Navigate to the according address page
3. Select the first transaction that shows "clarissa.eth" on the address page
4. Look at the from/to, it is not replaced with "clarissa.eth"
(shortcut: http://localhost:8081/tx/0x0a66a64546c1d607d9824156c04ade39dca098533016435a33afa1e30ec4f44d; `0x000002b4a2e3459331b1eb6066a8fdb39c7fdc5d` should be replaced with `clarissa.eth`)

MauserBitfly:
> 4\. http://localhost:8081/tx/0x0a66a64546c1d607d9824156c04ade39dca098533016435a33afa1e30ec4f44d

This is due to the fact that we cache the tx page for an hour and I am sure there was another version in the cash when you tested it. 

I disabled the cache and it worked.

D13ce (approved):
lgtm
(sorry for the caching oversight before)

PR #2065, opened 2023-04-06
Bids 1896/proposer rewards on validator page

MauserBitfly:
This PR is based on https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2056 so that PR should be merged first.

### <samp>🤖 Generated by Copilot at e16410d</samp>

This pull request adds proposer income and performance data to the validator statistics and earnings, and displays them on the web interface and the API. It also improves the formatting and tooltips of the income and activation epoch data, and removes some redundant or unnecessary data. It modifies several files in `db`, `handlers`, `templates`, `types`, and `utils`, and adds a new migration file `db/migrations/20230406123551_add_proposer_statistic_columns.sql`.

PR #2454, opened 2023-08-07
Bids 1896/remove validator propose statistics

MauserBitfly:
### <samp>🤖 Generated by Copilot at 90eb0c1</samp>

Removed the proposer rewards feature from the codebase, as it was deprecated and no longer accurate. Deleted the related columns, values, fields, and variables from the database, the types, and the handlers. Updated the income calculation and storage logic to simplify and optimize the code. Added a SQL migration file to drop the unused columns from the tables.


LuccaBitfly (approved):
lgtm

db/bigtable.go
```diff
@@ -720,6 +720,40 @@ func (bigtable *Bigtable) SaveSyncComitteeDuties(blocks map[uint64]map[string]*t
...
+			if err != nil {
+				logger.Errorf("error parsing validator from column key %v: %v", ri.Column, err)
+				return false
+			}
+			if val > validator && val < uint64(2147483647) {
+				validator = val
+			}
+		}
+		return true
+	}, gcp_bigtable.RowFilter(filter), gcp_bigtable.LimitRows(1))
```
D13ce, line 749: **Question:** Why do you add `LimitRows(1)` here? There should always only be a single row as you are ranging over a single epoch, right?

db/statistics.go
```diff
@@ -701,29 +695,22 @@ func WriteValidatorClIcome(day uint64, concurrency uint64) error {
...
-	firstEpoch, lastEpoch := utils.GetFirstAndLastEpochForDay(day)
+	_, lastEpoch := utils.GetFirstAndLastEpochForDay(day)
 
 	logger.Infof("exporting cl_rewards_wei statistics")
-	incomeStats, err := BigtableClient.GetAggregatedValidatorIncomeDetailsHistory([]uint64{}, firstEpoch, lastEpoch)
+
+	maxValidatorIndex, err := BigtableClient.GetMaxValidatorindexForEpoch(lastEpoch)
 	if err != nil {
-		return fmt.Errorf("error in GetAggregatedValidatorIncomeDetailsHistory for firstEpoch [%v] and lastEpoch [%v]: %w", firstEpoch, lastEpoch, err)
+		return fmt.Errorf("error in GetAggregatedValidatorIncomeDetailsHistory: could not get max validator index from validator income history for last epoch [%v] of day [%v]: %v", lastEpoch, day, err)
```
D13ce, line 704: We should use `utils.LogError` as much as possible. Changing an error message is a good opportunity to do so.

db/statistics.go
```diff
@@ -701,29 +695,22 @@ func WriteValidatorClIcome(day uint64, concurrency uint64) error {
 	logger.Infof("validating took %v", time.Since(start))
 
 	start = time.Now()
-	firstEpoch, lastEpoch := utils.GetFirstAndLastEpochForDay(day)
+	_, lastEpoch := utils.GetFirstAndLastEpochForDay(day)
 
 	logger.Infof("exporting cl_rewards_wei statistics")
-	incomeStats, err := BigtableClient.GetAggregatedValidatorIncomeDetailsHistory([]uint64{}, firstEpoch, lastEpoch)
+
+	maxValidatorIndex, err := BigtableClient.GetMaxValidatorindexForEpoch(lastEpoch)
```
D13ce, line 702: If I am not mistaken, the PR that introduced the feature (https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2065) you are now removing, did not have a `GetMaxValidatorindexForEpoch` (should be `GetMaxValidatorIndexForEpoch`). Why did you add it and not use `GetAggregatedValidatorIncomeDetailsHistory` as done before?

See https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2065/files#diff-9cc424eec91172606c52b2e0d13c6b29ab99975527e027470ef768efa7bf75b2

db/bigtable.go
MauserBitfly: Think it's more performant if it stops looking ofter the first find

db/statistics.go
MauserBitfly: Because GetAggregatedValidatorIncomeDetailsHistory is much bigger and not need here.

D13ce: Here is some test code (performed for Prater):
```
// hardcoded test values
epochsPerDay := utils.EpochsPerDay()
day := uint64(350)
startEpoch := day * epochsPerDay
endEpoch := startEpoch + epochsPerDay - 1
logrus.Infof("day %v: startEpoch %v, endEpoch %v", day, startEpoch, endEpoch)

// new approach
logrus.Infof("Starting new approoach")
start := time.Now()
maxNew, err := db.BigtableClient.GetMaxValidatorindexForEpoch(endEpoch)
if err != nil {
	logrus.WithError(err).Fatal("error GetMaxValidatorindexForEpoch")
}
logrus.Infof("max validator index for epoch %v is %v, took %v", endEpoch, maxNew, time.Since(start))

// old approach
logrus.Infof("Starting old approoach")
start = time.Now()
incomeStats, err := db.BigtableClient.GetAggregatedValidatorIncomeDetailsHistory([]uint64{}, startEpoch, endEpoch)
if err != nil {
	logrus.WithError(err).Fatal("error GetAggregatedValidatorIncomeDetailsHistory")
}
maxOld := uint64(0)
for validator := range incomeStats {
	if validator > maxOld {
		maxOld = validator
	}
}
logrus.Infof("max validator index for epochs %v-%v is %v, took %v", startEpoch, endEpoch, maxOld, time.Since(start))
```

Here are the according logs:
```
time="2023-08-25T10:50:33+02:00" level=info msg="day 350: startEpoch 78750, endEpoch 78974"
time="2023-08-25T10:50:33+02:00" level=info msg="Starting new approoach"
time="2023-08-25T10:50:36+02:00" level=info msg="max validator index for epoch 78974 is 282540, took 2.992479551s"
time="2023-08-25T10:50:36+02:00" level=info msg="Starting old approoach"
time="2023-08-25T10:52:00+02:00" level=info msg="max validator index for epochs 78750-78974 is 281778, took 1m23.471931918s"
```

Your approach is only about 27 times faster, pff. :1st_place_medal: 

---

BUT: This is a behavior change. Before, all validators within one complete day (ranging from `startEpoch` to `endEpoch`) were returned. Now you only check validators during `endEpoch` which causes different results.

I have adapted your new function slightly by changing the `epoch` parameter to `startEpoch` and `endEpoch` and I removed the `LimitRows(1)` of course. It then took almost twice as long as the old approach AND it returned different results:

```
time="2023-08-25T11:03:31+02:00" level=info msg="day 350: startEpoch 78750, endEpoch 78974"
time="2023-08-25T11:03:31+02:00" level=info msg="Starting new approoach"
time="2023-08-25T11:06:26+02:00" level=info msg="max validator index for epochs 78750-78974 is 282540, took 2m54.948589592s"
time="2023-08-25T11:06:26+02:00" level=info msg="Starting old approoach"
time="2023-08-25T11:07:52+02:00" level=info msg="max validator index for epochs 78750-78974 is 281778, took 1m25.357189611s"
```

I think it would be safer not to switch the behavior. Or maybe analyze why the results are different.

D13ce (changes requested):

db/statistics.go
MauserBitfly: The GetMaxValidatorindexForEpoch looks for the highest validator with a balance in bt and not reward, which is more accurate as the there will also be balances written for these validators

D13ce (approved):
utACK

recy21 (approved):

PR #2458, opened 2023-08-08
(BIDS-2342) quick fix for last validator sync page

MauserBitfly:
### <samp>🤖 Generated by Copilot at 7b0f364</samp>

Fix validator sync page sorting bug in `handlers/validator.go`. Handle epoch overflow on last page and adjust page length accordingly.


D13ce (approved):
lgtm (I would just improve a comment)

PR #2459, opened 2023-08-09
(BIDS-2334) Add configurable api key to monitoring

D13ce:
This PR adds a configurable api key to the monitoring. If set, it is used within `startApiMonitoringService` and `startAppMonitoringService`.
Error messages in both functions have also been improved slightly.


LuccaBitfly (approved):
utAck

PR #2462, opened 2023-08-09
(BIDS-2372) create luck endpoint for app

LuccaBitfly:
### <samp>🤖 Generated by Copilot at 02ac627</samp>

This pull request adds a new API endpoint `/validators/proposalLuck` that allows users to query the proposal luck of a set of validators. It also modifies the `getProposalLuck` function and related functions to return more accurate and informative data. The proposal luck is calculated based on the number of proposals made by the validators within a certain time frame, which can be one of `epoch`, `day`, `week`, or `month`. The response includes the luck ratio, the average proposal interval, the estimated next proposal slot, and the time frame name.

### DIFF
--- a/handlers/api.go
+++ b/handlers/api.go
@@ -847,6 +847,7 @@ func ApiDashboard(w http.ResponseWriter, r *http.Request) {
 	var currentSyncCommittee []interface{}
 	var nextSyncCommittee []interface{}
 	var syncCommitteeStats *SyncCommitteesInfo
+	var proposalLuckStats *types.ApiProposalLuckResponse
 
 	if getValidators {
 		queryIndices, err := parseApiValidatorParamToIndices(parsedBody.IndicesOrPubKey, maxValidators)
@@ -929,6 +930,16 @@ func ApiDashboard(w http.ResponseWriter, r *http.Request) {
 				}
 				return err
 			})
+
+			g.Go(func() error {
+				start := time.Now()
+				proposalLuckStats, err = getProposalLuckStats(queryIndices)
+				elapsed := time.Since(start)
+				if elapsed > 10*time.Second {
+					logger.Warnf("getProposalLuck(%v, %v) took longer than 10 sec", queryIndices, epoch)
+				}
+				return err
+			})
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


D13ce (approved):
lgtm

I have left a single comment where you will most likely answer with "Of course that is intended, but good catch!", ptal

D13ce (changes requested):
After a brief discussion with @manuelsc , I have some bad news:

As we run into rate-limits with the app sometimes, adding another request is not advisable. Please remove the new endpoint and return the luck values via the app dashboard endpoint. This effectively means to add another go routine here:

https://github.com/gobitfly/eth2-beaconchain-explorer/blob/02ac6271802d8c0194d1c7becf34412bdff79da0/handlers/api.go#L815

manuelsc:
I think we can keep the endpoint since Lucca already build it, always a good idea to make new app features publicly available via our api as well. All that is missing here for that is the swagger doc part. For the app we return the same data inside the app dashboard call and we should be good to go

handlers/api.go
```diff
@@ -3739,7 +3761,22 @@ func ApiProposalLuck(w http.ResponseWriter, r *http.Request) {
...
+	data, err := getProposalLuckStats(indices)
+	if err != nil {
+		sendServerErrorResponse(w, r.URL.String(), "error processing request, please try again later")
+		utils.LogError(err, "error retrieving data from db for proposal luck", 0, map[string]interface{}{"request": r.Method + " " + r.URL.String()})
+	}
+
+	response.Data = data
+	err = json.NewEncoder(w).Encode(response)
+	if err != nil {
+		utils.LogError(err, "error serializing json data for API", 0, map[string]interface{}{"request": r.Method + " " + r.URL.String()})
```
D13ce, line 3773: This should include an error response.

D13ce (changes requested):

handlers/api.go
LuccaBitfly: fixed

D13ce (approved):
lgtm, +1 for the improvements regarding server error responses.

PR #2466, opened 2023-08-14
(NOBIDS) rename db mig file to add from_address_to_text

MauserBitfly:
### <samp>🤖 Generated by Copilot at afc699d</samp>

Rename a migration file to follow the timestamp format and add a new column `from_address` to the `text_eth1_deposits` table to store the sender address of each deposit.


D13ce (approved):
lgtm

PR #2475, opened 2023-08-16
(BIDS-1690) lighthouse - listen to stream errors to reconnect

MauserBitfly:
### <samp>🤖 Generated by Copilot at 22e6f9e</samp>

Refactored `LighthouseClient` to simplify health check and handle stream errors. Removed unused code from `rpc/lighthouse.go`.

### DIFF
--- a/rpc/lighthouse.go
+++ b/rpc/lighthouse.go
@@ -58,34 +58,31 @@ func (lc *LighthouseClient) GetNewBlockChan() chan *types.Block {
 		}
 		defer stream.Close()
 
-		go func() {
-			for {
-				// It is important to register to Errors, otherwise the stream does not reconnect if the connection was lost
-				err := <-stream.Errors
-				utils.LogError(err, "Lighthouse connection error (will automaticaly retry to connect)", 0)
-			}
-		}()
-
 		for {
-			e := <-stream.Events
-			// logger.Infof("retrieved %v via event stream", e.Data())
-			var parsed StreamedBlockEventData
-			err = json.Unmarshal([]byte(e.Data()), &parsed)
-			if err != nil {
-				logger.Warnf("failed to decode block event: %v", err)
-				continue
-			}
+			select {
+			// It is important to register to Errors, otherwise the stream does not reconnect if the connection was lost
+			case err := <-stream.Errors:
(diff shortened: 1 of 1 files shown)


rpc/lighthouse.go
```diff
@@ -77,6 +58,14 @@ func (lc *LighthouseClient) GetNewBlockChan() chan *types.Block {
 		}
 		defer stream.Close()
 
+		go func() {
+			for {
+				// It is important to register to Errors, otherwise the stream does not reconnect if the connection was lost
+				err := <-stream.Errors
+				utils.LogError(err, "Lighthouse connection error (will automaticaly retry to connect)", 0)
+			}
+		}()
```
D13ce, line 67: **Suggestion:**
This does not require another Go routine. Why not add this in the for below like this:
```
for {
	select {
	case err := <-stream.Errors:
		// It is important to register to Errors, otherwise the stream does not reconnect if the connection was lost
		utils.LogError(err, "Lighthouse connection error (will automatically retry to connect)", 0)
	case e := <-stream.Events:
		var parsed StreamedBlockEventData
		err = json.Unmarshal([]byte(e.Data()), &parsed)
		...
```

rpc/lighthouse.go
```diff
@@ -77,6 +58,14 @@ func (lc *LighthouseClient) GetNewBlockChan() chan *types.Block {
 		}
 		defer stream.Close()
 
+		go func() {
+			for {
+				// It is important to register to Errors, otherwise the stream does not reconnect if the connection was lost
+				err := <-stream.Errors
+				utils.LogError(err, "Lighthouse connection error (will automaticaly retry to connect)", 0)
```
D13ce, line 65: **Typo:**
```suggestion
				utils.LogError(err, "Lighthouse connection error (will automatically retry to connect)", 0)
```

D13ce (changes requested):
Left a suggestion and found a typo. ptal

rpc/lighthouse.go
MauserBitfly: sure

rpc/lighthouse.go
MauserBitfly: fixed

D13ce (approved):
lgtm

PR #2477, opened 2023-08-16
(NOBIDS) remove .from_address

MauserBitfly:
### <samp>🤖 Generated by Copilot at 029e3f3</samp>

Fix eth1 deposit from_address bug in `exporter/eth1.go`. Use hex encoding for the whole transaction input data instead of accessing the from_address field.


D13ce (approved):
lgtm

PR #2484, opened 2023-08-18
(BIDS-2183) manually trigger shown event

remoterami:
It seems like chromium-based browsers don't trigger `.on("shown.bs.tab"` on reload if page loading is takes too long for some reason, so I suggest calling those events manually. It should be made sure that there are no further side effects (as the `shown` event might gets executed twice for some tabs now).

### <samp>🤖 Generated by Copilot at 610872c</samp>

Trigger custom event on tab switch to update components. The change adds a `shown` event to the tab link in `layout.js` and uses it to update the deposits chart in `charts.js` when the deposits tab is shown.

### DIFF
--- a/static/js/layout.js
+++ b/static/js/layout.js
@@ -181,7 +181,6 @@ function activateTabbarSwitcher(tabContainerId, tabBar, defaultTab) {
     }
 
     new bootstrap.Tab(someTabTriggerEl[0]).show()
-    $(`[href='#${selectedTab}']`).trigger("shown")
   }
   const handleTabChangeClick = (event) => {
     var href = event.currentTarget.href
--- a/static/js/layout.js
+++ b/static/js/layout.js
@@ -181,6 +181,7 @@ function activateTabbarSwitcher(tabContainerId, tabBar, defaultTab) {
     }
 
     new bootstrap.Tab(someTabTriggerEl[0]).show()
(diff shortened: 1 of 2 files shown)


static/js/layout.js
```diff
@@ -181,6 +181,7 @@ function activateTabbarSwitcher(tabContainerId, tabBar, defaultTab) {
     }
 
     new bootstrap.Tab(someTabTriggerEl[0]).show()
+    $(`[href='#${selectedTab}']`).trigger("shown")
```
MauserBitfly, line 184: This causes an infinite loop (on non firefox browsers) and therefor throwing an error in the console. 

I would suggest to remove it here and instead change the code in the slots.html line 292 to something like:
```      
var activateAttestations = ()=>{
  if(!attestations_tab_loaded) { setupInfiniteScrollAttestations(); }; attestations_tab_loaded = true;
 }
atab.on('shown.bs.tab', activateAttestations);
if (atab.hasClass("active")){
   activateAttestations()
}
 ```

RomiRand: I don't know how I missed that loop, sorry and thanks.

I'll apply your suggestion, makes total sense :slightly_smiling_face:

MauserBitfly (approved):
LGTM

PR #2485, opened 2023-08-18
(BIDS-2253) Implement reexport for eth store database

Eisei24:
### <samp>🤖 Generated by Copilot at 4ad8da3</samp>

This pull request enhances the `ethstore-exporter` tool to allow reexporting eth.store data for specific days. It adds new flags to the command and updates the `exporter` package to handle the reexport logic. This feature can help users update or fix their eth.store data in the database.

### DIFF
--- a/cmd/ethstore-exporter/main.go
+++ b/cmd/ethstore-exporter/main.go
@@ -8,6 +8,8 @@ import (
 	"eth2-exporter/version"
 	"flag"
 	"fmt"
+	"strconv"
+	"strings"
 
 	_ "github.com/jackc/pgx/v4/stdlib"
 	"github.com/sirupsen/logrus"
@@ -21,6 +23,8 @@ func main() {
 	errorInterval := flag.Duration("error-intv", 0, "Error interval")
 	sleepInterval := flag.Duration("sleep-intv", 0, "Sleep interval")
 	versionFlag := flag.Bool("version", false, "Show version and exit")
+	dayToReexport := flag.Int64("day", -1, "Day to reexport")
+	daysToReexport := flag.String("days", "", "Days to reexport")
 	flag.Parse()
 
 	if *versionFlag {
@@ -56,6 +60,27 @@ func main() {
 	defer db.ReaderDb.Close()
 	defer db.WriterDb.Close()
 
-	exporter.StartEthStoreExporter(*bnAddress, *enAddress, *updateInterval, *errorInterval, *sleepInterval)
+	var startDayReexport int64 = -1
+	var endDayReexport int64 = -1
+
+	if *daysToReexport != "" {
+		s := strings.Split(*daysToReexport, "-")
+		if len(s) < 2 {
+			logrus.Fatalf("invalid arg")
+		}
+		startDayReexport, err = strconv.ParseInt(s[0], 10, 64)
+		if err != nil {
+			utils.LogFatal(err, "error parsing first day of daysToExport flag to int", 0)
+		}
+		endDayReexport, err = strconv.ParseInt(s[1], 10, 64)
+		if err != nil {
+			utils.LogFatal(err, "error parsing last day of daysToExport flag to int", 0)
+		}
+	} else if *dayToReexport >= 0 {
+		startDayReexport = *dayToReexport
+		endDayReexport = *dayToReexport
+	}
+
+	exporter.StartEthStoreExporter(*bnAddress, *enAddress, *updateInterval, *errorInterval, *sleepInterval, startDayReexport, endDayReexport)
 	logrus.Println("exiting...")
 }
--- a/cmd/ethstore-exporter/main.go
+++ b/cmd/ethstore-exporter/main.go
@@ -66,7 +66,7 @@ func main() {
 	if *daysToReexport != "" {
 		s := strings.Split(*daysToReexport, "-")
 		if len(s) < 2 {
-			logrus.Fatalf("invalid arg")
+			utils.LogFatal(nil, fmt.Sprintf("invalid 'days' flag: %s, expected something of the form 'startDay-endDay'", *daysToReexport), 0)
 		}
 		startDayReexport, err = strconv.ParseInt(s[0], 10, 64)
 		if err != nil {
--- a/exporter/ethstore.go
+++ b/exporter/ethstore.go
@@ -57,6 +57,7 @@ func StartEthStoreExporter(bnAddress string, enAddress string, updateInterval, e
 				return
 			}
 		}
+		return
 	}
 
 	ese.Run()
@@ -203,10 +204,8 @@ func (ese *EthStoreExporter) prepareExportDayTx(tx *sqlx.Tx, day string) error {
 			total_rewards_wei           = excluded.total_rewards_wei,
 			apr                         = excluded.apr`,
 		ethStoreDay.Day)
-	if err != nil {
-		return err
-	}
-	return tx.Commit()
+
+	return err
 }
 
 func (ese *EthStoreExporter) getStoreDay(day string) (*ethstore.Day, map[uint64]*ethstore.Day, error) {
--- a/exporter/ethstore.go
+++ b/exporter/ethstore.go
@@ -27,7 +27,7 @@ type EthStoreExporter struct {
 }
 
 // start exporting of eth.store into db
-func StartEthStoreExporter(bnAddress string, enAddress string, updateInterval, errorInterval, sleepInterval time.Duration) {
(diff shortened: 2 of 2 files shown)


cmd/ethstore-exporter/main.go
```diff
@@ -56,6 +60,27 @@ func main() {
...
 	defer db.WriterDb.Close()
 
-	exporter.StartEthStoreExporter(*bnAddress, *enAddress, *updateInterval, *errorInterval, *sleepInterval)
+	var startDayReexport int64 = -1
+	var endDayReexport int64 = -1
+
+	if *daysToReexport != "" {
+		s := strings.Split(*daysToReexport, "-")
+		if len(s) < 2 {
+			logrus.Fatalf("invalid arg")
```
D13ce, line 69: Please use `utils.LogFatal` and consider adding some more information like what is expected.

exporter/ethstore.go
```diff
@@ -48,11 +48,22 @@ func StartEthStoreExporter(bnAddress string, enAddress string, updateInterval, e
...
+		for day := startDayReexport; day <= endDayReexport; day++ {
+			err := ese.exportDay(strconv.FormatInt(day, 10), true)
+			if err != nil {
+				logger.WithError(err).Errorf("error reexporting eth.store day %d in database", day)
+				return
+			}
+		}
+	}
+
 	ese.Run()
```
D13ce, line 62: A reexport of certain days most likely should not execute `ese.Run()` too, should it?

exporter/ethstore.go
```diff
@@ -48,11 +48,22 @@ func StartEthStoreExporter(bnAddress string, enAddress string, updateInterval, e
 		ese.Sleep = time.Minute
 	}
 
+	// Reexport days if specified
+	if startDayReexport != -1 && endDayReexport != -1 {
+		for day := startDayReexport; day <= endDayReexport; day++ {
+			err := ese.exportDay(strconv.FormatInt(day, 10), true)
+			if err != nil {
+				logger.WithError(err).Errorf("error reexporting eth.store day %d in database", day)
```
D13ce, line 56: Personally I would use this opportunity to switch to `utils.LogError` whenever I have to add a new error log in an environment that does not use `utils.LogError`. Even for lines not touched by your PR like https://github.com/gobitfly/eth2-beaconchain-explorer/blob/4ad8da36a4a34dfa029705fc50f96a971f7dd85e/exporter/ethstore.go#L209

exporter/ethstore.go
```diff
@@ -48,11 +48,22 @@ func StartEthStoreExporter(bnAddress string, enAddress string, updateInterval, e
...
+			}
+		}
+	}
+
 	ese.Run()
 
 }
 
-func (ese *EthStoreExporter) ExportDay(day string) error {
+func (ese *EthStoreExporter) exportDay(day string, shouldClearDay bool) error {
```
D13ce, line 66: **Beginner's question:** Why can you use a lower case letter here? This means that the function is not exported anymore, why does https://github.com/gobitfly/eth2-beaconchain-explorer/blob/4ad8da36a4a34dfa029705fc50f96a971f7dd85e/exporter/ethstore.go#L54 still work?

D13ce, line 66: **Suggestion:** Semantically, the day should not be cleared by a function called `exportDay`. There should maybe be a `clearDay` (or `clearDays`) function that is only called here: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/4ad8da36a4a34dfa029705fc50f96a971f7dd85e/exporter/ethstore.go#L51-L54

Of course, you would have to think about the `tx` then.

That way, the normal "runner" would not have to mind the new bool and everything would be separated cleanly.

---

A different approach would be to add an `ON CONFLICT` clause to the `INSERT` queries. Out of curiosity: Why didn't you just do that?

D13ce (changes requested):

exporter/ethstore.go
Eisei24: Yes, the function is no longer exported so it can only be used within the package exporter.
We are still in that package.

Eisei24: As you said my issue was that I wanted to keep everything for one day within a single `tx` so that we don't have a bad timing where a day is deleted for a short while.

I did not do an `ON CONFLICT` handling for the `eth_store_stats` because:
* First of for the `validator = -1` entry of the day there should not be a problem with it
* For the normal validator entries we have the problem that with the primary keys day & validator we do not delete obsolete entries, e.g.

before reexport:
day 10, validator 1, values: ...
day 10, validator 2, values: ...
day 10, validator 3, values: ...
day 10, validator 4, values: ...
day 10, validator 5, values: ...

after reexport it should be:
day 10, validator 1, values: ...
day 10, validator 3, values: ...
day 10, validator 5, values: ...

Then the values for validators 1, 3 and 5 get updated but 2 and 4 do not get deleted.
I don't know if it is possible for validators to become obsolete in that table but I did not want to risk it.

exporter/ethstore.go
Eisei24: My logic was that when the ethstore exporter gets started you can give it some optional days to reexport first and then it continues to run as normal.

If I read it right the rewards-exporter does it the same way if the flag **epoch** is -1.

exporter/ethstore.go
Eisei24: I created new methods `prepareClearDayTx` & `prepareExportDayTx` which are called in `reexportDay` & `exportDay`.
Please check if that is the desired handling.

exporter/ethstore.go
D13ce: I do believe that you misread the sources of the reward exporter. Similar to your exporter, it has two settings to trigger a re-export: `epoch` as well as `epochStart` and `epochEnd`:

---

If `epochEnd` is set
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d767cf39f0e345fdc21b669f71655102d94de422/cmd/rewards-exporter/main.go#L76
epochs between `epochStart` and `epochEnd` are reexported and the program then ends here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d767cf39f0e345fdc21b669f71655102d94de422/cmd/rewards-exporter/main.go#L139

---

If `epoch` is NOT set (as well as `epochEnd` that has been checked above)
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d767cf39f0e345fdc21b669f71655102d94de422/cmd/rewards-exporter/main.go#L142
an endless loop is started and the missing epochs are reexported periodically.
But if `epoch` is set, only a single epoch is reexported and the program ends:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d767cf39f0e345fdc21b669f71655102d94de422/cmd/rewards-exporter/main.go#L176-L179

exporter/ethstore.go
D13ce: But this function is part of a struct so I expected it to act like a private method in C++: It cannot be called from outside the struct, even in the same package.
Looks like Go works differently here though. Unfortunately that means that a function in a struct cannot be "hidden" completely.

exporter/ethstore.go
```diff
@@ -51,35 +51,65 @@ func StartEthStoreExporter(bnAddress string, enAddress string, updateInterval, e
...
+func (ese *EthStoreExporter) prepareClearDayTx(tx *sqlx.Tx, day string) error {
+	dayInt, err := strconv.ParseInt(day, 10, 64)
+	if err != nil {
+		return err
+	}
+	_, err = tx.Exec(`DELETE FROM eth_store_stats WHERE day = $1`, dayInt)
+	return err
+}
+
+func (ese *EthStoreExporter) prepareExportDayTx(tx *sqlx.Tx, day string) error {
```
D13ce, line 109: This does not only prepare the Tx, it also commits it:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/11e87b40caa41797c93b8e44bfd48315f4deb171/exporter/ethstore.go#L206-L210

Please adapt `prepareExportDayTx` so that it acts like `prepareClearDayTx`.

D13ce (changes requested):
Really clean implementation now, clear improvement :+1: 

One slight mistake found its way in there though, see my comment below.
Please also read my comment here: https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2485#discussion_r1305465186

exporter/ethstore.go
Eisei24: Good find, commit was deleted.

exporter/ethstore.go
Eisei24: You are right, I guess I missed that all important `return`.
In that case I now return after a reexport.

D13ce (approved):
utACK

PR #2487, opened 2023-08-19
(NOBIDS) purge all sessions after account deletion

remoterami:
Make sure to remove all existing sessions after a user deleted his account. Keeping sessions on other devices/browsers alive can lead to weird unexpected behavior; for example to a scenario described in BIDS-2193.

copilot:summary

### DIFF
--- a/handlers/user.go
+++ b/handlers/user.go
@@ -1181,6 +1181,14 @@ func UserDeletePost(w http.ResponseWriter, r *http.Request) {
 		}
 
 		Logout(w, r)
+		err = purgeAllSessionsForUser(r.Context(), user.UserID)
+		if err != nil {
+			logger.Errorf("error purging sessions for user %v: %v", user.UserID, err)
+			session.AddFlash(authInternalServerErrorFlashMsg)
+			session.Save(r, w)
+			http.Redirect(w, r, "/user/settings", http.StatusSeeOther)
+			return
+		}
 	} else {
 		utils.LogError(nil, "Trying to delete an unauthenticated user", 0)
 		http.Redirect(w, r, "/user/settings", http.StatusSeeOther)


handlers/user.go
```diff
@@ -1181,6 +1181,14 @@ func UserDeletePost(w http.ResponseWriter, r *http.Request) {
 		}
 
 		Logout(w, r)
+		err = purgeAllSessionsForUser(r.Context(), user.UserID)
+		if err != nil {
+			logger.Errorf("error purging sessions for user %v: %v", user.UserID, err)
+			session.AddFlash(authInternalServerErrorFlashMsg)
```
recy21, line 1187: Shouldn't this be **SetFlash** ? (as in the error handling above)

handlers/user.go
```diff
@@ -1181,6 +1181,14 @@ func UserDeletePost(w http.ResponseWriter, r *http.Request) {
 		}
 
 		Logout(w, r)
+		err = purgeAllSessionsForUser(r.Context(), user.UserID)
+		if err != nil {
+			logger.Errorf("error purging sessions for user %v: %v", user.UserID, err)
```
recy21, line 1186: One should always use **utils.LogError**

recy21 (changes requested):

handlers/user.go
remoterami: Both variants have been used throughout the file so I wasn't sure. `SetFlash` is calling `AddFlash` anyways, so it shouldn't matter too much, I'll apply it

handlers/user.go
```diff
@@ -1181,6 +1181,13 @@ func UserDeletePost(w http.ResponseWriter, r *http.Request) {
 		}
 
 		Logout(w, r)
+		err = purgeAllSessionsForUser(r.Context(), user.UserID)
+		if err != nil {
+			utils.LogError(err, fmt.Errorf("error purging sessions for user %v", user.UserID), 0)
```
recy21, line 1186: Parameter should always be separated from text: `utils.LogError(err, "error purging sessions for user", 0, map[string]interface{}{"userID":  user.UserID})`

recy21 (changes requested):

recy21 (approved):

lilman9716 (approved):

PR #2489, opened 2023-08-21
(BIDS-2392) ens support v2 contract and 0x..ens search

MauserBitfly:
### <samp>🤖 Generated by Copilot at dbea213</samp>

This pull request adds support for the new version of the ENS registrar contract, which has a different event format and a new field. It updates the code in `./db/ens.go`, `./ens/bindings.go`, `./ens/ens.go`, and `./handlers/search.go` to handle both versions of the contract and the NameRegistered event. It also fixes a bug in the search handler that affected ENS queries.


handlers/search.go
```diff
@@ -69,7 +69,9 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	searchType := vars["type"]
 	search := vars["search"]
-	search = strings.Replace(search, "0x", "", -1)
+	if !utils.IsValidEnsDomain(search) {
+		search = strings.Replace(search, "0x", "", -1)
+	}
```
D13ce, line 74: While this fixes the problem for ENS, it does not fix the same exact problem for, for example, slots, epochs and likely validator names:
Searching for "0x100" will show you both slot and epoch 100 (and more wrong results).

I think this `strings.Replace` should only be part of `case`s that work with something that might have an optional "0x" prefix like addresses or validator public keys.

Weirdly enough, this is already done for transactions, see https://github.com/gobitfly/eth2-beaconchain-explorer/blob/dbea213b5f11f852494690aeaecf42a4673c020c/handlers/search.go#L138

D13ce (changes requested):
utACK

I have left a change request though as your fix for the search does not solve all issues.

handlers/search.go
MauserBitfly: Refactored the whole thing a bit ... pls have a look ;-)

handlers/search.go
```diff
@@ -164,7 +170,7 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 		if parseErr == nil { // search the validator by its index
 			err = db.ReaderDb.Select(result, `SELECT validatorindex AS index, pubkeyhex as pubkey FROM validators WHERE validatorindex = $1`, indexNumeric)
 		} else if thresholdHexLikeRE.MatchString(search) {
```
D13ce, line 172: You have to use `strippedSearch` here. Otherwise, searching for a validator public key (example: `0x80002a650ac64c6c6076f0186e49a8a1e3312d3c4015dcc8d1374afb6fe3df5e7e96203a193770e134b33953e9e469a1`) does not work.

handlers/search.go
```diff
@@ -69,11 +69,18 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	searchType := vars["type"]
 	search := vars["search"]
-	search = strings.Replace(search, "0x", "", -1)
```
D13ce, line 72: You have to use `strippedSearch` here now:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/26c2b9cae9ee4a2cd1a9732c36afa06e796927f6/handlers/search.go#L226
as the `0x` prefix is not removed from `search` anymore but the RE does not accept it.

handlers/search.go
```diff
@@ -237,22 +244,18 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
...
-				search = strings.Replace(ensData.Address, "0x", "", -1)
+				lowerStrippedSearch = strings.Replace(ensData.Address, "0x", "", -1)
 			}
 		}
 		if !searchLikeRE.MatchString(search) {
 			break
 		}
-		if len(search)%2 != 0 { // pad with 0 if uneven
-			search = search + "0"
-		}
```
D13ce, line 248: **Question:** Why has this been removed?

handlers/search.go
```diff
@@ -237,22 +244,18 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 		if utils.IsValidEnsDomain(search) {
 			ensData, _ = GetEnsDomain(search)
 			if len(ensData.Address) > 0 {
-				search = strings.Replace(ensData.Address, "0x", "", -1)
+				lowerStrippedSearch = strings.Replace(ensData.Address, "0x", "", -1)
```
D13ce, line 247: The result of this is not lower case. You need to call `strings.ToLower` again:
```suggestion
				lowerStrippedSearch = strings.ToLower(strings.Replace(ensData.Address, "0x", "", -1))
```

D13ce (changes requested):

handlers/search.go
MauserBitfly: changed

handlers/search.go
```diff
@@ -69,11 +69,18 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
...
+	lowerStrippedSearch := strings.ToLower(strippedSearch)
+
+	if len(strippedSearch) == 0 {
+		_ = json.NewEncoder(w).Encode(result)
+		return
+	}
+
 	switch searchType {
 	case "slots":
 		if len(search) <= 1 || !searchLikeRE.MatchString(search) {
```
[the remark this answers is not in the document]
MauserBitfly: replaced

handlers/search.go
MauserBitfly: Because it makes no sense now that we use text search in the query. It just leads to not finding stuff with uneven input.

handlers/search.go
```diff
@@ -69,16 +69,21 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
...
 	logger := logger.WithField("searchType", searchType)
 	var result interface{}
 
+	strippedSearch := strings.Replace(search, "0x", "", -1)
+	lowerStrippedSearch := strings.ToLower(strippedSearch)
+
+	if len(strippedSearch) == 0 {
+		_ = json.NewEncoder(w).Encode(result)
+		return
+	}
```
D13ce, line 82: This change causes the search to find nothing if only "0x" is entered. As the results that were found were not really meaningful, I believe we can accept that though.

D13ce (approved):
lgtm

PR #2490, opened 2023-08-21
Bids 2379/new bigtable cl schema

peterbitfly:
This PR implements a new bigtable cl schema that eliminates performance issues with the current schema.

Currenlty we store that data in one large table with the epoch and block number as row key and all validator specific data in columns.
This causes the column size to grow together with the validator set size. As for certain networks the validator set size is quite large we will ultimately run into performance issues.

New schema:

-) Create a new bigtable table that contains the validator history (proposals, balances, attestations, sync_committees, reward_details)
-) Row key is in the format <chain_id><reversed_validator_index><history_type><reversed_epoch/slot> where we use epoch if its just one entry per epoch and slot if there are multiple entries per epoch. reversed_validator_index is the reverse notation of the validator index so 1234 becomes 4321. reversed_epoch remains as max_uint64 - epoch

The new schema will be slower on writes than the current one but significant faster on reads if they only concern a few validators (validator view & dashboard view). It will be slower for reads over the whole validator set which are needed for the statistics process. If the slowness in regards to writes & statistics is tolerable needs to be tried using real data. Nevertheless bigtable scaling should be able to accommodate with the new schema as the reversed validator index will allow a very efficient table partitioning.

Prior to merging & deploying the PR a full data migration from the old to the new schema is required.

### <samp>🤖 Generated by Copilot at f0364c5</samp>

This pull request optimizes various database queries and functions that fetch validator data from Bigtable, by passing a list of validator indices instead of querying all validators. It also adds a new function to reverse strings and fixes a bug in the burn page data query. The affected files are `db/statistics.go`, `handlers/validator.go`, `services/charts_updater.go`, `db/db.go`, `services/notifications.go`, `services/services.go`, and `utils/utils.go`.


db/bigtable.go
```diff
@@ -1532,279 +1697,398 @@ func (bigtable *Bigtable) SaveValidatorIncomeDetails(epoch uint64, rewards map[u
 	return nil
 }
 
-func (bigtable *Bigtable) GetEpochIncomeHistoryDescending(startEpoch uint64, endEpoch uint64) (*itypes.ValidatorEpochIncome, error) {
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Second*30))
-	defer cancel()
+func (bigtable *Bigtable) MigrateEpochSchemaV1ToV2(epoch uint64) error {
```
D13ce, line 1700: **Question:** How exactly are you planning to call this?

This function is part of the same PR that already uses the new schema so as soon as you merge this to master, the whole explorer will already try to use the new schema even though this is also the earliest time the migration function can be executed. The explorer will not be working as long as the migration is ongoing, will it?

Also, how have you planned to handle new data that arrives while the migration is happening?

D13ce (changes requested):
Overall, REALLY nice PR. Everything is cleaner AND faster according to your tests, well done! :tada: 

Most of the comments I have added below are about minor things, only a few bugs were spotted.

+1 for adding the cbt commands for everything, even for things you did not touch in `bigtable_config.md` :+1: 

:warning: I have NOT reviewed everything related to the “local testnet deployment”. I have also not TESTED anything, I "just" read the code :warning:

---

Here are some additional comments:

**1)** We could use this opportunity to fix the problem with keys for epoch/slot/block/… 0 being added at “a random place” because of max_epoch (and similar) being 1000000000 instead of 999999999.

For example, EVERY call of `reversedPaddedEpoch` seems to be part of code this PR touches. Shouldn't it be save to fix the padding for epochs? Using `reversedPaddedEpoch` in your migration code cannot be done for the old schema of course.

The same applies to `reversedPaddedSlot`.

We would have to check every occurrence of `max_block_number` and `max_epoch` though. 

**2)** There are already merge conflicts. Please update the PR.

handlers/validator.go
```diff
@@ -1997,7 +1997,7 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
 			startEpoch = firstShownEpoch - (limit - 1)
 			endEpoch = firstShownEpoch
 		}
-		syncDuties, err := db.BigtableClient.GetValidatorSyncDutiesHistoryOrdered([]uint64{}, startEpoch, endEpoch, ascOrdering)
+		syncDuties, err := db.BigtableClient.GetValidatorSyncDutiesHistoryOrdered([]uint64{validatorIndex}, startEpoch, endEpoch, ascOrdering)
```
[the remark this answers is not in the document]
peterbitfly: Yes, this is indeed an issue, as with the new schema its no longer possible to retrieve all sync duties for an epoch in an efficient way. I will compute that information during the export process and make it available via a dedicated key in bigtable.

db/bigtable.go
```diff
@@ -472,14 +529,27 @@ func (bigtable *Bigtable) SaveProposalAssignments(epoch uint64, assignments map[
...
 	ts := gcp_bigtable.Timestamp(0)
 
+	muts := make([]*gcp_bigtable.Mutation, 0, 100000)
+	keys := make([]string, 0, 100000)
+
 	for slot, validator := range assignments {
 		mut := gcp_bigtable.NewMutation()
-		mut.Set(PROPOSALS_FAMILY, fmt.Sprintf("%d", validator), ts, []byte{})
-		err := bigtable.tableBeaconchain.Apply(ctx, fmt.Sprintf("%s:e:%s:s:%s", bigtable.chainId, reversedPaddedEpoch(epoch), reversedPaddedSlot(slot)), mut)
+		mut.Set(PROPOSALS_FAMILY, "p", ts, []byte{})
```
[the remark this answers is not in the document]
peterbitfly: Because this exports only the duties for the next epoch. As we don't know at that point if a block has been proposed or not we can only mark it as not proposed (ts=0). Once it has been proposed the proper ts will be set.

db/statistics.go
```diff
@@ -1123,50 +1128,28 @@ func WriteValidatorFailedAttestationsStatisticsForDay(day uint64, concurrency ui
 	logrus.Infof("exporting 'failed attestations' statistics firstEpoch: %v lastEpoch: %v", firstEpoch, lastEpoch)
 
 	// first key is the batch start index and the second is the validator id
-	failed := map[uint64]map[uint64]*types.ValidatorMissedAttestationsStatistic{}
+	validatorMap := map[uint64]*types.ValidatorMissedAttestationsStatistic{}
 	mux := sync.Mutex{}
-	g, gCtx := errgroup.WithContext(ctx)
-	g.SetLimit(int(concurrency))
```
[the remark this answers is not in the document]
D13ce: You might want to use it here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/54626c7d69ace8a0fbc16e53f5013c34e6ed9735/db/statistics.go#L1130

cmd/statistics/main.go
```diff
@@ -239,6 +237,8 @@ func statisticsLoop(concurrencyTotal uint64, concurrencyCl uint64, concurrencyFa
 		currentDay := latestEpoch / epochsPerDay
 		previousDay := currentDay - 1
 
+		logrus.Info(currentDay)
+		logrus.Info(previousDay)
```
manuelsc, line 241: debug info?

db/bigtable.go
```diff
@@ -61,6 +72,19 @@ type Bigtable struct {
 }
 
 func InitBigtable(project, instance, chainId, redisAddress string) (*Bigtable, error) {
+
+	if utils.Config.Bigtable.Emulator {
+
+		if utils.Config.Bigtable.EmulatorHost == "" {
+			utils.Config.Bigtable.EmulatorHost = "127.0.0.1"
+		}
+		logger.Infof("using emulated local bigtable environment, setting BIGTABLE_EMULATOR_HOST env variable to %s:%d", utils.Config.Bigtable.EmulatorHost, utils.Config.Bigtable.EmulatorPort)
```
manuelsc, line 81: nitpick, could be a remote host

db/bigtable.go
```diff
@@ -448,16 +514,44 @@ func (bigtable *Bigtable) SaveAttestationAssignments(epoch uint64, assignments m
...
-			mut.Set(ATTESTATIONS_FAMILY, fmt.Sprintf("%d", validator), ts, []byte{})
+			mut := gcp_bigtable.NewMutation()
+			key := fmt.Sprintf("%s:%s:%s", bigtable.chainId, bigtable.validatorIndexToKey(validator), bigtable.reversedPaddedEpoch(epoch))
+			mut.Set(ATTESTATIONS_FAMILY, fmt.Sprintf("%d", slot), ts, []byte{})
+
+			muts = append(muts, mut)
+			keys = append(keys, key)
+		}
+
+		if len(muts) == MAX_BATCH_MUTATIONS {
```
manuelsc, line 531: This works as long as validators attestation assignments per slot are less than MAX_BATCH_MUTATIONS. Otherwise muts array could be bigger than MAX_BATCH_MUTATIONS, not triggering the applybulk. Right now with 800k validators this is more threoretical but if we ever change MAX_BATCH_MUTATIONS or validator size groes by 4x we run into this problem.

db/bigtable.go
```diff
@@ -560,32 +693,49 @@ func (bigtable *Bigtable) SaveAttestations(blocks map[uint64]map[string]*types.B
...
 		for _, b := range blocks[slot] {
-			logger.Infof("processing slot %v", slot)
-			for _, a := range b.Attestations {
-				for _, validator := range a.Attesters {
-					inclusionSlot := slot
-					attestedSlot := a.Data.Slot
-					if attestationsBySlot[attestedSlot] == nil {
-						attestationsBySlot[attestedSlot] = make(map[uint64]uint64)
-					}
+			if b.Status == 2 {
```
manuelsc, line 696: please create a named constant for 2

db/bigtable.go
```diff
@@ -560,32 +693,49 @@ func (bigtable *Bigtable) SaveAttestations(blocks map[uint64]map[string]*types.B
...
+					attestationsBySlot[slot] = make(map[uint64]uint64)
+				}
+				for _, validator := range b.EmptySlotAttestationAssignments {
+					attestationsBySlot[slot][validator] = MAX_BLOCK_NUMBER // will result in the ts of the set bigtable cell being 0
+				}
+			} else {
+				// logger.Infof("processing slot %v", slot)
+				for _, a := range b.Attestations {
+					for _, validator := range a.Attesters {
+						inclusionSlot := slot
```
manuelsc, line 707: this part could be moved to the outer b.Attestations loop

db/bigtable.go
```diff
@@ -595,19 +745,40 @@ func (bigtable *Bigtable) SaveAttestations(blocks map[uint64]map[string]*types.B
 		}
 		bigtable.lastAttestationCacheMux.Unlock()
 
-		err := bigtable.tableBeaconchain.Apply(ctx, fmt.Sprintf("%s:e:%s:s:%s", bigtable.chainId, reversedPaddedEpoch(attestedSlot/utils.Config.Chain.Config.SlotsPerEpoch), reversedPaddedSlot(attestedSlot)), mutInclusionSlot)
-		if err != nil {
-			return err
-		}
+		attstart := time.Now()
 
+		if len(mutsInclusionSlot) == MAX_BATCH_MUTATIONS {
```
manuelsc, line 750: same issue as above, mutsInclusionSlot can get bigger than MAX_BATCH_MUTATIONS

db/bigtable.go
```diff
@@ -680,7 +863,13 @@ func (bigtable *Bigtable) SaveSyncComitteeDuties(blocks map[uint64]map[string]*t
 	for _, slot := range slots {
 		for _, b := range blocks[slot] {
 			if b.Status == 2 {
```
manuelsc, line 865: please constant here too

db/bigtable.go
```diff
@@ -701,152 +890,170 @@ func (bigtable *Bigtable) SaveSyncComitteeDuties(blocks map[uint64]map[string]*t
...
+	keys := make([]string, 0, utils.Config.Chain.Config.SlotsPerEpoch*utils.Config.Chain.Config.SyncCommitteeSize+1)
+
 	for slot, validators := range dutiesBySlot {
-		mut := gcp_bigtable.NewMutation()
+		participation := uint64(0)
 		for validator, participated := range validators {
+			mut := gcp_bigtable.NewMutation()
 			if participated {
-				mut.Set(SYNC_COMMITTEES_FAMILY, fmt.Sprintf("%d", validator), gcp_bigtable.Timestamp((max_block_number-slot)*1000), []byte{})
+				mut.Set(SYNC_COMMITTEES_FAMILY, "s", gcp_bigtable.Timestamp((MAX_BLOCK_NUMBER-slot)*1000), []byte{})
```
manuelsc, line 902: While with current mainnet config of 1 slot every 12 seconds this is save for a couple hundred years, MAX_BLOCK_NUMBER-slot could be potentially negative for a chain with lower slot times (potential L2 maybe).

db/bigtable.go
peterbitfly: remote hosts can be specified using the new `EmulatorHost` field in the config.

db/bigtable.go
```diff
@@ -1813,103 +2097,118 @@ func (bigtable *Bigtable) GetAggregatedValidatorIncomeDetailsHistory(validators
...
 	if endEpoch < startEpoch { // handle overflows
 		startEpoch = 0
 	}
 
-	ranges := gcp_bigtable.RowRangeList{}
-	if startEpoch == 0 { // special case when the 0 epoch is included
-		rangeEnd := fmt.Sprintf("%s:e:%s:s:%s", bigtable.chainId, reversedPaddedEpoch(0), ":")
-		rangeStart := fmt.Sprintf("%s:e:%s:s:", bigtable.chainId, reversedPaddedEpoch(0))
-		ranges = append(ranges, gcp_bigtable.NewRange(rangeStart, rangeEnd))
+	ranges := make(gcp_bigtable.RowRangeList, 0, int((endEpoch-startEpoch))*len(validatorIndices))
```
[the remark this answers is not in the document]
manuelsc: To add to that, any startEpoch = endEpoch will result in the allocated cap to be 0. While this code does works, the lower cap results in an append being amortized

db/bigtable.go
```diff
@@ -1542,279 +1831,400 @@ func (bigtable *Bigtable) SaveValidatorIncomeDetails(epoch uint64, rewards map[u
 	return nil
 }
 
-func (bigtable *Bigtable) GetEpochIncomeHistoryDescending(startEpoch uint64, endEpoch uint64) (*itypes.ValidatorEpochIncome, error) {
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Second*30))
-	defer cancel()
+func (bigtable *Bigtable) MigrateEpochSchemaV1ToV2(epoch uint64) error {
```
D13ce, line 1834: **Question:** This function contains some error logs that do not stop the function. Personally I think a zero error policy should be applied. Here is an example:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/ef0180fc0f0250c8c83a06986f59a5affa18e5c3/db/bigtable.go#L1874

db/bigtable.go
```diff
@@ -1542,279 +1831,400 @@ func (bigtable *Bigtable) SaveValidatorIncomeDetails(epoch uint64, rewards map[u
...
 		if err != nil {
-			logger.Errorf("error decoding income data for row %v: %v", r.Key(), err)
+			logger.Errorf("error parsing epoch from row key %v: %v", r.Key(), err)
 			return false
 		}
+
+		rowKeyEpoch = MAX_EPOCH - rowKeyEpoch
+
+		if epoch != rowKeyEpoch {
+			logger.Errorf("retrieved different epoch than requested, requested: %d, retrieved: %d", epoch, rowKeyEpoch)
```
D13ce, line 1874: **Suggestion:** Since this function will not be executed periodically, I suggest to add AS MUCH information as possible to every single log entry. "Spam" is not an issue if this is only done once.

Here, for example, why not add the complete row key too?

Also, maybe keep logs like https://github.com/gobitfly/eth2-beaconchain-explorer/blob/ef0180fc0f0250c8c83a06986f59a5affa18e5c3/db/bigtable.go#L1861?

db/bigtable.go
```diff
@@ -1542,279 +1831,400 @@ func (bigtable *Bigtable) SaveValidatorIncomeDetails(epoch uint64, rewards map[u
...
+	err = bigtable.SaveValidatorBalances(epoch, validators)
+	if err != nil {
+		return err
+	}
+	// logger.Infof("migrated balance data in %v", time.Since(start))
+	// start = time.Now()
 
-	ranges := bigtable.getEpochRanges(startEpoch, endEpoch)
-	res := make(map[uint64]map[uint64]*itypes.ValidatorEpochIncome, len(validators))
+	i := 0
```
D13ce, line 2035: **Nitpick:** This should be removed.

db/bigtable.go
```diff
@@ -861,83 +1068,89 @@ func (bigtable *Bigtable) GetValidatorAttestationHistory(validators []uint64, st
...
+					attesterSlotString := strings.Replace(ri.Column, ATTESTATIONS_FAMILY+":", "", 1)
+					attesterSlot, err := strconv.ParseUint(attesterSlotString, 10, 64)
+					if err != nil {
+						logger.Errorf("error parsing slot from row key %v: %v", r.Key(), err)
+						return false
+					}
+					inclusionSlot := MAX_BLOCK_NUMBER - uint64(ri.Timestamp)/1000
+
+					status := uint64(1)
+					if inclusionSlot == MAX_BLOCK_NUMBER {
```
manuelsc, line 1113: I don't think this check is necessary. We save ri.Timestamp with `(MAX_BLOCK_NUMBER-inclusionSlot)*1000` so (lets assume MAX_BLOCK_NUMBER 99) 99-0=99. *1000 = 99000
And convert back to inclusionSlot by `MAX_BLOCK_NUMBER - uint64(ri.Timestamp)/1000`. So 99 - (99000)/1000 = 99 - 99 = 0.
This case only handles when we store inclusionSlot `MAX_BLOCK_NUMBER` but then we shouldn't manually correct the inclusion slot to 0.

db/bigtable.go
D13ce, line 1834: **Question:** `SYNC_COMMITTEES_PARTICIPATION_FAMILY` is not touched by the migration file. Is that intended?

D13ce (changes requested):
Everything related to bigtable is SO MUCH cleaner, really delightful! :tada:

There are some comments from my previous review (https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2490#pullrequestreview-1602019082) you have not addressed though.
Please navigate to https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2490 and press
image
Scroll through the conversations and also press
image
to see all of them.

I have tried to resolve as much as possible so only the ones that are unresolved should matter.

---

This review adds a lot of comments as well. Please don't forget to click "Load more" here too.

db/bigtable.go
```diff
@@ -861,83 +1068,89 @@ func (bigtable *Bigtable) GetValidatorAttestationHistory(validators []uint64, st
...
+					if res[validator] == nil {
+						res[validator] = make([]*types.ValidatorAttestation, 0)
+					}
+
+					if len(res[validator]) > 0 && res[validator][len(res[validator])-1].AttesterSlot == attesterSlot {
+						// don't override successful attestion, that was included in a different slot
+						if status == 1 || res[validator][len(res[validator])-1].Status != 1 {
+							res[validator][len(res[validator])-1].InclusionSlot = inclusionSlot
+							res[validator][len(res[validator])-1].Status = status
+							res[validator][len(res[validator])-1].Delay = int64(inclusionSlot - attesterSlot)
```
manuelsc, line 1130: question: wouldn't this have to be `inclusionSlot - attesterSlot - 1` as well?

db/bigtable.go
```diff
@@ -1016,157 +1287,172 @@ func (bigtable *Bigtable) GetValidatorMissedAttestationHistory(validators []uint
...
+					attesterSlot, err := strconv.ParseUint(attesterSlotString, 10, 64)
+					if err != nil {
+						logger.Errorf("error parsing slot from row key %v: %v", r.Key(), err)
+						return false
+					}
+
+					inclusionSlot := MAX_BLOCK_NUMBER - uint64(ri.Timestamp)/1000
+
+					status := uint64(1)
+					if inclusionSlot == MAX_BLOCK_NUMBER {
```
manuelsc, line 1335: not sure if this is necessary. Same as with the other occurence of this, inclusionSlot should be 0 if you want to handle the 0 case. As for the MAX_BLOCK_NUMBER slot case, not sure if setting the status for this future slot is necessary

db/bigtable.go
peterbitfly: will be done in a separate PR to not further bloat this one

db/bigtable.go
peterbitfly: even with chains with a block time of one second this should last around 30 years

db/bigtable.go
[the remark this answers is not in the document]
peterbitfly: I've added a `+1` to ensure sufficient space is allocated

db/bigtable.go
```diff
@@ -388,46 +417,83 @@ func machineMetricRowParts(r string) (bool, uint64, string, string) {
...
+			return err
 		}
 	}
-	err := bigtable.tableBeaconchain.Apply(ctx, fmt.Sprintf("%s:e:b:%s", bigtable.chainId, reversedPaddedEpoch(epoch)), mut)
 
+	key := fmt.Sprintf("%s:%s:%s", bigtable.chainId, VALIDATOR_HIGHEST_ACTIVE_INDEX_FAMILY, epochKey)
+	err := bigtable.tableValidatorBalances.Apply(ctx, key, mut)
 	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
peterbitfly: I've rearranged the code and added a comment

db/bigtable.go
```diff
@@ -472,14 +566,27 @@ func (bigtable *Bigtable) SaveProposalAssignments(epoch uint64, assignments map[
...
 	ts := gcp_bigtable.Timestamp(0)
 
+	muts := make([]*gcp_bigtable.Mutation, 0, len(assignments))
+	keys := make([]string, 0, len(assignments))
+
 	for slot, validator := range assignments {
 		mut := gcp_bigtable.NewMutation()
-		mut.Set(PROPOSALS_FAMILY, fmt.Sprintf("%d", validator), ts, []byte{})
-		err := bigtable.tableBeaconchain.Apply(ctx, fmt.Sprintf("%s:e:%s:s:%s", bigtable.chainId, reversedPaddedEpoch(epoch), reversedPaddedSlot(slot)), mut)
+		mut.Set(PROPOSALS_FAMILY, "p", ts, []byte{})
```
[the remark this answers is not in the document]
peterbitfly: not sure how the google bigtable client will handle that, also `ApplyBulk` requires an array of mutations and keys with the same size.

cmd/misc/main.go
```diff
@@ -539,7 +546,7 @@ func indexMissingBlocks(start uint64, end uint64, bt *db.Bigtable, client *rpc.E
 					logrus.Infof("block [%v] not found so we need to index it", j)
 					if _, err := db.BigtableClient.GetBlockFromBlocksTable(j); err != nil {
 						logrus.Infof("could not load [%v] from blocks table so we need to fetch it from the node and save it", j)
-						bc, _, err := client.GetBlock(int64(j))
+						bc, _, err := client.GetBlock(int64(j), "parity/geth")
```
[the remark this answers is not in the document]
peterbitfly: can be done in a separate pr

cmd/rewards-exporter/main.go
```diff
@@ -66,6 +66,24 @@ func main() {
 	defer db.ReaderDb.Close()
 	defer db.WriterDb.Close()
 
+	if bnAddress == nil || *bnAddress == "" {
+		if utils.Config.Indexer.Node.Host == "" {
```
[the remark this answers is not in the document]
peterbitfly: can be done in a separate pr

handlers/api.go
```diff
@@ -1177,9 +1177,12 @@ func getSyncCommitteeSlotsStatistics(validators []uint64, epoch uint64) (types.S
 
 	// validator_stats is updated only once a day, everything missing has to be collected from bigtable (which is slower than validator_stats)
 	// check when the last update to validator_stats was
-	lastExportedDay := services.LatestExportedStatisticDay()
 	epochsPerDay := utils.EpochsPerDay()
-	lastExportedEpoch := ((lastExportedDay + 1) * epochsPerDay) - 1
+	lastExportedDay, err := services.LatestExportedStatisticDay()
+	lastExportedEpoch := uint64(0)
+	if err == nil {
```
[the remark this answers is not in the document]
peterbitfly: should be fixed in a separate pr

handlers/common.go
```diff
@@ -99,18 +104,25 @@ func GetValidatorEarnings(validators []uint64, currency string) (*types.Validato
...
+			err := db.GetValidatorBalanceForDay(validators, lastStatsDay, &lastBalance)
+			if err != nil {
+				return err
+			}
+		}
+		err := db.GetValidatorDepositsForSlots(validators, firstSlot, lastSlot, &lastDeposits)
+		if err != nil {
+			return err
+		}
+		return db.GetValidatorWithdrawalsForSlots(validators, firstSlot, lastSlot, &lastWithdrawals)
```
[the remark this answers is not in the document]
peterbitfly: can be done in a separate pr

handlers/eth1Block.go
```diff
@@ -119,7 +119,7 @@ func Eth1Block(w http.ResponseWriter, r *http.Request) {
 func GetExecutionBlockPageData(number uint64, limit int) (*types.Eth1BlockPageData, error) {
 	block, err := db.BigtableClient.GetBlockFromBlocksTable(number)
 	if diffToHead := int64(services.LatestEth1BlockNumber()) - int64(number); err != nil && diffToHead < 0 && diffToHead >= -5 {
-		block, _, err = rpc.CurrentErigonClient.GetBlock(int64(number))
+		block, _, err = rpc.CurrentErigonClient.GetBlock(int64(number), "parity/geth")
```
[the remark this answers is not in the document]
peterbitfly: there are only two trace types available right now (parity type traces and geth type traces). I'd like to move everything away from flags, so that change can be done in this scope of that refactor.

handlers/validator.go
```diff
@@ -1588,6 +1594,10 @@ func ValidatorHistory(w http.ResponseWriter, r *http.Request) {
 
 		// if there are additional epochs with duties we have to go through all of them as there can be gaps (after the exit before the duty)
 		for i := endEpoch; i >= startEpoch; i-- {
+			if i > endEpoch {
+				break
+			}
```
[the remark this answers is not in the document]
peterbitfly: this guards against underflows when start epoch is 0

db/bigtable.go
```diff
@@ -1000,8 +1213,66 @@ func (bigtable *Bigtable) GetLastAttestationSlots(validators []uint64) (map[uint
 	return res, nil
 }
 
+func (bigtable *Bigtable) GetSyncParticipationBySlot(slot uint64) (uint64, error) {
+
+	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(time.Minute*5))
+	defer cancel()
+
+	key := fmt.Sprintf("%s:%s:%s", bigtable.chainId, SYNC_COMMITTEES_PARTICIPATION_FAMILY, bigtable.reversedPaddedSlot(slot))
```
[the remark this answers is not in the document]
peterbitfly: for simplicity I would not like to duplicate the values into a second variable, we should keep in mind to not change family names

handlers/validator.go
```diff
@@ -1999,24 +1968,47 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
...
+			if syncDuties[validatorIndex] == nil {
+				tableData = append(tableData,
+					[]interface{}{
+						fmt.Sprintf("%d", utils.SyncPeriodOfEpoch(epoch)),
+						utils.FormatEpoch(epoch),
+						utils.FormatBlockSlot(slot),
+						utils.FormatSyncParticipationStatus(0, slot),
+						utils.FormatSyncParticipations(participation),
+					},
+				)
```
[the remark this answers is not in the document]
peterbitfly: yes

db/bigtable.go
```diff
@@ -1542,279 +1831,400 @@ func (bigtable *Bigtable) SaveValidatorIncomeDetails(epoch uint64, rewards map[u
...
+		EffectiveBalance         uint64
+		Balance                  uint64
+		IncomeDetails            *itypes.ValidatorEpochIncome
+	}
+
+	epochData := make(map[uint64]*validatorEpochData)
+	filter := gcp_bigtable.LatestNFilter(1)
+	ctx := context.Background()
+
+	prefixEpochRange := gcp_bigtable.PrefixRange(fmt.Sprintf("%s:e:b:%s", bigtable.chainId, bigtable.reversedPaddedEpoch(epoch)))
```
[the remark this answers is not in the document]
peterbitfly: we will likely perform the migration using data sourced from the nodes instead of bigtable, so that methode will be obsolete.

db/bigtable.go
peterbitfly: yes, it is currently not implemented

db/bigtable.go
```diff
@@ -404,24 +416,41 @@ func (bigtable *Bigtable) SaveValidatorBalances(epoch uint64, validators []*type
 		binary.LittleEndian.PutUint64(effectiveBalanceEncoded, validator.EffectiveBalance)
 
 		combined := append(balanceEncoded, effectiveBalanceEncoded...)
-		mut.Set(VALIDATOR_BALANCES_FAMILY, fmt.Sprintf("%d", validator.Index), ts, combined)
+		mut := &gcp_bigtable.Mutation{}
+		mut.Set(VALIDATOR_BALANCES_FAMILY, "b", ts, combined)
+		key := fmt.Sprintf("%s:%s:%s", bigtable.chainId, bigtable.validatorIndexToKey(validator.Index), reversedPaddedEpoch(epoch))
```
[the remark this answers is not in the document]
D13ce: Will be part of an upcoming PR.

db/bigtable.go
```diff
@@ -1283,221 +1456,212 @@ func (bigtable *Bigtable) GetValidatorEffectiveness(validators []uint64, epoch u
...
+		logrus.Infof("retrieving validator balance stats for validators %v - %v", vals[0], vals[len(vals)-1])
 
-					if epoch == endEpoch {
-						resultContainer.res[validator].EndBalance = balance
-						resultContainer.res[validator].EndEffectiveBalance = effectiveBalance
+		res, err := bigtable.GetValidatorBalanceHistory(vals, startEpoch, endEpoch)
+		if err != nil {
+			return nil, err
+		}
+		resultContainer.mu.Lock()
```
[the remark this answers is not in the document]
D13ce: Will be done in a separate PR.

db/db.go
```diff
@@ -1079,7 +1027,13 @@ func saveValidators(data *types.EpochData, tx *sqlx.Tx, client rpc.Client) error
 
 	if data.Epoch == 0 {
 		var err error
-		genesisBalances, err = BigtableClient.GetValidatorBalanceHistory([]uint64{}, 0, 0)
+
+		indices := make([]uint64, 0, len(data.Validators))
```
manuelsc, line 1031: code style: I have seen this construct a few times now. This might be is a good fit for a new utily function to reduce boilerplate code

db/bigtable.go
peterbitfly: this was a leftover comment from a debugging session, can be deleted

db/bigtable.go
```diff
@@ -861,83 +1068,89 @@ func (bigtable *Bigtable) GetValidatorAttestationHistory(validators []uint64, st
...
+						}
+					} else {
+						res[validator] = append(res[validator], &types.ValidatorAttestation{
+							Index:          validator,
+							Epoch:          utils.EpochOfSlot(attesterSlot),
+							AttesterSlot:   attesterSlot,
+							CommitteeIndex: 0,
+							Status:         status,
+							InclusionSlot:  inclusionSlot,
+							Delay:          int64(inclusionSlot) - int64(attesterSlot) - 1,
```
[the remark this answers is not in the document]
peterbitfly: done

exporter/sync_committees_count.go
```diff
@@ -26,8 +27,7 @@ func exportSyncCommitteesCount() error {
 		return err
 	}
 
-	currEpoch := utils.TimeToEpoch(time.Now())
-	currentPeriod := utils.SyncPeriodOfEpoch(uint64(currEpoch))
+	currentPeriod := utils.SyncPeriodOfEpoch(services.LatestFinalizedEpoch())
```
manuelsc, line 30: Not really a fan of the exporter depending on data from the frontend service. Maybe latestFinalizedEpoch data should be exported by the exporter instead?

peterbitfly: change to `db.GetLatestFinalizedEpoch()`

db/db.go
peterbitfly: should be done in a separate PR

db/bigtable.go
```diff
@@ -642,324 +722,369 @@ func (bigtable *Bigtable) SaveProposals(blocks map[uint64]map[string]*types.Bloc
...
+		})
+	}
+
+	if err := g.Wait(); err != nil {
 		return nil, err
 	}
 
 	// Find all missed and orphaned slots
 	slots := []uint64{}
+	maxSlot := (endEpoch + 1) * utils.Config.Chain.Config.SlotsPerEpoch
```
D13ce, line 1057: **Question:** Shouldn't this be `maxSlot := ((endEpoch + 1) * utils.Config.Chain.Config.SlotsPerEpoch) - 1` as below we use `slot <= maxSlot`?

db/bigtable.go
```diff
@@ -1599,386 +1850,243 @@ func (bigtable *Bigtable) SaveValidatorIncomeDetails(epoch uint64, rewards map[u
...
-		keys = append(keys, fmt.Sprintf("%s:e:%s:s:%s", bigtable.chainId, reversedPaddedEpoch(slot/utils.Config.Chain.Config.SlotsPerEpoch), reversedPaddedSlot(slot)))
+	if endEpoch < startEpoch { // handle overflows
+		startEpoch = 0
 	}
-	keys = append(keys, fmt.Sprintf("%s:e:b:%s", bigtable.chainId, reversedPaddedEpoch(epoch)))
 
-	// for _, k := range keys {
-	// 	logger.Info(k)
-	// }
+	ranges := make(gcp_bigtable.RowRangeList, 0, int((endEpoch-startEpoch))*len(validatorIndices)+1)
```
D13ce, line 2021: **Minor:** Should be
```
ranges := make(gcp_bigtable.RowRangeList, 0, int(endEpoch-startEpoch+1)*len(validatorIndices))
```
Also applies to https://github.com/gobitfly/eth2-beaconchain-explorer/blob/150bfa89b67c1dab30277f30eff6199932542075/db/bigtable.go#L2041

db/bigtable.go
```diff
@@ -1599,386 +1850,243 @@ func (bigtable *Bigtable) SaveValidatorIncomeDetails(epoch uint64, rewards map[u
...
-			rangeEnd = fmt.Sprintf("%s:e:b:%s%s", bigtable.chainId, reversedPaddedEpoch(startEpoch+1), "\x00")
-			rangeStart = fmt.Sprintf("%s:e:b:%s", bigtable.chainId, reversedPaddedEpoch(endEpoch))
-			ranges = append(ranges, gcp_bigtable.NewRange(rangeStart, rangeEnd))
-		}
-	} else {
-		// epochs are sorted descending, so start with the largest epoch and end with the smallest
-		// add \x00 to make the range inclusive
-		rangeEnd := fmt.Sprintf("%s:e:b:%s%s", bigtable.chainId, reversedPaddedEpoch(startEpoch), "\x00")
-		rangeStart := fmt.Sprintf("%s:e:b:%s", bigtable.chainId, reversedPaddedEpoch(endEpoch))
+	ranges := make(gcp_bigtable.RowRangeList, 0, len(validatorIndices)+1)
```
D13ce, line 2063: **Question:** Why the `+1`?

cmd/migrations/bigtable/main.go
```diff
@@ -59,80 +58,12 @@ func main() {
 		i := i
```
D13ce, line 58: **Minor:** This is not needed anymore.

D13ce:
The Postgres table `chain_head` is not used anymore and should be removed with a migration file.

db/db.go
```diff
@@ -737,67 +743,6 @@ func SaveBlock(block *types.Block, forceSlotUpdate bool) error {
 	return nil
 }
 
-func UpdateChainHead(head *types.ChainHead) error {
```
D13ce, line 740: The Postgres table `chain_head` is not used anymore and should be removed with a migration file.

D13ce (approved):
utACK :tada:

PR #2492, opened 2023-08-21
(BIDS-2361) make the ValidatorSync function prettier

MauserBitfly:
### <samp>🤖 Generated by Copilot at 851a319</samp>

Refactor `ValidatorSync` handler to use maps, contexts, and errgroups for sync committee duties, and remove unused function `GetValidatorSyncDutiesHistoryOrdered` from `db` package.

### DIFF
--- a/handlers/validator.go
+++ b/handlers/validator.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"bytes"
+	"context"
 	"database/sql"
 	"encoding/hex"
 	"encoding/json"
@@ -18,13 +19,13 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/lib/pq"
 	protomath "github.com/protolambda/zrnt/eth2/util/math"
-	"golang.org/x/exp/slices"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/gorilla/csrf"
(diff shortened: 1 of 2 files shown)


handlers/validator.go
```diff
@@ -1882,190 +1879,117 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
...
+							mux.Unlock()
+							continue // this is not the slot we are looking for
+						}
+
+						// We want to know how many validators successfully participated in this duty
+						if duty.Status == 1 {
+							pageSlot.Participation++
+						}
+
+						// Get the status if the dutiy belongs to our validator
```
Eisei24, line 1966: **typo:** duty

handlers/validator.go
```diff
@@ -1882,190 +1879,117 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
...
+		if descOrdering {
+			utils.ReverseSlice(syncSlots)
+		}
+
+		// let's pick the slots we want to display on the requested page
+		pageSlots := syncSlots[start:protomath.MinU64(start+length, totalCount)]
+		length = uint64(len(pageSlots)) // Last page might be shorter
+		epochs := []uint64{}
+		last := uint64(0)
+		// we gether all epochs for that page (the table currently displays 10 items per page, so it displays max 2 epoches. But we want to be future proof, so we support bigger page sizes)
```
Eisei24, line 1913: **typo:** gather

Eisei24 (approved):
lgtm

This function now looks much more readable.
One potential issue is that we now call `GetValidatorSyncDutiesHistory` (which checks the bigtable) once for each epoch that is displayed on the current page as opposed to before where we had an epoch range as long as we were in the same sync committee period.

As a result these pages (every 4th page) take a bit longer to load.
@recy21 Please consider this when doing your check.

handlers/validator.go
MauserBitfly: fixed

handlers/validator.go
MauserBitfly: fixed

MauserBitfly:
> lgtm
> 
> This function now looks much more readable. One potential issue is that we now call `GetValidatorSyncDutiesHistory` (which checks the bigtable) once for each epoch that is displayed on the current page as opposed to before where we had an epoch range as long as we were in the same sync committee period.
> 
> As a result these pages (every 4th page) take a bit longer to load. @recy21 Please consider this when doing your check.

Thanks. Yes every 4th page is a little slower than the other 3 (because it needs to load 2 epochs), but still faster then the old way ;-).

peterbitfly (approved):
LGTM DB wise, did not check the business logic

PR #2494, opened 2023-08-22
(BIDS-2400) don't switch tab on new url

MauserBitfly:
### <samp>🤖 Generated by Copilot at 7847e8e</samp>

Prevent tabbar switcher from activating on anchor links. Fix a bug in `layout.js` where the wrong tab would be shown when navigating within the same page using hashes.


D13ce (approved):
lgtm

PR #2496, opened 2023-08-22
(BIDS-2403) make it lower to get higher

MauserBitfly:
### <samp>🤖 Generated by Copilot at 45910a9</samp>

Fixed a case-sensitive search bug for eth1 addresses in `handlers/search.go`. Lowercased the search parameter before calling `FindValidatorIndicesByEth1Address`.


D13ce (approved):
lgtm

PR #2501, opened 2023-08-23
(BIDS-2381) change latest finalized epoch handling

MauserBitfly:
### <samp>🤖 Generated by Copilot at f781ea0</samp>

This pull request simplifies and improves the way the finalization status of epochs is handled in the eth2-beaconchain-explorer project. It removes the `finalized` column from the `epochs` table and replaces it with a query to the new `chain_head` table, which stores the latest chain head information from the beacon node. It updates various packages and files, such as `cmd`, `db`, `eth1data`, `exporter`, `handlers`, and `services`, to use the new table and the latest finalized epoch to query and display the epoch data. It also adds some validation and performance improvements to the rewards exporter and the eth1 data retrieval. It includes a new migration file to create and drop the `chain_head` table.


db/statistics.go
```diff
@@ -1868,16 +1868,16 @@ func WriteGraffitiStatisticsForDay(day int64) error {
...
+	latestFinalizedEpoch, err := GetLatestFinalizedEpoch()
 	if err != nil {
-		return err
+		return fmt.Errorf("error getting latest finalized epoch from db %w", err)
 	}
 
-	if finalizedCount < epochsPerDay {
-		return fmt.Errorf("delaying export as not all epochs for day %v finalized. %v of %v", day, finalizedCount, epochsPerDay)
+	if lastEpoch > latestFinalizedEpoch {
+		return fmt.Errorf("delaying chart series export as not all epochs for day %v finalized. last epoch of the day [%v] last finalized epoch [%v]", day, lastEpoch, latestFinalizedEpoch)
```
Eisei24, line 1879: **minor:** A copy&paste error message? This has nothing to do with the chart series.

handlers/slot.go
```diff
@@ -205,14 +206,15 @@ func getAttestationsData(slot uint64, onlyFirst bool) ([]*types.BlockPageAttesta
...
+	latestFinalizedEpoch := services.LatestFinalizedEpoch()
 	blockPageData := types.BlockPageData{}
 	blockPageData.Mainnet = utils.Config.Chain.Config.ConfigName == "mainnet"
 	// for the first slot in an epoch the previous epoch defines the finalized state
 	err := db.ReaderDb.Get(&blockPageData, `
 		SELECT
 			blocks.epoch,
-			COALESCE(epochs.finalized, false) AS epoch_finalized,
-			COALESCE(prev_epoch.finalized, false) AS prev_epoch_finalized,
+			(COALESCE(epochs.epoch, 0) <= $3) AS epoch_finalized,
```
Eisei24, line 216: **issue:** The **COALESCE** behaviour would change from **false** => **true** if the value is not found.

handlers/slot.go
```diff
@@ -205,14 +206,15 @@ func getAttestationsData(slot uint64, onlyFirst bool) ([]*types.BlockPageAttesta
...
+			(GREATEST((blocks.slot-1)/$2-1,0) <= $3) AS prev_epoch_finalized,
```
Eisei24, line 217: **issue:** Also the case here I think. Removing the left join removes the possibility that it could return no rows which before would result in a **false** for the **prev_epoch_finalized**.
Now the behaviour could be different.

Please think this through again.
If you say that there are no problems it's fine with me.

Eisei24 (commented):
Please have a look at my comments and fix potential issues.

handlers/slot.go
MauserBitfly: Thought it through ... the COALESCE is just a precaution and should not trigger. Only if we have some bigger problem in the data - and then this is the least of the problems.

handlers/slot.go
MauserBitfly: Like the other comment + it's more performant this way.

Eisei24 (approved):
utACK

db/db.go
```diff
@@ -734,6 +733,67 @@ func SaveBlock(block *types.Block) error {
 	return nil
 }
 
+func UpdateChainHead(head *types.ChainHead) error {
+	count := 0
+	err := ReaderDb.Get(&count, "select count(*) from chain_head")
```
recy21, line 738: Please use upper case for SQL stuff, like SELECT, COUNT and FROM. Thanks.

db/db.go
```diff
@@ -512,16 +512,15 @@ func GetAllEpochs() ([]uint64, error) {
...
-	err := WriterDb.Get(&count, "SELECT COUNT(*) FROM epochs WHERE epoch >= $1 AND epoch <= $2 AND finalized", startEpoch, endEpoch)
-
-	if err != nil {
-		return 0, fmt.Errorf("error counting finalized epochs [%v -> %v] from DB: %w", startEpoch, endEpoch, err)
+// Get latest finalized epoch
+func GetLatestFinalizedEpoch() (uint64, error) {
+	var latestFinalized uint64
+	err := WriterDb.Get(&latestFinalized, "SELECT finalized_epoch FROM chain_head")
+	if err != nil && err != sql.ErrNoRows {
+		logger.Errorf("error retrieving latest exported finalized epoch from the database: %v", err)
```
recy21, line 520: One should always use utils.LogError

eth1data/eth1data.go
```diff
@@ -34,9 +35,14 @@ func GetEth1Transaction(hash common.Hash) (*types.Eth1TxData, error) {
...
-			err := db.ReaderDb.Get(&data.Epoch,
-				`select epochs.finalized, epochs.globalparticipationrate from blocks left join epochs on blocks.epoch = epochs.epoch where blocks.exec_block_number = $1 and blocks.status='1';`,
-				data.BlockNumber)
+			latestFinalizedEpoch := services.LatestFinalizedEpoch()
+			err := db.ReaderDb.Get(&data.Epoch, `
+				SELECT (epochs.epoch <= $2) AS finalized, epochs.globalparticipationrate 
+				FROM blocks 
+				LEFT JOIN epochs ON blocks.epoch = epochs.epoch 
+				WHERE blocks.exec_block_number = $1 
+				AND blocks.status='1';`,
```
recy21, line 44: It's strange that we never use ';' at the end when in Golang, but I got used to it.
Anyway, you used here ';', which should be removed, in sake of consistency.

eth1data/eth1data.go
```diff
@@ -228,9 +234,18 @@ func GetEth1Transaction(hash common.Hash) (*types.Eth1TxData, error) {
 	}
 
 	if txPageData.BlockNumber != 0 {
-		err := db.ReaderDb.Get(&txPageData.Epoch,
-			`select epochs.finalized, epochs.globalparticipationrate from blocks left join epochs on blocks.epoch = epochs.epoch where blocks.exec_block_number = $1 and blocks.status='1';`,
-			&txPageData.BlockNumber)
+		latestFinalizedEpoch := services.LatestFinalizedEpoch()
+		err := db.ReaderDb.Get(&txPageData.Epoch, `
```
recy21, line 238: This looks similar to SELECT a few lines above, but is different formatted. Why?

recy21 (changes requested):

eth1data/eth1data.go
MauserBitfly: Ah yes, moved the query to the db.go and and call it now from both places.

recy21 (approved):

PR #2511, opened 2023-08-29
(BIDS-2378) Serve validator rank top% via app dashboard api

D13ce:
With this PR, the app's dashboard API now also serves the validator "Top X%" rank for every queried validator.

### DIFF
--- a/handlers/api.go
+++ b/handlers/api.go
@@ -1328,6 +1328,7 @@ func getRocketpoolValidators(queryIndices []uint64) ([]interface{}, error) {
 }
 
 func validators(queryIndices []uint64) ([]interface{}, error) {
+	// we use MAX(validatorindex)+1 instead of COUNT(*) for querying the rank_count for performance-reasons
 	rows, err := db.ReaderDb.Query(`
 	SELECT 
 		validators.validatorindex,
@@ -1345,7 +1346,8 @@ func validators(queryIndices []uint64) ([]interface{}, error) {
 		COALESCE(validator_performance.cl_performance_31d, 0) AS performance31d,
 		COALESCE(validator_performance.cl_performance_365d, 0) AS performance365d,
 		COALESCE(validator_performance.cl_performance_total, 0) AS performanceTotal,
-		rank7d,
+		COALESCE(validator_performance.rank7d, 0) AS rank7d,
+		COALESCE(validator_performance_count.total_count, 0) AS rankcount,
 		w.total as total_withdrawals
 	FROM validators
 	LEFT JOIN validator_performance ON validators.validatorindex = validator_performance.validatorindex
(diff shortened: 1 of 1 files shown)


handlers/api.go
```diff
@@ -1357,6 +1359,7 @@ func validators(queryIndices []uint64) ([]interface{}, error) {
 		WHERE validatorindex = ANY($1)
 		GROUP BY validatorindex
 	) as w ON w.index = validators.validatorindex
+	LEFT JOIN (SELECT MAX(validatorindex)+1 FROM validator_performance WHERE validatorindex >= 0 AND validatorindex < 2147483647) validator_performance_count(total_count) ON true
```
LuccaBitfly, line 1362: Doing this with `LEFT JOIN ... ON TRUE` is kinda weird, you can prob do this with a normal subquery.

handlers/api.go
```diff
@@ -1413,6 +1449,17 @@ func validators(queryIndices []uint64) ([]interface{}, error) {
...
 		}
+
+		if rankCount > 0 {
+			rank7d, ok := eMap["rank7d"].(int64)
+			if !ok {
+				logger.Errorf("error converting rank7d to int64")
+				continue
+			}
+			if rank7d > 0 {
+				eMap["rankPercentage"] = (float64(rank7d) / float64(rankCount)) * 100
```
LuccaBitfly, line 1460: Consider calculating the rank percentage in the query, which would absolve you from the need to have the same value in every row.

LuccaBitfly (changes requested):
Please have a look at your issues.

handlers/api.go
D13ce: Good idea, this will increase readability without costing performance. Will do!

handlers/api.go
D13ce: Done that, love that!

handlers/api.go
```diff
@@ -1369,20 +1372,53 @@ func validators(queryIndices []uint64) ([]interface{}, error) {
 		return nil, fmt.Errorf("error converting validators to json: %w", err)
 	}
 
-	balances, err := db.BigtableClient.GetValidatorBalanceHistory(queryIndices, services.LatestEpoch(), services.LatestEpoch())
-	if err != nil {
-		return nil, fmt.Errorf("error getting validator balances from bigtable: %w", err)
+	// get rankCount once so it can be reused later (part of query above to reduce db connections)
+	rankCount := int64(0)
+	if len(data) > 0 {
+		eMap, ok := data[0].(map[string]interface{})
```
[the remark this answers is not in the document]
D13ce: I copy&pasted the name from here: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/f8a19c79741acda33d997b231e1abc22ee1d2fbf/handlers/api.go#L1414

Here it makes sense as it stands for "entry map". In my case it did not but the code part has been refactored anyways.

handlers/api.go
```diff
@@ -1328,6 +1328,7 @@ func getRocketpoolValidators(queryIndices []uint64) ([]interface{}, error) {
 }
 
 func validators(queryIndices []uint64) ([]interface{}, error) {
```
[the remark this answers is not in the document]
LuccaBitfly: Did you mean to have "infor" in the function name?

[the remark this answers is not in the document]
D13ce: No, fixed.

LuccaBitfly (approved):
lgtm

PR #2512, opened 2023-08-29
(BIDS-2325) input length dependent search delay

remoterami:
Added an input length dependent time delay until searches get executed. Only applied to 4 main search bars which execute parallel requests.

1 character start search after 3.5s
2 characters start search after 3s
...
5 or more characters start search after 1.5s 

### <samp>🤖 Generated by Copilot at 23152e4</samp>

This pull request improves the search features in various pages of the eth2-beaconchain-explorer by adding a `debounce` function that limits and delays the number of requests sent to the server. The `debounce` function is applied to the `transport._get` functions of several Bloodhound instances that provide search suggestions. The goal is to enhance the performance and user experience of the search features by reducing the request frequency and the response time. The files affected are `static/js/dashboard.js`, `static/js/layout.js`, `static/js/notificationsCenter.js`, and `static/js/validatorRewards.js`.

### DIFF
--- a/static/js/dashboard.js
+++ b/static/js/dashboard.js
@@ -692,6 +692,24 @@ $(document).ready(function () {
   }
   create_validators_typeahead("input[aria-controls='validators']", "#validators")
 
+  var timeWait = 0
+  var debounce = function (context, func) {
+    var timeout, result
+
+    return function () {
+      var args = arguments,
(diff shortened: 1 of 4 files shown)


Eisei24 (approved):
lgtm

PR #2518, opened 2023-08-30
(BIDS-2424) stop search if ens name is valid but not registered

remoterami:
### <samp>🤖 Generated by Copilot at b25ee16</samp>

Improve search ahead functionality for ENS names. Handle unregistered ENS names in `handlers/search.go` by returning them as search queries.

### DIFF
--- a/handlers/search.go
+++ b/handlers/search.go
@@ -235,12 +235,8 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 			LEFT JOIN validators ON validators.pubkey = eth1_deposits.publickey
 			WHERE validators.pubkey IS NULL AND ENCODE(eth1_deposits.publickey, 'hex') LIKE ($1 || '%')`, lowerStrippedSearch)
 	case "indexed_validators_by_eth1_addresses":
-		if !utils.IsValidEnsDomain(search) && !utils.IsEth1Address(search) {
-			break
-		}
 		search = ReplaceEnsNameWithAddress(search)
-		if utils.IsValidEnsDomain(search) {
-			// unregistered ENS name
+		if !utils.IsEth1Address(search) {
 			break
 		}
 		result, err = FindValidatorIndicesByEth1Address(strings.ToLower(search))
--- a/handlers/search.go
+++ b/handlers/search.go
@@ -238,6 +238,11 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 		if !utils.IsValidEnsDomain(search) && !utils.IsEth1Address(search) {
 			break
 		}
+		search = ReplaceEnsNameWithAddress(search)
+		if utils.IsValidEnsDomain(search) {
+			// unregistered ENS name
+			break
+		}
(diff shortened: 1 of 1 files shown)


handlers/search.go
```diff
@@ -238,6 +238,11 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 		if !utils.IsValidEnsDomain(search) && !utils.IsEth1Address(search) {
 			break
 		}
+		search = ReplaceEnsNameWithAddress(search)
```
D13ce, line 241: While this works, it would be cleaner if `ReplaceEnsNameWithAddress` would return two values:
`string, bool`

- The `string` would be the same as now so it either is the address for a given ens name (if any) OR it is the same string that was passed to the function.
- The `bool` would tell the caller whether the passed parameter was an ens name that has an address associated to it (`true`) or not (`false`), i.e. `false` would be returned whenever the given parameter is not a valid ens name or there is no address associated with it.

You then could do something like this:
```
search, found = ReplaceEnsNameWithAddress(search)
if !found {
	break
}
```

This would improve readability. Of course this would require you to touch more code-parts as `ReplaceEnsNameWithAddress` is used quite a lot. In many cases something like
```
search, _ = ReplaceEnsNameWithAddress(search)
```
should suffice. Maybe we could refactor some other parts?

D13ce (commented):
Changes work but they COULD be improved. I left a comment, ptal and let me know what you think.

handlers/search.go
remoterami: Thanks for your input. I agree, the function could be reworked. I didn't do it for the same reason you mentioned, it'd be necessary to touch a few places.

However with your implementation suggestion I'm not sure if it'd really have the desired effect. If we only return `true` if it "was an ens name that has an address associated to it", that'd mean passing valid EL addresses themselves also return `false`. So if I'm not missing anything your function call wouldn't suffice to not `break` out in that case, we'd still need to add some kind of filter (which again defeats the purpose of the `bool`). So it wouldn't really improve my version imo, it's just a little more explicit.

Now I could also assume what you might've meant / is even more useful is "return `true` only if a valid EL address has been returned (whether because it already was a valid address before, or because it got converted after a successful ENS resolution), otherwise return `false`". That would obviously solve the issue. Furthermore this allows us to save many instances of duplicated code, as in almost all cases the calling function checks if a valid address was returned anyways (so we'd actually use the `bool` result, instead of throwing it away via `search, _ = ReplaceEnsNameWithAddress(search)`).

I'll quickly push these changes so we can see how it'd effect things. But lmk if we shouldn't go with that version.

D13ce: Okay, sorry, you are right, if you would have implemented my exact suggestion, this would not have worked for all cases.

But: The way your PR looks after your second commit now is good enough for me to approve it :+1:

Creating an internal ticket to refactor the function and related parts sounds like a good idea though. Would you mind doing that?

D13ce (approved):
lgtm (but please read my reply to https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2518/files#r1313028340)

PR #2519, opened 2023-08-30
(BIDS-2440) improve performance of withrdawals-page

guybrush:
### <samp>🤖 Generated by Copilot at cce4b89</samp>

This pull request improves the database and web interface for withdrawals and BLS public key changes in the eth2 beacon chain explorer. It uses binary fields and indexes for faster and more accurate queries, displays address names and metadata, and truncates and links long addresses. It also adds a new SQL migration file to alter the indexes on the `blocks_bls_change` and `blocks_withdrawals` tables.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -2297,16 +2297,7 @@ func GetWithdrawalsCountForQuery(query string) (uint64, error) {
 	var err error = nil
 
 	trimmedQuery := strings.ToLower(strings.TrimPrefix(query, "0x"))
-	// Check whether the query can be used for a validator, slot or epoch search
-	if uiQuery, parseErr := strconv.ParseUint(query, 10, 64); parseErr == nil {
-		searchQuery := `
-			WHERE w.validatorindex = $1
-				OR block_slot = $1
-				OR block_slot BETWEEN $1*$2 AND ($1+1)*$2-1`
-		err = ReaderDb.Get(&count, fmt.Sprintf(withdrawalsQuery, searchQuery),
-			uiQuery, utils.Config.Chain.Config.SlotsPerEpoch)
-
-	} else if addressRE.MatchString(query) {
+	if addressRE.MatchString(query) {
 		searchQuery := `WHERE address = $1`
 		addr, decErr := hex.DecodeString(trimmedQuery)
 		if err != nil {
@@ -2314,6 +2305,14 @@ func GetWithdrawalsCountForQuery(query string) (uint64, error) {
 		}
 		err = ReaderDb.Get(&count, fmt.Sprintf(withdrawalsQuery, searchQuery),
 			addr)
+	} else if uiQuery, parseErr := strconv.ParseUint(query, 10, 64); parseErr == nil {
+		// Check whether the query can be used for a validator, slot or epoch search
+		searchQuery := `
+			WHERE w.validatorindex = $1
+				OR block_slot = $1
+				OR block_slot BETWEEN $1*$2 AND ($1+1)*$2-1`
+		err = ReaderDb.Get(&count, fmt.Sprintf(withdrawalsQuery, searchQuery),
+			uiQuery, utils.Config.Chain.Config.SlotsPerEpoch)
 	}
 
 	if err != nil {
@@ -2363,15 +2362,7 @@ func GetWithdrawals(query string, length, start uint64, orderBy, orderDir string
 
 	trimmedQuery := strings.ToLower(strings.TrimPrefix(query, "0x"))
 	if trimmedQuery != "" {
-		// Check whether the query can be used for a validator, slot or epoch search
-		if uiQuery, parseErr := strconv.ParseUint(query, 10, 64); parseErr == nil {
-			searchQuery := `
-				WHERE w.validatorindex = $3
-					OR w.block_slot = $3
-					OR w.block_slot BETWEEN $3*$4 AND ($3+1)*$4-1`
-			err = ReaderDb.Select(&withdrawals, fmt.Sprintf(withdrawalsQuery, searchQuery, orderBy, orderDir),
-				length, start, uiQuery, utils.Config.Chain.Config.SlotsPerEpoch)
-		} else if addressRE.MatchString(query) {
+		if addressRE.MatchString(query) {
 			searchQuery := `WHERE address = $3`
 			addr, decErr := hex.DecodeString(trimmedQuery)
 			if decErr != nil {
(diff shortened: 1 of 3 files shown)


db/db.go
```diff
@@ -2295,16 +2300,20 @@ func GetWithdrawalsCountForQuery(query string) (uint64, error) {
 	// Check whether the query can be used for a validator, slot or epoch search
 	if uiQuery, parseErr := strconv.ParseUint(query, 10, 64); parseErr == nil {
 		searchQuery := `
-				WHERE w.validatorindex = $1
-					OR block_slot = $1
-					OR (block_slot / $3) = $1
-					OR address_text LIKE ($2 || '%')`
```
Eisei24, line 2301: **question:** Why was the address search removed here? Are there not valid addresses that can be parsed as **uint64**?

db/db.go
```diff
@@ -2354,17 +2367,18 @@ func GetWithdrawals(query string, length, start uint64, orderBy, orderDir string
 		if uiQuery, parseErr := strconv.ParseUint(query, 10, 64); parseErr == nil {
 			searchQuery := `
 				WHERE w.validatorindex = $3
-					OR block_slot = $3
-					OR (block_slot / $5) = $3
-					OR address_text LIKE ($4 || '%')`
```
Eisei24, line 2359: **question:** Why was the address search removed here? Are there not valid addresses that can be parsed as **uint64**?

db/db.go
```diff
@@ -2939,37 +2953,27 @@ func GetBLSChangesCountForQuery(query string) (uint64, error) {
...
-		) AS val ON val.validatorindex = bls.validatorindex`
-
 	// Check whether the query can be used for a validator, slot or epoch search
 	if uiQuery, parseErr := strconv.ParseUint(query, 10, 64); parseErr == nil {
 		searchQuery := `
 			WHERE bls.validatorindex = $1			
 				OR block_slot = $1
-				OR (block_slot / $3) = $1
-				OR pubkey_text LIKE ($2 || '%')
-				OR deposit_address_text LIKE ($2 || '%')`
```
Eisei24, line 2963: **question:** Why was the address search removed here? Are there not valid addresses that can be parsed as **uint64**?

db/db.go
```diff
@@ -3015,40 +3018,28 @@ func GetBLSChanges(query string, length, start uint64, orderBy, orderDir string)
...
-			) AS val ON val.validatorindex = bls.validatorindex`
-
 		// Check whether the query can be used for a validator, slot or epoch search
 		if uiQuery, parseErr := strconv.ParseUint(query, 10, 64); parseErr == nil {
 			searchQuery := `
 				WHERE bls.validatorindex = $3			
 					OR block_slot = $3
-					OR (block_slot / $5) = $3
-					OR pubkey_text LIKE ($4 || '%')
-					OR deposit_address_text LIKE ($4 || '%')`
```
Eisei24, line 3034: **question:** Why was the address search removed here? Are there not valid addresses that can be parsed as **uint64**?

db/db.go
```diff
@@ -50,8 +51,8 @@ var maxSqlNumber = uint64(9223372036854775807)
 
 const MaxSqlInteger = 2147483647
 
-var addressLikeRE = regexp.MustCompile(`^(0x)?[0-9a-fA-F]{0,40}$`)
-var blsLikeRE = regexp.MustCompile(`^(0x)?[0-9a-fA-F]{0,96}$`)
+var addressRE = regexp.MustCompile(`^(0x)?[0-9a-fA-F]{40}$`)
+var blsRE = regexp.MustCompile(`^(0x)?[0-9a-fA-F]{96}$`)
```
Eisei24, line 55: **suggestion:** We already have the **addressRE** as **eth1AddressRE** in **utils**.
We could move the **blsRE** over there to have the regular expression more grouped together.

db/db.go
guybrush: has not been removed, but we only search for exact matches

db/db.go
guybrush: has not been removed, but we only search for exact matches

db/db.go
guybrush: has not been removed, but we only search for exact matches

db/db.go
guybrush: would rather let it be like it is, we can change this in another pr

db/db.go
```diff
@@ -2341,7 +2353,7 @@ func GetWithdrawals(query string, length, start uint64, orderBy, orderDir string
 			w.amount
 		FROM blocks_withdrawals w
 		INNER JOIN blocks b ON w.block_root = b.blockroot AND b.status = '1'
-		%s
+		%s 
 		ORDER BY %s %s
 		LIMIT $1
 		OFFSET $2`
```
recy21, line 2359: As I learned offset is bad, can we find a way without?

db/db.go
```diff
@@ -2292,19 +2297,22 @@ func GetWithdrawalsCountForQuery(query string) (uint64, error) {
...
-			uiQuery, trimmedQuery, utils.Config.Chain.Config.SlotsPerEpoch)
-	} else if addressLikeRE.MatchString(query) {
-		searchQuery := `WHERE address_text LIKE ($1 || '%')`
+			addr)
+	} else if uiQuery, parseErr := strconv.ParseUint(query, 10, 64); parseErr == nil {
+		// Check whether the query can be used for a validator, slot or epoch search
+		searchQuery := `
+			WHERE w.validatorindex = $1
+				OR block_slot = $1
+				OR block_slot BETWEEN $1*$2 AND ($1+1)*$2-1`
```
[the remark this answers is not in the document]
recy21: A little below, it's like that w. before block_slot

db/db.go
```diff
@@ -2939,37 +2952,27 @@ func GetBLSChangesCountForQuery(query string) (uint64, error) {
...
-
-		err = ReaderDb.Get(&count, fmt.Sprintf(blsQuery, joinQuery, searchQuery),
-			uiQuery, trimmedQuery, utils.Config.Chain.Config.SlotsPerEpoch)
-	} else if blsLikeRE.MatchString(query) {
-		searchQuery := `
-				WHERE pubkey_text LIKE ($1 || '%')
-				OR deposit_address_text LIKE ($1 || '%')`
-		err = ReaderDb.Get(&count, fmt.Sprintf(blsQuery, joinQuery, searchQuery),
-			trimmedQuery)
+				OR block_slot BETWEEN $1*$2 AND ($1+1)*$2-1`
```
recy21, line 2973: `bls.block_slot` (add bls.)

db/migrations/20230830144100_improve_blocks_bls_change_pubkey_text_idx.sql
```diff
@@ -0,0 +1,24 @@
+-- +goose Up
```
recy21, line 1: Missing CONCURRENTLY for dropping & creating an index.

recy21 (changes requested):

db/db.go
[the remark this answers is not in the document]
guybrush: done

db/db.go
guybrush: the offset is limited to 10k here, this will not be a problem

offset is not bad its just not useful without limits for paging of close to large tables

db/db.go
guybrush: done

db/migrations/20230830144100_improve_blocks_bls_change_pubkey_text_idx.sql
guybrush: done

recy21 (approved):

PR #2524, opened 2023-08-31
(BIDS-2443) Remove education route, handler and html file

D13ce:
This PR removes the inaccessible "education" page from the repo. This includes the route, the handler and the HTML file.


Eisei24 (approved):
lgtm

PR #2529, opened 2023-09-04
New total supply endpoint

mxmar:
This PR adds new `Beaconcha.in` explorer API endpoint to integrate with 3rd party markets and exchanges (`CoinGecko`, `Coinmarketcap`, they need to know the `TotalSupply` for native tokens like `LYX` is for example). 

There was an official foundation founds allocation made in chain config under `alloc`: https://github.com/lukso-network/network-configs/blob/e3a2e914d76db0c5e93d93ade6caafef9936a6d9/mainnet/shared/genesis.json#L34 (and supply was announced and explained in https://medium.com/lukso/its-happening-the-genesis-validators-are-coming-ce5e07935df6 blog post for example).

State:
- It's placed under `Misc` API category and documented (Swagger)
- It's based on `genesis.json` allocation (for foundation, allocated in chain config)
- It makes calculation of sum of `genesis.json` (alloc) supply, total withdrawn amount and current consensus validator balances (current balances based on https://lighthouse-book.sigmaprime.io/validator-inclusion.html#global)

### DIFF
--- a/handlers/network.go
+++ b/handlers/network.go
@@ -1,0 +1,75 @@
+package handlers
+
+import (
+	"encoding/json"
+	"eth2-exporter/db"
+	"eth2-exporter/rpc"
+	"eth2-exporter/services"
+	"eth2-exporter/types"
+	"eth2-exporter/utils"
+	"github.com/ethereum/go-ethereum/params"
+	"math/big"
+	"net/http"
+)
+
+const (
+	GWei = 1e9
+)
+
+// Supply godoc
+// @Summary Get total supply of all native network tokens in existence
+// @Tags Misc
+// @Description Returns information about total supply of all native network tokens in existence (in wei).
+// @Produce  json
+// @Success 200 {object} types.ApiResponse{data=types.SupplyResponse} "Success"
+// @Failure 400 {object} types.ApiResponse "Failure"
+// @Failure 500 {object} types.ApiResponse "Server Error"
+// @Router /api/v2/totalsupply [get]
+func Supply(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	genesisTotalSupply := utils.Config.Chain.GenesisTotalSupply
+
+	totalAmountWithdrawn, _, err := db.GetTotalAmountWithdrawn()
+	if err != nil {
+		logger.WithError(err).Error("error getting total amount withdrawn from db")
+	}
+
+	latestFinalizedEpoch := services.LatestFinalizedEpoch()
+	if err != nil {
+		logger.WithError(err).Error("error getting latest finalized epoch")
+	}
+
+	chainIDBig := new(big.Int).SetUint64(utils.Config.Chain.Config.DepositChainID)
+	rpcClient, err := rpc.NewLighthouseClient("http://"+utils.Config.Indexer.Node.Host+":"+utils.Config.Indexer.Node.Port, chainIDBig)
+	if err != nil {
+		logger.WithError(err).Error("new total supply Lighthouse client in monitor error")
+	}
+
+	// Get the total staked gwei that was active (i.e., able to vote) during the latestFinalizedEpoch epoch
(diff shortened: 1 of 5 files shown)


mxmar:
@recy21 Should I create PR with `switch cfg.Chain.Name` conditions for `lukso` mainnet and testnet? Can I add support for the official LUKSO Mainnet and Testnet in `beaconcha.in` explorer? Like it is with `gnosis` chain for example? Link: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d78aa9d771443d07a791bf111ed5dd312dd22bed/utils/utils.go#L410

recy21:
@mxmar 

Thank you for your interest in our project!
Created an internal ticket, a developer will have a look / review your code.

recy21:
> @recy21 Should I create PR with `switch cfg.Chain.Name` conditions for `lukso` mainnet and testnet? Can I add support for the official LUKSO Mainnet and Testnet in `beaconcha.in` explorer? Like it is with `gnosis` chain for example? Link:
> 
> https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d78aa9d771443d07a791bf111ed5dd312dd22bed/utils/utils.go#L410

Not at the moment

handlers/network.go
```diff
@@ -0,0 +1,75 @@
...
+
+	response := &types.ApiResponse{
+		Status: "OK",
+		Data:   data,
+	}
+
+	err = json.NewEncoder(w).Encode(response)
+	if err != nil {
+		logger.WithError(err).WithField("route", r.URL.String()).Error("error encoding json response")
+		http.Error(w, "Internal server error", http.StatusServiceUnavailable)
```
Eisei24, line 71: Similar to other API this could be
```suggestion
sendServerErrorResponse(w, r.URL.String(), "could not serialize data results")
```

handlers/network.go
```diff
@@ -0,0 +1,75 @@
...
+		return
```
Eisei24, line 72: Unnecessary return.

handlers/network.go
```diff
@@ -0,0 +1,75 @@
...
+// @Failure 500 {object} types.ApiResponse "Server Error"
+// @Router /api/v2/totalsupply [get]
+func Supply(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	genesisTotalSupply := utils.Config.Chain.GenesisTotalSupply
+
+	totalAmountWithdrawn, _, err := db.GetTotalAmountWithdrawn()
+	if err != nil {
+		logger.WithError(err).Error("error getting total amount withdrawn from db")
```
Eisei24, line 35: Should this function continue if there is an error?
Seems like this could be added
```suggestion
		logger.WithError(err).Error("error getting total amount withdrawn from db")
		sendServerErrorResponse(w, r.URL.String(), "Appropriate message here")
		return
```
This also applies to the other errors.

Eisei24 (changes requested):

Buttaa:
Hi @mxmar , we had a closer look and tried to merge this PR. But unfortunately, there is no standardized way to get the balances of the genesis block. Thus, we decided to close this PR.

frozeman:
There is a standardised way:
```
+ Take all account balances from genesis block accounts
+ eligibleEther (currently staked)
+ totalAmountWithdrawn
- deposit contract balance
- burned coins
```
This is an endpoint surely all chains need.

Buttaa:
> @Buttaa 
> 
> There is a standardised way:
> 
> ```
> 
> + Take all account balances from genesis block accounts
> 
> + eligibleEther (currently staked)
> 
> + totalAmountWithdrawn
> 
> - deposit contract balance
> 
> - burned coins
> 
> ```
> 
> This is an endpoint surely all chains need.

We looked for a genesis file for each network but were not able to find one. I assume this is because some networks are POA, some are not.

frozeman:
Maybe a config could help here.
The endpoint could be option, and you can define a initial balance.

PR #2530, opened 2023-09-05
(BIDS-2432) Remove text-primary style from 0x in FormatAddressLong

D13ce:
This PR causes the `0x` prefix of address not to be color orange anymore.

Current example: https://prater.beaconcha.in/address/0xe6a651e0a30fda0dc7f8b709287f5b76e0109202
image

This PR replaces https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2522.

### DIFF
--- a/utils/eth1.go
+++ b/utils/eth1.go
@@ -240,9 +240,9 @@ func FormatHashLong(hash common.Hash) template.HTML {
 	if len(address) > 4 {
 		htmlFormat := `
 		<div class="d-flex text-monospace">
-			<span class="">%s</span>
+			%s
 			<span class="flex-shrink-1 text-truncate">%s</span>
-			<span class="">%s</span>
+			%s
 		</div>`
 
 		return template.HTML(fmt.Sprintf(htmlFormat, address[:4], address[4:len(address)-4], address[len(address)-4:]))
--- a/utils/eth1.go
+++ b/utils/eth1.go
@@ -256,9 +256,9 @@ func FormatAddressLong(address string) template.HTML {
 	}
 	address = FixAddressCasing(address)
 	test := `
-	<span class="text-monospace mw-100"><span class="text-primary">%s</span><span class="text-truncate">%s</span><span class="text-primary">%s</span></span>`
+	<span class="text-monospace mw-100"><span class="text-truncate">%s</span><span class="text-primary">%s</span><span class="text-truncate">%s</span><span class="text-primary">%s</span></span>`
 	if len(address) > 4 {
-		return template.HTML(fmt.Sprintf(test, address[:6], address[6:len(address)-4], address[len(address)-4:]))
+		return template.HTML(fmt.Sprintf(test, address[:2], address[2:6], address[6:len(address)-4], address[len(address)-4:]))
 	}
 
 	return template.HTML(address)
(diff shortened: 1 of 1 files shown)


utils/eth1.go
```diff
@@ -256,9 +256,9 @@ func FormatAddressLong(address string) template.HTML {
 	}
 	address = FixAddressCasing(address)
 	test := `
```
LuccaBitfly, line 258: why is this named `test`?

utils/eth1.go
```diff
@@ -256,9 +256,9 @@ func FormatAddressLong(address string) template.HTML {
...
-	<span class="text-monospace mw-100"><span class="text-primary">%s</span><span class="text-truncate">%s</span><span class="text-primary">%s</span></span>`
+	<span class="text-monospace mw-100"><span class="text-truncate">%s</span><span class="text-primary">%s</span><span class="text-truncate">%s</span><span class="text-primary">%s</span></span>`
```
LuccaBitfly, line 259: I'm pretty sure that text-truncate doesn't do anything in a span surrounded by spans that do not have it. Would suggest removing the text-truncate spans.

utils/eth1.go
```diff
@@ -256,9 +256,9 @@ func FormatAddressLong(address string) template.HTML {
...
 	if len(address) > 4 {
-		return template.HTML(fmt.Sprintf(test, address[:6], address[6:len(address)-4], address[len(address)-4:]))
+		return template.HTML(fmt.Sprintf(test, address[:2], address[2:6], address[6:len(address)-4], address[len(address)-4:]))
```
LuccaBitfly, line 261: Don't know it its okay to assume an existing index of 6 here, please let me know if it is.

LuccaBitfly (changes requested):

utils/eth1.go
D13ce: To **test** you code review skills.

But seriously, I don't know, I did not choose the name. But of course, your implied "Please use a meaningful name for this" is still valid so I've changed the name accordingly.

I even fixed it at in a neighboring function. AND I have made both functions more performant for some cases.

All thanks to you :heart:

utils/eth1.go
D13ce: You are right. Removed and tested it, works perfectly fine.

utils/eth1.go
D13ce: It is okay (was okay before, won't refactor in this PR).

utils/eth1.go
```diff
@@ -237,14 +237,15 @@ func FormatAddressAsTokenLink(token, address []byte, name string, verified bool,
...
-		<span class="flex-shrink-1 text-truncate">%s</span>
-		<span class="">%s</span>
-	</div>`
 	if len(address) > 4 {
-		return template.HTML(fmt.Sprintf(test, address[:4], address[4:len(address)-4], address[len(address)-4:]))
+		htmlFormat := `
+		<div class="d-flex text-monospace">
+			<span class="">%s</span>
+			<span class="flex-shrink-1 text-truncate">%s</span>
+			<span class="">%s</span>
```
LuccaBitfly, line 245:
```suggestion
			<span>%s</span>
			<span class="flex-shrink-1 text-truncate">%s</span>
			<span>%s</span>
```
😅

LuccaBitfly (changes requested):
Sorry

LuccaBitfly (approved):
looks like you removed the text-primary style from 0x to me

PR #2534, opened 2023-09-06
(BIDS-2454) Update dependencies and solve breaking changes

D13ce:
This PR updates `go_ens` to include https://github.com/wealdtech/go-ens/issues/34

That way it is now able to hash emojis including zero width joiners.

For example:
🚣‍♂️ = "Male person rowing boat"

This is a single emoji composed of
🚣 = "Gender neutral person rowing boat"
\u200d = Zero width joiner
♂️ = "Mail sign"

As 🚣‍♂️ therefore includes 3 signs, "🚣‍♂️.eth" is a valid ENS name. Said name was not hashable be `go_ens` before though.

---

Here is some test code to verify the changes:
```Go
str := "🚣‍♂️"
nameHash, err := go_ens.NameHash(str)
if err != nil {
   fmt.Println("error for %s: %v", str, err)
} else {
   fmt.Println(nameHash)
}
```

Without this PR, this returns an err. With this PR, everything's fine.

Thanks to @mcdee for solving this issue so quickly :+1:

---

Several other dependencies have been updated along the way introducing breaking changes in `go_ethereum`: https://github.com/ethereum/go-ethereum/commit/67ac5f0ae797fbafa1e57e982cab4c2d291244c3

`func (tx *Transaction) AsMessage` has been removed [here](https://github.com/ethereum/go-ethereum/commit/67ac5f0ae797fbafa1e57e982cab4c2d291244c3#diff-8fb609018eccff10931645347515b8ecc194992cc4ef9e84ebdea87682dd50c3), instead `func TransactionToMessage` added [here](https://github.com/ethereum/go-ethereum/commit/67ac5f0ae797fbafa1e57e982cab4c2d291244c3#diff-8c398ef0a79f97ba6d497a99247a815b3a51918de9e16954215b6073f907c92e) has to be used.

The breaking changes have been addressed of course.


utils/utils.go
```diff
@@ -854,7 +854,7 @@ func TryFetchContractMetadata(address []byte) (*types.ContractMetadata, error) {
 // 	}
 
 // 	if resp.StatusCode == 200 {
-// 		body, err := ioutil.ReadAll(resp.Body)
+// 		body, err := io.ReadAll(resp.Body)
 // 		if err != nil {
 // 			return nil, err
 // 		}
```
LuccaBitfly, line 860: Why not remove this?

services/configuration.go
```diff
@@ -16,10 +16,12 @@ const (
 	ConfigurationKeyHardforkName     types.ExplorerConfigurationKey = "HardforkName"
 )
 
-/***
+/*
+**
 This is the list of possible configurations that can be changed in the explorer administration
 Per default these values will be taken, overridden by the values from the db, if they exist.
-***/
+**
```
LuccaBitfly, line 23: Is this change intentional? Not saying this is wrong to do.

LuccaBitfly (changes requested):

services/configuration.go
D13ce: Yes, this is indeed intentional because otherwise https://github.com/gobitfly/eth2-beaconchain-explorer/actions/runs/6094260445/job/16535457583

utils/utils.go
D13ce: Done.

LuccaBitfly (approved):
lgtm

PR #2535, opened 2023-09-06
(BIDS-2452) limit ens name size to 2KB

remoterami:
### <samp>🤖 Generated by Copilot at a16cd40</samp>

Removed ENS name validation and added length check for validator names in `db/ens.go`. This enabled users to set custom names for validators in the explorer.

### DIFF
--- a/db/ens.go
+++ b/db/ens.go
@@ -311,7 +311,7 @@ func (bigtable *Bigtable) TransformEnsNameRegistered(blk *types.Eth1Block, cache
 func verifyName(name string) error {
 	// limited by max capacity of db (caused by btrees of indexes); tests showed maximum of 2684 (added buffer)
 	if len(name) > 2048 {
-		return fmt.Errorf("name too long")
+		return fmt.Errorf("name too long: %v", name)
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


db/ens.go
```diff
@@ -308,10 +308,9 @@ func (bigtable *Bigtable) TransformEnsNameRegistered(blk *types.Eth1Block, cache
 	return bulkData, bulkMetadataUpdates, nil
 }
 
-// Ens names are only supported to a length of 40
-// Source: https://github.com/wealdtech/go-ens/blob/5b323a4ef0472f06c515723b060b166843b9db08/resolver.go#L190
 func verifyName(name string) error {
-	if (strings.HasPrefix(name, "0x") && len(name) > 42) || (!strings.HasPrefix(name, "0x") && len(name) > 40) {
+	// limited by max capacity of db (caused by btrees of indexes); tests showed maximum of 2684 (added buffer)
+	if len(name) > 2048 {
 		return fmt.Errorf("name too long")
```
D13ce, line 314: This is not touched by you, but we should consider adding the name that causes this to the error message. This should be either be done here directly or here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/50591777634fd593c904a90881e2d5fe15f5887e/db/ens.go#L184
and here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/50591777634fd593c904a90881e2d5fe15f5887e/db/ens.go#L226

Thoughts?

D13ce (approved):
utACK

PR #2540, opened 2023-09-07
(BIDS-2457) remove previous odao members

remoterami:
### <samp>🤖 Generated by Copilot at 200ecef</samp>

This pull request improves the integration of RocketPool DAO data in the `exporter/rocketpool.go` file. It adds logic to manage the DAO members in the database and avoid stale data.

### DIFF
--- a/exporter/rocketpool.go
+++ b/exporter/rocketpool.go
@@ -1117,6 +1117,7 @@ func (rp *RocketpoolExporter) SaveDAOMembers() error {
 
 		valueStrings := make([]string, 0, batchSize)
 		valueArgs := make([]interface{}, 0, batchSize*nArgs)
+		addresses := make([][]byte, 0, batchSize)
 		for i, d := range data[start:end] {
 			for j := 0; j < nArgs; j++ {
 				valueStringsArgs[j] = i*nArgs + j + 1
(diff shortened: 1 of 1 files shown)


exporter/rocketpool.go
```diff
@@ -1130,12 +1131,39 @@ func (rp *RocketpoolExporter) SaveDAOMembers() error {
...
+				unbonded_validator_count = excluded.unbonded_validator_count
+			`, strings.Join(valueStrings, ","))
 		_, err := tx.Exec(stmt, valueArgs...)
 		if err != nil {
 			return fmt.Errorf("error inserting into rocketpool_dao_members: %w", err)
 		}
+
+		_, err = tx.Exec(`
+			DELETE FROM rocketpool_dao_members
+			WHERE NOT address = ANY($1)`, addresses)
```
Eisei24, line 1163: **suggestion:** Could use a 
```suggestion
			WHERE NOT address = ANY($1)`, pq.ByteaArray(addresses))
```

Eisei24 (approved):
utACK

exporter/rocketpool.go
remoterami: Done, but that's the first/only use of `pq` in that file, so have to import

PR #2547, opened 2023-09-13
(BIDS-2466) Removed redirect from session data

Eisei24:
### <samp>🤖 Generated by Copilot at 474c8fa</samp>

Enhanced login and logout features to support OAuth2 authorization flows with third-party applications. Updated `login.html` template and `handlers/auth.go` and `handlers/user.go` files to handle OAuth2 parameters and redirections.

### DIFF
--- a/cmd/explorer/main.go
+++ b/cmd/explorer/main.go
@@ -555,7 +555,6 @@ func main() {
 
 			oauthRouter := router.PathPrefix("/user").Subrouter()
 			oauthRouter.HandleFunc("/authorize", handlers.UserAuthorizeConfirm).Methods("GET")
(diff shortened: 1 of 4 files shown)


manuelsc (approved):
lgtm

There is a slight catch here though, with sessions this also worked when the user had no account, registered and then logged in. And the second scenario where the user pressed on forgot password, reset it and then logged in again.

But we can do this in two separate tasks if you like, for now it's important to ship this fix as soon as possible. If you think these two cases can be addressed before the next explorer version feel free to go for it, otherwise please create a bids so these cases won't get lost.

Eisei24:
A new BIDS has been created to discuss and fix the desired handling of the redirect with the register and "forgot password" pages.

PR #2563, opened 2023-09-19
(BIDS-2470) Set a limit for the API endpoint

Eisei24:
### <samp>🤖 Generated by Copilot at 11e3df5</samp>

Improved the `ApiValidator` endpoints in `handlers/api.go` to support queries with more than 100 validators and prevent errors. Removed unnecessary import.


D13ce (approved):
lgtm

PR #2565, opened 2023-09-20
(BIDS-2495) refactor income charts

LuccaBitfly:
This PR overhauls both income charts that exist on the validator and dashboard respectively a bit.

1. The execution data is now deduped in the backend so that each day only has one chart point (2 per day with consensus), which also removes the weird lines in the execution columns (this is approved). This hopefully removes all tooltip shenanigans we had to deal with the past few weeks (the bug this BIDS is concerned with was fixed automatically, didn't even have to do anything).
2. The chart options are now located in their own JS file, so that dashboard and vali can both be changed at the same time.
3. Only the currently selected currency is shown in the income tooltips, as this was buggy on mainnet for other currencies (this is approved).

copilot:summary


static/js/income_chart_options.js
```diff
@@ -0,0 +1,132 @@
...
+        // add total if hovered point contains rewards for both EL and CL
+        if (tooltip.chart.hoverPoints.length > 1) {
+          text += `<b>Total:</b> ${getIncomeChartValueString(total, currency)}`
+        }
+
+        return text
+      },
+      exporting: {
+        scale: 1,
+      },
```
remoterami, line 112: I think this is not needed, `exporting` is neither included nor defined on tooltips

remoterami (approved):
utACK

(left a small comment)

static/js/income_chart_options.js
LuccaBitfly: used wrong indentation, fixed

remoterami: :+1: But is exporting of the charts even desired? As neither of boths graphs can be exported currently. If yes, that's not working because the setting seems to get overwritten in `highcharts-global-options.js`

LuccaBitfly: I don't know tbh, the chart on the dashboard had that option set and I didn't want to remove it. I can remove it if you want to.

D13ce (commented):
Really nice re-factor, hats off :+1: 

One question though: The income chart on the dashboard showed for example `1.00000 ETH` to caption the ticks on the Y-Axis, now it only shows `1 ETH` (which is consistent with the validator's page income chart).
=> Is this intended?

image

Furthermore, I have added a suggestion, ptal.

handlers/common.go
```diff
@@ -727,3 +728,59 @@ func GetWithdrawableCountFromCursor(epoch uint64, validatorindex uint64, cursor
...
+
+	// Now populate the chartData array using the dayRewardMap
+	for day, reward := range dayRewardMap {
+		ts := float64(utils.DayToTime(day).Unix() * 1000)
+		chartData = append(chartData, &types.ChartDataPoint{
+			X:     ts,
+			Y:     utils.ExchangeRateForCurrency(currency) * reward,
+			Color: color,
+		})
+	}
```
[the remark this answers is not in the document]
LuccaBitfly: fixed

LuccaBitfly:
I did not consciously make that change, but it is better imo.

D13ce:
Okay, as discussed, it is slightly better than before but not perfect. In the scope of this PR, this is good enough though, so ...

D13ce (approved):
... lgtm

PR #2566, opened 2023-09-20
(BIDS-2499) fixed url query conversion

remoterami:
### <samp>🤖 Generated by Copilot at 6827f23</samp>

Fixed a browser compatibility bug in `static/js/datatable_loader.js` by using `$.param` instead of `new URLSearchParams` to serialize query parameters. This improved the performance and usability of the data tables in the website.


D13ce (approved):
lgtm

PR #2567, opened 2023-09-20
Bids 2481/all datatables should be debounced

Eisei24:
### <samp>🤖 Generated by Copilot at a29b4ae</samp>

The pull request refactors the code for loading data from the server for various DataTables in the web pages. It introduces a common function `dataTableLoader` in `datatable_loader.js` that handles the server requests with retries and extra query parameters. It replaces the `ajax` option for many DataTables with a call to this function, and imports the `datatable_loader.js` file in the templates that use it. This improves the reliability and performance of the data loading and rendering.

### DIFF
--- a/templates/validator/tables.html
+++ b/templates/validator/tables.html
@@ -201,7 +201,7 @@
                         {{ if .NextWithdrawalRow }}
                           let row = {{.NextWithdrawalRow}}
                           if (row && row.length) {
-                            data = row.concat(data)
+                            data.data = row.concat(data.data)
                           }
                         {{ end }}
                         return data


static/js/datatable_loader.js
```diff
@@ -25,14 +27,17 @@ function dataTableLoader(path) {
   }
 
   const doFetch = (tableData, callback) => {
-    fetch(`${path}?${new URLSearchParams(tableData)}`)
+    fetch(`${path}?${new URLSearchParams(tableData)}&${param}`)
```
D13ce, line 30: I am not the reviewer of this PR but I have reviewed https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2566 just now and it touches the same line of code slightly differently.

@Eisei24 , is your change here to fix the filter? It was not working on the `/slots` page before and a change in this line fixed it.

If so, please take a look at https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2566 and maybe discuss with @remoterami what approach is better.

Eisei24: We want both changes, they don't really conflict with each other.
I merged the current master for this branch.

templates/validators.html
```diff
@@ -116,7 +117,7 @@ <h6 class="dropdown-header"><span>Filter state</span> <span><span class="text-su
                 stateLoadCallback: function (settings) {
                   return JSON.parse(localStorage.getItem("DataTables_" + settings.sInstance))
                 },
-                ajax: '/validators/data?filterByState=' + state,
+                ajax: dataTableLoader('/validators/data', new URLSearchParams([["filterByState", state]])),
```
LuccaBitfly, line 120: Debouncing does not work if a filter is set

LuccaBitfly (changes requested):

templates/validators.html
Eisei24: This works in a weird way.
If you refresh the page and then change the filter it does not debounce further quick page changes.
But if you refresh the page with a filter already set the quick page changes ARE debounced.

templates/validator/tables.html
```diff
@@ -207,8 +205,7 @@
                           }
                         {{ end }}
                         return data
-                      },
-                    },
+                      }),
```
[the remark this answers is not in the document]
Eisei24: I do not get this `Processing...`, the row for the next withdrawal is just missing.
This should not be an issue that appeared because of my changes, the current master also does not show the row.
If I remember correctly I did test it on a branch where it still worked and my changes looked fine but I can't find such a branch that starts anymore, maybe because of further db changes.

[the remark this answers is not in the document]
LuccaBitfly: Master branch:
image
Your branch (stays like this):
image
Your branch after manuelly setting `NextWithdrwawalRow` to nil 
image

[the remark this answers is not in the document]
Eisei24: Thanks for the screenshots.
This should be fixed now. I could not test it on Sepolia so I used the manual entry.

LuccaBitfly (approved):
lgtm

PR #2571, opened 2023-09-21
(BIDS-2440) fix query for filtered bls-changes-count on withdrawals-page

guybrush:
### <samp>🤖 Generated by Copilot at b838c5e</samp>

Fixed a bug in `db/db.go` that affected the BLS key change search feature. Used a more efficient database query to fetch the count of BLS key changes for a given query.


D13ce (approved):
lgtm

PR #2576, opened 2023-09-26
(BIDS-2504) added name resolution to deposit address

remoterami:
Adds address name resolution to deposit addresses.
I'm aware that adding the `FromName` to the `Eth1Deposit` type in `exporter.go` is a bit of a workaround, lmk if that's too bad style and I should rather introduce new types (which would be redundant for most fields) used for frontend representation.

### <samp>🤖 Generated by Copilot at e92a323</samp>

This pull request adds a feature to show address names for validator deposits if available. It modifies the `db`, `types`, `utils`, and `templates` packages to retrieve, store, format, and display the address names.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -594,20 +594,20 @@ func GetValidatorDeposits(publicKey []byte) (*types.ValidatorDeposits, error) {
 	}
 	if len(deposits.Eth1Deposits) > 0 {
 		deposits.LastEth1DepositTs = deposits.Eth1Deposits[len(deposits.Eth1Deposits)-1].BlockTs
-	}
 
-	// retrieve address names from bigtable
-	names := make(map[string]string)
-	for _, v := range deposits.Eth1Deposits {
-		names[string(v.FromAddress)] = ""
-	}
-	names, _, err = BigtableClient.GetAddressesNamesArMetadata(&names, nil)
-	if err != nil {
-		return nil, err
-	}
+		// retrieve address names from bigtable
+		names := make(map[string]string)
+		for _, v := range deposits.Eth1Deposits {
+			names[string(v.FromAddress)] = ""
+		}
+		names, _, err = BigtableClient.GetAddressesNamesArMetadata(&names, nil)
+		if err != nil {
+			return nil, err
+		}
 
-	for k, v := range deposits.Eth1Deposits {
-		deposits.Eth1Deposits[k].FromName = names[string(v.FromAddress)]
+		for k, v := range deposits.Eth1Deposits {
+			deposits.Eth1Deposits[k].FromName = names[string(v.FromAddress)]
+		}
 	}
 
 	err = ReaderDb.Select(&deposits.Eth2Deposits, `
--- a/db/db.go
+++ b/db/db.go
@@ -596,6 +596,20 @@ func GetValidatorDeposits(publicKey []byte) (*types.ValidatorDeposits, error) {
 		deposits.LastEth1DepositTs = deposits.Eth1Deposits[len(deposits.Eth1Deposits)-1].BlockTs
 	}
 
+	// retrieve address names from bigtable
+	names := make(map[string]string)
+	for _, v := range deposits.Eth1Deposits {
+		names[string(v.FromAddress)] = ""
+	}
+	names, _, err = BigtableClient.GetAddressesNamesArMetadata(&names, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	for k, v := range deposits.Eth1Deposits {
+		deposits.Eth1Deposits[k].FromName = names[string(v.FromAddress)]
+	}
+
 	err = ReaderDb.Select(&deposits.Eth2Deposits, `
 		SELECT 
 			blocks_deposits.block_slot,
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -2831,7 +2831,7 @@ func (bigtable *Bigtable) GetAddressErc721TableData(address string, search strin
 			utils.FormatTimestamp(t.Time.AsTime().Unix()),
 			from,
 			to,
-			utils.FormatAddressAsLink(t.TokenAddress, "", false, true),
+			utils.FormatAddressAsLink(t.TokenAddress, "", true),
 			new(big.Int).SetBytes(t.TokenId).String(),
 		}
 	}
@@ -2946,7 +2946,7 @@ func (bigtable *Bigtable) GetAddressErc1155TableData(address string, search stri
 			utils.FormatTimestamp(t.Time.AsTime().Unix()),
 			from,
 			to,
-			utils.FormatAddressAsLink(t.TokenAddress, "", false, true),
+			utils.FormatAddressAsLink(t.TokenAddress, "", true),
 			new(big.Int).SetBytes(t.TokenId).String(),
 			new(big.Int).SetBytes(t.Value).String(),
 		}
--- a/templates/relays.html
+++ b/templates/relays.html
@@ -129,7 +129,7 @@
                         <td>
                           <span class="convertmetoutf8" id="slot_{{ .Slot }}" aria-hex-data="{{ .BlockExtraData }}"></span>
                         </td>
-                        <td>{{ formatAddressAsLink .ProposerFeeRecipient "" false false }}</td>
+                        <td>{{ formatAddressAsLink .ProposerFeeRecipient "" false }}</td>
                         <td>{{ formatBuilder .Builder }}</td>
                       </tr>
                     {{ end }}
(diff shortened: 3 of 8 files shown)


db/db.go
```diff
@@ -596,6 +596,20 @@ func GetValidatorDeposits(publicKey []byte) (*types.ValidatorDeposits, error) {
 		deposits.LastEth1DepositTs = deposits.Eth1Deposits[len(deposits.Eth1Deposits)-1].BlockTs
 	}
 
+	// retrieve address names from bigtable
+	names := make(map[string]string)
+	for _, v := range deposits.Eth1Deposits {
+		names[string(v.FromAddress)] = ""
+	}
+	names, _, err = BigtableClient.GetAddressesNamesArMetadata(&names, nil)
```
D13ce, line 604: **Question:** Wouldn't it be enough to simply call
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/8c2348627597d5abb97ac16efe506a89f1fda4f1/db/ens.go#L604
here?

I am pretty sure that `GetAddressesNamesArMetadata` is the wrong function. For example, for the zero address on Sepolia it returns "ZERO_ADDRESS" as name. We want to show the ENS though and "ZERO_ADDRESS" is not a valid END name.

If I am right, please also consider removing
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/e92a323f8087281e9b9b908467db7026d85388b1/utils/format.go#L342-L346

We want to show an ENS name like done for example [HERE](https://prater.beaconcha.in/address/0x000002b4A2e3459331b1eB6066A8FdB39c7fdC5D):
image

There should already be a format function. If not it would be good to implement one and to use it wherever ENS names are shown.

db/db.go
```diff
@@ -596,6 +596,20 @@ func GetValidatorDeposits(publicKey []byte) (*types.ValidatorDeposits, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for k, v := range deposits.Eth1Deposits {
+		deposits.Eth1Deposits[k].FromName = names[string(v.FromAddress)]
+	}
+
```
D13ce, line 612: **Suggestion:** All this code should be placed in the if above:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/e92a323f8087281e9b9b908467db7026d85388b1/db/db.go#L595-L597

Yes, there SHOULD not be a validator without deposit but still it would increase readability of the code.

D13ce (changes requested):

db/db.go
RomiRand: That screenshot is exactly (one of the spots) where I was getting that function `GetAddressesNamesArMetadata` from (coming from [`AddressTransactionsTableGrid`](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/templates/execution/address.html#L535C6-L535C6) with `TransactionsTable` used [here](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/handlers/eth1Account.go#L257), which filled `fromName` by that functions results [here](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/db/bigtable_eth1.go#L2043)). The function also calls the suggested function `GetEnsNamesForAddress` (inside of [bigtable.GetAddressNames](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/db/bigtable_eth1.go#L1953)).

I also couldn't verify that "ZERO_ADDRESS" gets returned without db access, but that won't be the case anyways.

So it's not like the function is unrelated to ENS, but of course it also adds some overhead. The main reason I picked it is that, as mentioned, other places use it too - I thought that there might be other information than ENS (contract metadata, user/team-defined tags etc) that could be relevant, so it was used in other places. But I agree, if we just want bare ENS information, we can simply skip the overhead. Please let me know what you think of this, if you still think it should be changed I will do so.

D13ce: Good write down. I agree with your reasoning. There's always a "but" though.

BUT: The example in the screenshot uses
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/7e539bd11547e1e266642f7806328358972404f2/utils/eth1.go#L107
here:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/6a5f43b1cd0c0b17e7b9679e68f64d5c22e496b7/db/bigtable_eth1.go#L2053

Why did you have to introduce a new formatting function here:
https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2576/files#diff-5020779d5bcfaa28d8f25d2f4799326a80ad320c4a6f9ee0e17be6407b7c5e00R342-R347
?

I am mostly confused as we already have several places to show an ENS/address name but this function is only used for one of those. Is there potential to clean this up?

remoterami: I'm sorry for the confusion, should've explained my reasoning in more details. Since I don't want to take too much of your time, there's a TLDR and I can push another suggestion so we don't just keep discussing this back and forth.. I didn't want this to become that big, after all this is just a small issue :sweat_smile: 

**TLDR**
Essentially because a) I didn't like the _registered html-template formatting functions_ available in that case for various reasons and/or b) storing HTML-formatted data returned by `FormatAddress` among db query results [here](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2576/files#diff-aa43bc4292f439dd71618a51c3fc4eb6957ff2c8255460848da513340cc3120fR321) is bad style imo.

**Longer version**
So from my understanding, there seem to be two ways to handle frontend formatting inside of the `handlers` (relevant to this discussion at least, there could be more):
1. Some backend data is pulled from `bigtable_eth1.go`. It defines some functions `...TableData` which return formatted data. It's not necessary to apply any more formatting in the template `.html`s as a last step. E.g. [your example](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/6a5f43b1cd0c0b17e7b9679e68f64d5c22e496b7/db/bigtable_eth1.go#L2053), used in html [here](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/templates/execution/address.html#L529-L567)
2. Other data is pulled from `db.go`. It does not define functions returning data designed for frontend use. Now it's necessary to apply formatting to the results in the template `.html`s. The deposits table referenced in this PR is such a case: data set [here](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/handlers/validator.go#L161-L180), formatting is done [around here](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2576/files#diff-62f51fd5acc30d2f4d9b8d1831f44cff0b6762c96a25b270bf3e0a7046d8e976L419). Now, the set of available formatting functions is defined [here](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/utils/utils.go#L80-L243), and I couldn't really find one suitable for ENS names in that list (so I assumed all cases have simply been handled by case 1). There are two potentially relevant ones:
    1. `formatAddressLong`
        -  We don't currently have and don't want long addresses here if it's not an ENS
        -  Leveraging the function's early out only if a `fromName` [was set](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2576/files#diff-62f51fd5acc30d2f4d9b8d1831f44cff0b6762c96a25b270bf3e0a7046d8e976R421) (but otherwise keep `formatEth1Address` as formatting) would feel like function misuse (also it wouldn't even work if we're not actually dealing with an ENS but a custom tag or whatever got returned, see discussion above)
        - Also that function doesn't apply the href to `/address/`. Changing this behavior would obviously affect other places which I don't think is desired. And applying the href by surrounding the result with another span kind of defeats the function's purpose, as it does almost nothing more than `formatHTML` provides.
    2. `formatAddressAsLink`
        Not a bad fit actually. Unlike in i) we could even use it for the non-ens default address case for consistency reasons. The only thing I didn't like was the additional [short address after the name](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/utils/eth1.go#L205C72-L205C80). I didn't want to affect other places again, but actually that part isn't used anyways because the function seems to be never called with a `name`.

So to sum it all up, I think there are a couple ways to solve this:
-  Define a new struct/use the generic `types.DataTableResponse` to separate backend db query data and frontend-formatted data properly, as mentioned in the initial PR comment (basically changing from method 2 to 1). That'd also be good for future changes like this, but atm it'd basically just copy the entire struct and seemed like overkill
-  Also register `utils.FormatAddress` which you linked as another template formatting function so going forward we have exactly the same behavior of ENS formatting, regardless of if method 1 or 2 is being used. Only thing I don't like too much about this is having to pass all those parameter in the html templates.. Maybe we could add another shorter version of that function (and register this one instead) which removes the `token`, `verified` (which isn't used anyways), `isContract` and `link` arguments by replacing them with default parameters
- Go with the proposed workaround solution using a new function
- Go with the existing `formatAddressAsLink` but modify it slightly (remove the short address after the ens, see above)

If we want to clean this all up, I suggest using the second option and removing `formatAddressAsLink` entirely, replacing all its uses with `utils.FormatAddress`.

D13ce (approved):
utACK

(by the way, you did not "take too much of my time" at all :) )

vladisvobo (approved):

PR #2577, opened 2023-09-27
(BIDS-2510) using the minipool contract address with the most recent update

remoterami:
### <samp>🤖 Generated by Copilot at 2296fad</samp>

Fixed a bug in `validator.go` that caused incorrect Rocketpool data to be shown on the validator page. Modified the SQL query to order and limit the results by the latest minipool and node status.


manuelsc (approved):
utACK

PR #2579, opened 2023-09-28
(NOBIDS) Remove unsued variables

D13ce:
This PR simply removes 3 variables that are not used.

### DIFF
--- a/types/templates.go
+++ b/types/templates.go
@@ -365,8 +365,6 @@ type ValidatorPageData struct {
 	ExecutedAttestationsCount                uint64
 	MissedAttestationsCount                  uint64
 	UnmissedAttestationsPercentage           float64 // missed/(executed+orphaned)
-	StatusProposedCount                      uint64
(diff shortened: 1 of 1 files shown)


types/templates.go
```diff
@@ -403,7 +401,6 @@ type ValidatorPageData struct {
 	Watchlist                                []*TaggedValidators
 	SubscriptionFlash                        []interface{}
 	User                                     *User
-	AverageAttestationInclusionDistance      float64
```
Eisei24, line 406: **minor:** There is still an error message left that uses this variable name.
It is weirdly worded as "efficiency", "effectiveness" and "average inclusion distance" are kind of used for the same values. In any case I would rewrite the error message since AverageAttestationInclusionDistance is not something you try to retrieve that would cause this error.

Eisei24 (changes requested):
Please have a look at the minor improvement we could have.

Eisei24 (approved):
lgtm

PR #2584, opened 2023-10-02
Updates on p2p.org saas info

rgb2hsl:
1. Service link is now https://p2p.org/networks/ethereum
2. Fee is changed from 10% to 5%
3. p2p.org used 50% Besu and 50% Geth, so there's no supermajority client issue in EL (if you need any proofs — please, let me know)


rgb2hsl:
Hello :) Any updates on this?

PR #2585, opened 2023-10-02
(BIDS-2451) Improve and unify error logs for subscriptions

D13ce:
This PR improves and unifies error messages in relation to bundled subscriptions.

### DIFF
--- a/handlers/user.go
+++ b/handlers/user.go
@@ -1735,6 +1735,8 @@ func MultipleUsersNotificationsSubscribe(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	errFields["jsonObjects"] = jsonObjects
+
 	if len(jsonObjects) > 100 {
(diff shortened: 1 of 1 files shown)


handlers/user.go
```diff
@@ -1723,16 +1723,20 @@ func MultipleUsersNotificationsSubscribe(w http.ResponseWriter, r *http.Request)
 		EventThreshold float64 `json:"event_threshold"`
 	}
 
+	errFields := map[string]interface{}{
+		"route": r.URL.String(),
+	}
```
LuccaBitfly, line 1728: You'll probably want to include the request body here too, since this is a POST request there will be relevant data in there too. Same goes for `func MultipleUsersNotificationsUnsubscribe`

LuccaBitfly (changes requested):

PR #2587, opened 2023-10-03
(BIDS-2428) fix email change

LuccaBitfly:
Fixes email changes, adds further security by requiring the user to confirm current password before being able to send the email change request (-> bad actor can't change email if they have access to a users session = more confidentiality = more security = more secure = more better), also changes e-mail in stripe if user is a stripe customer

### <samp>🤖 Generated by Copilot at 73e89eb</samp>

This pull request adds a feature that allows users to change their email address in the settings page, with proper validation and confirmation steps. It also improves the user interface and the code quality of the email change functionality in `handlers/user.go` and `templates/user/settings.html`.


templates/user/settings.html
```diff
@@ -402,14 +402,14 @@ <h3 class="h5">Update Email</h3>
...
-                          <input value="{{ .Subscription.Email }}" disabled inputmode="email" type="text" maxlength="100" class="form-control" autocomplete="email" id="email" name="email" />
-                          <div class="input-group-append">
-                            <button id="email-edit-button" onclick="toggleButton(this)" type="button" class="btn btn-outline-primary">Edit</button>
-                          </div>
-                        </div>
+                        <label for="password">Confirm Current Password</label>
+                        <input required type="password" minlength="5" maxlength="256" class="form-control" autocomplete="password" id="password" name="password" />
+                      </div>
+                      <div class="form-group">
+                        <label for="email">New Email address</label>
```
Eisei24, line 409: **minor:** It was written like this before too but should the first letter of `address` be capitalized to be consistent with the rest of the page?

handlers/user.go
```diff
@@ -1298,50 +1322,86 @@ func UserUpdateEmailPost(w http.ResponseWriter, r *http.Request) {
...
-	email := r.FormValue("email")
 
-	if !utils.IsValidEmail(email) {
-		session.AddFlash("Error: Invalid email format!")
+	// check if password is correct
+	formPassword := r.FormValue("password")
+
+	err = bcrypt.CompareHashAndPassword([]byte(userData.Password), []byte(formPassword))
+	if err != nil {
+		session.AddFlash("Error: Invalid email or password!")
```
Eisei24, line 1331: **minor:** This Flash message was copied from the login where it makes sense that we don't tell the user which of them is wrong.
However here the email can't be "wrong" it can just have the wrong format as the second flash message says.
So should this one just be something like `Error: Wrong password!`?

handlers/user.go
```diff
@@ -1298,50 +1322,86 @@ func UserUpdateEmailPost(w http.ResponseWriter, r *http.Request) {
...
+		session.AddFlash("Error: Invalid email format!")
+		session.Save(r, w)
 		http.Redirect(w, r, "/user/settings", http.StatusSeeOther)
 		return
-	} else if existingEmails.Email != "" {
-		session.AddFlash("Error: Email already exists please choose a unique email")
+	}
+
+	emailExists := false
+	db.FrontendWriterDB.Get(&emailExists, "EXISTS (SELECT email FROM users WHERE email = $1)", newEmail)
```
Eisei24, line 1354: **issue:** This does not appear to be valid syntax. I can only find uses of `EXISTS` in the context of `WHERE EXISTS` online.

Furthermore there is no error handling for this line.

handlers/user.go
```diff
@@ -1388,15 +1448,15 @@ func UserConfirmUpdateEmail(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var emailExists string
-	db.FrontendWriterDB.Get(&emailExists, "SELECT email FROM users WHERE email = $1", user.NewEmail)
-	if emailExists != "" {
+	emailExists := false
+	db.FrontendWriterDB.Get(&emailExists, "EXISTS (SELECT email FROM users WHERE email = $1)", user.NewEmail)
```
Eisei24, line 1452: **issue:** Same syntax issue with `EXISTS` as above.

Eisei24 (changes requested):

handlers/user.go
```diff
@@ -1283,13 +1282,38 @@ func UserUpdatePasswordPost(w http.ResponseWriter, r *http.Request) {
...
+		utils.LogError(err, "error user data for email change request", 0, map[string]interface{}{"userID": user.UserID})
+		session.AddFlash("Error: Error processing request, please try again later.")
+		session.Save(r, w)
+		http.Redirect(w, r, "/user/settings", http.StatusSeeOther)
+		return
+	}
+
+	// check if email change request is ratelimited
+	now := time.Now()
+	if userData.ConfirmTs != nil && (*userData.ConfirmTs).Add(authConfirmEmailRateLimit).After(now) {
```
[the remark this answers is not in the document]
LuccaBitfly: Wasn't aware that Go automatically dereferences pointers if they are passed to receiver functions of their type. Thx, fixed

templates/user/settings.html
LuccaBitfly: fixed

handlers/user.go
LuccaBitfly: writing "wrong password" flat out seems like an anti-pattern to me, changed it to "invalid credentials"

handlers/user.go
LuccaBitfly: fixed

handlers/user.go
LuccaBitfly: fixed

Eisei24 (commented):
The users table is inconsistent across out networks.
It should be created like this
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/03773f7aa2b023a848ab644523f103d04beb4d18/db/migrations/20230330131622_init_database.sql#L518
but currently:
* Sepolia: Missing **user_group** column
* Prater: Missing **email_change_to_value** column
* Mainnet: Missing **user_group** and **email_change_to_value** columns
* Users: Missing nothing

While only the Users db is the important one in the end the others should be updated for testing purposes and consistency.

handlers/user.go
```diff
@@ -1362,7 +1429,7 @@ func UserConfirmUpdateEmail(w http.ResponseWriter, r *http.Request) {
 		NewEmail  string    `db:"email_change_to_value"`
 	}{}
 
-	err = db.FrontendWriterDB.Get(&user, "SELECT id, email, email_confirmation_ts, email_confirmed FROM users WHERE email_confirmation_hash = $1", hash)
+	err = db.FrontendWriterDB.Get(&user, "SELECT id, email, email_confirmation_ts, email_confirmed, email_change_to_value FROM users WHERE email_confirmation_hash = $1", hash)
 	if err != nil {
 		logger.Errorf("error retreiveing email for confirmation_hash %v %v", hash, err)
```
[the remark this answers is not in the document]
LuccaBitfly: This is a really funny way to misspell retrievieng. Unfortunately fixed

LuccaBitfly:
I've been working on getting the users' tables to be consistent with each other, but it's taking more time than expected. To move forward with this issue, I'm assigning it back to you. But please take solace in the fact that I will continue to do my best to resolve the user table problem.

Eisei24 (approved):
utACK

handlers/user.go
```diff
@@ -1431,9 +1443,15 @@ func UserConfirmUpdateEmail(w http.ResponseWriter, r *http.Request) {
 
 	err = db.FrontendWriterDB.Get(&user, "SELECT id, email, email_confirmation_ts, email_confirmed, email_change_to_value FROM users WHERE email_confirmation_hash = $1", hash)
 	if err != nil {
-		logger.Errorf("error retrieving email for confirmation_hash %v %v", hash, err)
-		utils.SetFlash(w, r, authSessionName, "Error: This confirmation link is invalid / outdated.")
-		http.Redirect(w, r, "/confirmation", http.StatusSeeOther)
+		logger.Errorf("error retrieving user data for updating email; hash: %v; err: %v", hash, err)
```
D13ce, line 1446: Should use `utils.LogError`. Would require multiple changes in this file.

LuccaBitfly: Thx, changed to `utils.LogError` for all functions I touched.

handlers/user.go
```diff
@@ -1414,67 +1422,96 @@ func UserConfirmUpdateEmail(w http.ResponseWriter, r *http.Request) {
...
+		}
+		utils.SetFlash(w, r, authSessionName, "Error: This link is invalid / outdated.")
+		http.Redirect(w, r, "/user/settings", http.StatusSeeOther)
+		return
+	}
+
+	// validate data
+
+	if sessionUser.UserID != uint64(user.ID) {
+		utils.SetFlash(w, r, authSessionName, "Error: This link is invalid / outdated.")
```
Eisei24, line 1459: **minor:** Maybe a confusing flash message since as a user I would not know that the reason for the failure was that a different user was logged in.

Eisei24 (changes requested):

handlers/user.go
```diff
@@ -1414,67 +1422,96 @@ func UserConfirmUpdateEmail(w http.ResponseWriter, r *http.Request) {
...
+		ID               int64     `db:"id"`
+		Email            string    `db:"email"`
+		ConfirmTs        time.Time `db:"email_confirmation_ts"`
+		Confirmed        bool      `db:"email_confirmed"`
+		NewEmail         string    `db:"email_change_to_value"`
+		StripeCustomerId string    `db:"stripe_customer_id"`
 	}{}
 
-	err = db.FrontendWriterDB.Get(&user, "SELECT id, email, email_confirmation_ts, email_confirmed, email_change_to_value FROM users WHERE email_confirmation_hash = $1", hash)
+	err = db.FrontendWriterDB.Get(&user, "SELECT id, email, COALESCE(email_confirmation_ts, '399-01-01 BC'::timestamp), email_confirmed, COALESCE(email_change_to_value, ''), COALESCE(stripe_customer_id, '') FROM users WHERE email_confirmation_hash = $1", hash)
```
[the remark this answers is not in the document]
LuccaBitfly: Fixed

handlers/user.go
LuccaBitfly: This has security implications. If an authenticated user would get another message they would know that this is the valid hash for some other user, which we probably don't want.

Eisei24: Thank you for the clarification.

Eisei24 (approved):
utACK

handlers/user.go
```diff
@@ -1298,109 +1336,197 @@ func UserUpdateEmailPost(w http.ResponseWriter, r *http.Request) {
...
-		}
+		utils.LogError(err, "error sending email-change confirmation email", 0, map[string]interface{}{"userID": user.UserID, "newEmail": newEmail})
+		session.AddFlash(authInternalServerErrorFlashMsg)
 		session.Save(r, w)
 		http.Redirect(w, r, "/user/settings", http.StatusSeeOther)
 		return
 	}
 
-	session.AddFlash("Verification link sent to your new email " + email)
+	session.AddFlash("Verification link sent to your new email " + newEmail)
```
manuelsc, line 1395: Small ux improvement: tell the user that confirmation link is only valid for 30min

manuelsc (changes requested):

handlers/user.go
```diff
@@ -1298,109 +1336,197 @@ func UserUpdateEmailPost(w http.ResponseWriter, r *http.Request) {
...
+		session.AddFlash(authInternalServerErrorFlashMsg)
+		session.Save(r, w)
 		http.Redirect(w, r, "/user/settings", http.StatusSeeOther)
 		return
-	} else if existingEmails.Email != "" {
-		session.AddFlash("Error: Email already exists please choose a unique email")
+	}
+
+	if emailExists {
+		session.AddFlash("Error: Email already exists, please choose a unique email")
```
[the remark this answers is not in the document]
LuccaBitfly: The sign up also displays a similar error message when trying to register an existing email
image
Is this also bad practice?

handlers/user.go
```diff
@@ -1298,109 +1336,197 @@ func UserUpdateEmailPost(w http.ResponseWriter, r *http.Request) {
...
 		session.Save(r, w)
 		http.Redirect(w, r, "/user/settings", http.StatusSeeOther)
 		return
 	}
 
-	var existingEmails struct {
-		Count int
-		Email string
+	// validate new email
+	newEmail := r.FormValue("email")
```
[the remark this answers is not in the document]
LuccaBitfly: fixed, ptal

handlers/user.go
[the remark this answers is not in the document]
LuccaBitfly: fixed, ptal (haven't touched sign up though)

handlers/user.go
```diff
@@ -1298,109 +1336,197 @@ func UserUpdateEmailPost(w http.ResponseWriter, r *http.Request) {
...
 
-	_, session, err := getUserSession(r)
+	sessionUser, session, err := getUserSession(r)
 	if err != nil {
-		logger.Errorf("error retrieving session: %v", err)
+		utils.LogError(err, "error retrieving session for email update confirmation", 0, map[string]interface{}{"hash": hash})
 		http.Error(w, "Internal server error", http.StatusInternalServerError)
 		return
 	}
+	if !sessionUser.Authenticated {
```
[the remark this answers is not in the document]
LuccaBitfly: I was a bit on the fence on what to do here, since at the end of that function all user sessions (for that user) get purged and the user gets logged out. I was mainly concerned with what happens if the user clicks on the confirmation link, but is logged into a different account on that device.
With my new changes, not being authenticated or being logged into a different user account will not prevent an email change.

handlers/user.go
```diff
@@ -1298,109 +1336,197 @@ func UserUpdateEmailPost(w http.ResponseWriter, r *http.Request) {
...
+	err = db.FrontendWriterDB.Get(&emailExists, "SELECT EXISTS (SELECT email FROM users WHERE email = $1)", user.NewEmail)
+	if err != nil {
+		utils.LogError(err, "error checking if email exists", 0, map[string]interface{}{"email": user.NewEmail})
+		utils.SetFlash(w, r, authSessionName, "Error: Could not update email. Please try again later.")
+		http.Redirect(w, r, "/user/settings", http.StatusSeeOther)
+		return
+	}
+
+	if emailExists {
+		utils.SetFlash(w, r, authSessionName, "Error: Could not update email. The new email already exists, please send a new update request with a different email.")
```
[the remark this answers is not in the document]
LuccaBitfly: Maybe fixed, ptal?

handlers/user.go
```diff
@@ -1298,109 +1336,197 @@ func UserUpdateEmailPost(w http.ResponseWriter, r *http.Request) {
...
 
-	_, err = db.FrontendWriterDB.Exec(`UPDATE users SET email = $1, email_confirmation_hash = '' WHERE id = $2`, user.NewEmail, user.ID)
+	// everything is fine, update email
+	// update users email in Stripe, if user has a subscription
+	if user.StripeCustomerId != "" {
+		err = updateStripeCustomerEmail(user.StripeCustomerId, user.NewEmail)
+		if err != nil {
+			utils.LogError(err, "error updating user email in stripe", 0, map[string]interface{}{"customerID": user.StripeCustomerId, "newEmail": user.NewEmail})
+			utils.SetFlash(w, r, authSessionName, "Error: Could not update email. Please try again. If this error persists please contact <a href=\"https://support.bitfly.at/support/home\">support</a>.")
+			http.Redirect(w, r, "/user/settings", http.StatusSeeOther)
```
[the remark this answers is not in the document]
LuccaBitfly: fixed, ptal

handlers/user.go
LuccaBitfly: fixed, ptal

services/services.go
```diff
@@ -83,6 +83,9 @@ func Init() {
 	ready.Add(1)
 	go latestExportedStatisticDayUpdater(ready)
 
+	ready.Add(1)
+	go stripeEmailUpdater(ready)
```
manuelsc, line 87: The issue by placing it here is, that all of our frontend instances will run this check. So 4 mainnet instances and another one for each testnet instance we have.  
I'm note sure if we have a service besides the notification sender that runs only once though. Probably best to talk with StefanF what the best approach here is. Maybe introduce a new config flag that is only set on one instance, or a microservice approach of adding a new service for such things (we could combine it with the notification sender server and rename it to user service or something of that kind).

services/stripe_email_updater.go
```diff
@@ -0,0 +1,92 @@
...
+		if err != nil {
+			utils.LogError(err, "error getting pending users for stripe email update service", 0)
+			time.Sleep(time.Second * 10)
+			continue
+		}
+
+		// update stripe customer email
+		var updatedUsers []string
+		for _, user := range pendingUsers {
+			if user.StripeCustomerId == "" {
```
manuelsc, line 35: small nitpick, you can get rid of this check if you adjust the query and check for that in the where part

services/stripe_email_updater.go
```diff
@@ -0,0 +1,92 @@
...
+		if len(updatedUsers) > 0 {
+			_, err := db.FrontendWriterDB.Exec("UPDATE users SET stripe_email_pending = false WHERE email = ANY($1)", pq.Array(updatedUsers))
+			if err != nil {
+				utils.LogError(err, "error setting stripe_email_pending flag false for users, stripe email was updated", 0, map[string]interface{}{"emails": updatedUsers})
+				time.Sleep(time.Second * 10)
+				continue
+			}
+		}
+
+		if firstRun {
```
manuelsc, line 57: This goes together with my remark above that this shouldn't be part of the frontend instance service. The reason we have the firstrun is, that those services collect data that is needed for the frontend to show data. We don't need to do wait for that here, you can just remove it

services/stripe_email_updater.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		// update stripe customer email
+		var updatedUsers []string
+		for _, user := range pendingUsers {
+			if user.StripeCustomerId == "" {
+				utils.LogError(fmt.Errorf("user has no stripe_customer_id"), "error updating stripe customer email, this should never happen", 0, map[string]interface{}{"email": user.Email})
+				continue
+			}
+			err := updateStripeCustomerEmail(user.StripeCustomerId, user.Email)
```
manuelsc, line 39: Fortunately stripe has very generous rate limits https://stripe.com/docs/rate-limits
But just in case that sometime we exceed 100 updates per second, we should add a short sleep in between requests just to make sure in case they change it sometime and as a safety so we don't bombard them in case of an error. Maybe something like 5 requests per second on our end

services/stripe_email_updater.go
```diff
@@ -0,0 +1,92 @@
...
+	// see https://stripe.com/docs/api/customers/update
+	apiEndpoint := "https://api.stripe.com/v1/customers/" + stripeCustomerId
+
+	data := url.Values{}
+	data.Set("email", newEmail)
+	req, err := http.NewRequest(http.MethodPost, apiEndpoint, bytes.NewBufferString(data.Encode()))
+	if err != nil {
+		return fmt.Errorf("error creating email change request for stripe: %w", err)
+	}
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", utils.Config.Frontend.Stripe.SecretKey))
```
manuelsc, line 75: Just as a note to keep in mind, if this service will run outside the frontend, this config entry might not be available. Best to coordinate with StefanF again depending on how we limit the instances this service runs on

handlers/user.go
```diff
@@ -1298,177 +1335,193 @@ func UserUpdateEmailPost(w http.ResponseWriter, r *http.Request) {
...
-https://%[1]s/settings/email/%[2]s
-
-Best regards,
-
-%[1]s
-`, utils.Config.Frontend.SiteDomain, emailConfirmationHash)
-	err = mail.SendTextMail(newEmail, subject, msg, []types.EmailAttachment{})
-	if err != nil {
-		return err
+	if sessionUser.UserID == uint64(user.ID) {
```
manuelsc, line 1509: I might be wrong about that but this seems redundant. `purgeAllSessionsForUser` destroys the session store entries for the user so this special case shouldn't be necessary, right?

db/migrations/temp_add_stripe_update_column.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query - add column stripe_email_pending';
+ALTER TABLE users ADD COLUMN IF NOT EXISTS stripe_email_pending BOOLEAN NOT NULL DEFAULT FALSE;
```
manuelsc, line 4: In case we are not sure if our current dataset is 100% synced, we could set it to true for stripe_customer_id users to sync everyone initially. Thoughts?

manuelsc (changes requested):

services/stripe_email_updater.go
LuccaBitfly: I want an error to be thrown here, because if a user account is not a stripe customer but has its stripe pending flag set to true, something went very wrong, which we should be alerted to.

services/services.go
LuccaBitfly: created user service for this

services/stripe_email_updater.go
LuccaBitfly: removed, ptal

services/stripe_email_updater.go
LuccaBitfly: fixed, ptal

services/stripe_email_updater.go
LuccaBitfly: User service will have to have that config field set.

handlers/user.go
LuccaBitfly: seems so, fixed, ptal

db/migrations/temp_add_stripe_update_column.sql
LuccaBitfly: If we want to do this I would set the flag manuely for all users AFTER we tested that the implementation actually works for test users.

userService/stripe_email_updater.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+func stripeEmailUpdater() {
+	for {
+		// fetch all users with pending stripe email updates
+		var pendingUsers []struct {
+			Email            string `db:"email"`
+			StripeCustomerId string `db:"stripe_customer_id"`
+		}
+		err := db.FrontendReaderDB.Select(&pendingUsers, "SELECT email, COALESCE(stripe_customer_id. '') FROM users WHERE stripe_email_pending")
```
manuelsc, line 23: Super minor, I understand your point about throwing an error and this query is probably fine for our users, so feel free to ignore. But limiting the select on stripe_customer_id would be more performant due to the unique constraint, no full table scan needed. But we can always change this later on if we are unhappy with the performance at some point

manuelsc (approved):
lgtm utACK

PR #2588, opened 2023-10-03
(BIDS-2462) shortened long ENS names

remoterami:
### <samp>🤖 Generated by Copilot at c06db5f</samp>

Truncate very long addresses in `FormatAddressLong` function. This improves the address display on the web interface and avoids layout issues.

### DIFF
--- a/utils/eth1.go
+++ b/utils/eth1.go
@@ -253,10 +253,7 @@ func FormatHashLong(hash common.Hash) template.HTML {
 
 func FormatAddressLong(address string) template.HTML {
 	if IsValidEnsDomain(address) {
-		if len(address) > 59 {
-			return template.HTML(fmt.Sprintf("%s…%s", address[:9], address[len(address)-50:]))
-		}
-		return template.HTML(address)
+		return template.HTML(fmt.Sprintf(`<span data-truncate-middle="%s"></span>.eth`, strings.TrimSuffix(address, ".eth")))
 	}
 	address = FixAddressCasing(address)
 	if len(address) > 4 {
--- a/utils/eth1.go
+++ b/utils/eth1.go
@@ -253,6 +253,9 @@ func FormatHashLong(hash common.Hash) template.HTML {
 
 func FormatAddressLong(address string) template.HTML {
 	if IsValidEnsDomain(address) {
+		if len(address) > 59 {
+			return template.HTML(fmt.Sprintf("%s…%s", address[:9], address[len(address)-50:]))
+		}
 		return template.HTML(address)
 	}
 	address = FixAddressCasing(address)
--- a/static/js/layout.js
+++ b/static/js/layout.js
@@ -825,6 +825,41 @@ function getIncomeChartValueString(value, currency) {
   return `${trimCurrency(value)} ${currency}`
 }
 
+$("[data-truncate-middle]").each(function (item) {
+  truncateMiddle(this)
+  addEventListener("resize", (event) => {
+    truncateMiddle(this)
+  })
+  addEventListener("copy", (event) => {
+    copyDots(event, this)
+  })
+})
+
+// function for trimming an placing ellipsis in the middle when text is overflowing
+function truncateMiddle(element) {
+  element.innerHTML = element.getAttribute("data-truncate-middle")
+  const parent = element.parentElement
+  // get ratio of visible width to full width
+  const ratio = parent.offsetWidth / parent.scrollWidth
+  if (ratio < 1) {
+    const removeCount = Math.ceil((parent.innerText.length * (1 - ratio)) / 2) + 1
+    const originalText = element.getAttribute("data-truncate-middle")
+    element.innerHTML = originalText.substr(0, originalText.length / 2 - removeCount) + "…" + originalText.substr(originalText.length / 2 + removeCount)
+  }
+}
+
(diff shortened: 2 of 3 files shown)


utils/eth1.go
```diff
@@ -253,6 +253,9 @@ func FormatHashLong(hash common.Hash) template.HTML {
 
 func FormatAddressLong(address string) template.HTML {
 	if IsValidEnsDomain(address) {
+		if len(address) > 59 {
+			return template.HTML(fmt.Sprintf("%s…%s", address[:9], address[len(address)-50:]))
+		}
```
LuccaBitfly, line 258: This does not work for different screen sizes, especially mobile.
I have written a solution for this problem some time ago in the `feature/newfrontend` branch, which you can use if you want to.
I added this to `layout.js`, which is basically some logic that takes a long text and truncates as little text as is necessary from the middle and replaces it with an ellipsis:
```js
$("[data-truncate-middle]").each(function (item) {
  truncateMiddle(this)
  addEventListener("resize", (event) => {
    truncateMiddle(this)
  })
  addEventListener("copy", (event) => {
    copyDots(event, this)
  })
})

// function for trimming an placing ellipsis in the middle when text is overflowing
function truncateMiddle(element) {
  element.innerHTML = element.getAttribute("data-truncate-middle")
  const parent = element.parentElement
  // get ratio of visible width to full width
  const ratio = parent.offsetWidth / parent.scrollWidth
  if (ratio < 1) {
    const removeCount = Math.ceil((parent.innerText.length * (1 - ratio)) / 2) + 1
    const originalText = element.getAttribute("data-truncate-middle")
    element.innerHTML = originalText.substr(0, originalText.length / 2 - removeCount) + "…" + originalText.substr(originalText.length / 2 + removeCount)
  }
}

// function for inserting correct text into clipboard when copying ellipsis of text truncated with 'truncateMiddle()'
function copyDots(event, element) {
  const selection = document.getSelection()
  if (selection.toString().includes("…")) {
    const originalText = element.getAttribute("data-truncate-middle")
    const diff = originalText.length - (element.innerText.length - 1)
    const replaceText = selection.toString().replace("…", originalText.substr(originalText.length / 2 - diff / 2, diff))
    event.clipboardData.setData("text/plain", replaceText)
    event.preventDefault()
  }
}
```
(don't know if this is the most elegant solution, it works but if you can improve it pls do)
Usage would be something like:
```HTML
<span data-truncate-middle="<insert full length text>"></span>
```
So in your case probably:
```go
	if IsValidEnsDomain(address) {
		return template.HTML(fmt.Sprintf(`<span data-truncate-middle="%s"></span>.eth`, strings.TrimSuffix(address, ".eth")))
	}
```
Also make sure that the parent element of the span (located inside `ensSearch.html`) doesn't have some inline display property, which could happen if it was inside a flexbox. Result could look something like this:
image
Really small screen sizes could look like this:
image
which is probably not optimal as it would happen to normal length ENS names too, so if that happened you could implement a solution for that.

LuccaBitfly (changes requested):

utils/eth1.go
remoterami: Thanks for the detailed reply, really appreciate it!

I applied your suggestions. but was still getting situations like this:
2462_split_lines
I feel like that's related to this section
> Also make sure that the parent element of the span (located inside ensSearch.html) doesn't have some inline display property, which could happen if it was inside a flexbox.

But I'm not too sure about it as a bootstrap noob. I see that the parent div defines `class="d-md-flex"`, but removing the flex (or other classes) did not change the result.
I managed to avoid this by adding a simple `text-nowrap`; let me know if you know of a better solution.

LuccaBitfly: Really? Having it just like this:
```
    <div class="py-2 my-3 justify-content-md-between">
      <h1 class="h4 mb-1 mb-md-0"><i class="fas fa-search mr-2 position-relative"></i>Ens Search for: {{ formatAddressLong .Data.Search }}</h1>
```
the text doesn't wrap for me.

remoterami: Yes :sweat_smile:
Just tried it again (just like you posted), and on my PC and phone I get the result as described above in brave/chrome and firefox... I'm on a local testnet though, could this have an effect?

LuccaBitfly: Idk, in my browser your changes don't result in the text being truncated:
image
I would suggest the following:

- Remove the `d-md-flex` as otherwise the text doesn't truncate for me + I'm 99% it's useless and does nothing.
- Remove the `text-nowrap` and insert a fix line break after the `Ens Search for:` so that the ENS name has its own line.

Oh and I'm also noticing that it should be "ENS" instead of "Ens", right?

remoterami: Yes. With the latest commit f1507b953fd9de3f6d6e5363347c92220af476f3 it's working for me, too. As mentioned, when using `text-nowrap` I can avoid the line brake.

However when using the code from [your reply](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2588#discussion_r1349937805), the text **does** wrap for me.

I'm sorry for the confusion. If using using `text-nowrap` is a valid solution to you, feel free to resolve.

Yes correct, I'll change it to "ENS"

LuccaBitfly (approved):
lgtm

PR #2590, opened 2023-10-04
(BIDS-2516) cleaned up finalized bars on index page

remoterami:
Besides the BIDS, this does clean up some other issues related to the slot visualization on the frontpage shortly after genesis:
- A rounding behavior difference between golang and js which would show different results for the same value on the page (like 91.76% vs 91.77%)
- The finalized bar is now immediately shown if 100% global participation have been reached
- For consistency reasons there are now also 5 bars on the frontpage instead of 4, just like on `/chars/slotviz`
- Removed some debug output

However it's still not using the same implementation as `/charts/slotviz` as there are some minor differences. But that can probably also be figured out if desired.

### <samp>🤖 Generated by Copilot at df128e1</samp>

This pull request improves the slot visualization chart on the genesis page by updating the data source, the epoch range, and the formatting of the global participation rate. It also fixes a typo in an error message in `services/services.go`.


templates/index/genesis.html
LuccaBitfly, line 1: I know this is not your fault, but I don't think having a hardcoded slot viz just for the genesis period is a good idea. We already have the capability to display the slot viz on the "normal" index page and I believe this can be implemented here too.
If we only want to display the first 4 epochs on the genesis page: the slot viz usually fetches it's data from `/launchMetrics`, which I think will always contain the latest 4 epochs - not the first 4 - so I suggest we make the fetch URL configurable and create a new endpoint which only returns the desired epochs, so that it can be set for the genesis page.

db/db.go
```diff
@@ -2124,9 +2124,6 @@ func GetSlotVizData(latestEpoch uint64) ([]*types.SlotVizEpochs, error) {
 	res := []*types.SlotVizEpochs{}
 
 	for _, b := range blks {
-		if b.Globalparticipationrate == 1 && !b.Finalized {
-			b.Globalparticipationrate = 0
-		}
```
LuccaBitfly, line 2129: I believe this may be necessary since we store the global participation rate as 1.0 per default and only later update it to the real value, which results in wrong data for a short time for new epochs. Not sure if this still happens, pls check.

LuccaBitfly (changes requested):

db/db.go
remoterami: I couldn't confirm this, seems like it's initialized with 0 [here](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/db/db.go#L779).

templates/index/genesis.html
```diff
@@ -1,15 +1,19 @@
...
   </div>
   <div style="position:relative" class="card mt-3 index-stats">
     <div class="card-header pt-3">
       <div class="row">
         {{ template "networkStats" }}
       </div>
     </div>
+    <div class="mt-4">
+      <div class="w-100 my-2" id="slotsViz"></div>
+    </div>
```
LuccaBitfly, line 15: Doesn't fit within the boundaries.
image
You should rather replace the div with id `checklist` with the slot viz, as this was also where it was inserted before this PR.
```html
<div class="card-body">
  <div class="row">
    <div class="col-lg-12">
      <div class="m-2" id="slotsViz"></div>
      <h5 class="my-3">Genesis Checklist</h5>
```

templates/index/genesis.html
```diff
@@ -1,15 +1,19 @@
...
+    {{ template "SlotViz" . }}
```
LuccaBitfly, line 16: > code line above not relevant for this

The slot viz is not shown if `.Data.SlotVizData` is nil, which happens if the config is not explicitly configured to show the slot viz. This shouldn't be necessary for the genesis index page, it should show the slot viz regardless (as it did before). You could change the `index.go` handler so that `.Data.SlotVizData` is also filled when the genesis period is active.

services/configuration.go
```diff
@@ -26,7 +26,7 @@ var DefaultExplorerConfiguration types.ExplorerConfigurationMap = types.Explorer
 	ConfigurationCategorySlotViz: {
 		ConfigurationKeyVisibleFromEpoch: {Value: "0", DataType: "int"},
 		ConfigurationKeyHardforkEpoch:    {Value: "0", DataType: "int"},
-		ConfigurationKeyVisibleToEpoch:   {Value: "0", DataType: "int"},
+		ConfigurationKeyVisibleToEpoch:   {Value: "4", DataType: "int"},
```
LuccaBitfly, line 29: I'm curious: What does this do?

LuccaBitfly (changes requested):
Better than before :), but still some issues.

services/configuration.go
remoterami: It adds a more appropriate default value for new networks imo. This will cause the index page to show the slot visualizer for the first few epochs without additional effort ([see here](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/handlers/index.go#L73-L87)). That can be useful for local testnets or forks of the project, because you don't have to manually insert the configs into the database first.

templates/index/genesis.html
remoterami: I think it shouldn't be replaced, but rather added alongside. Otherwise we'd also have to adjust the selector [here](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/handlers/index.go#L91), which makes it inconsistent because for post genesis there is no checklist, thus that naming would be confusing.

I missed the boundaries issue because on a local testnet the numbers aren't that big, thanks for pointing it out :+1: 

Pushed a new solution, I hope that's fine, lmk what you think :slightly_smiling_face:

services/configuration.go
LuccaBitfly: I see, this also resolves https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2590#discussion_r1369787009, nice

templates/index/index.html
```diff
@@ -355,16 +355,13 @@
...
+      {{ if and (not .Data.Genesis) (not .Data.GenesisPeriod) }}
+        {{ template "preGenesis" $ }}
+      {{ else if .Data.GenesisPeriod }}
+        {{ template "genesis" . }}
+      {{ else }}
+        {{ template "postGenesis" .Data }}
       {{ end }}
     </div>
   </div>
-  {{ if .Data.SlotVizData }}{{ template "SlotViz" . }}{{ end }}
```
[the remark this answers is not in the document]
LuccaBitfly: After getting what the default config value in [`configuration.go`](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2590/files/98710283401953aa713b21f794a87ba3519825c0#diff-01a8648eeaa476522c14e214bfd9cc4626d6f5edc9a6fca1973144a12467ef9d) does, I understand now that `.SlotVizData` being set is necessary when `.GenesisPeriod` is active, so you could probably leave the old line, i.e. 
```go
   {{ if .Data.SlotVizData }}{{ template "SlotViz" . }}{{ end }}
```

templates/index/genesis.html
LuccaBitfly: I don't get why the selector would need to be adjusted. Your new changes look good, but the `checklist` div has no use anymore, right? I believe it can be removed, that's what I meant in the first place.

PR #2592, opened 2023-10-05
(NOBIDS) make frontend compatible with old and new bigtable schema

peterbitfly:
Introduces a new config value `v2SchemaCutOffEpoch` that defines after which epoch the new bigtable schema can be used, provides compatibility with the old schema during a potential long running schema migration.

### <samp>🤖 Generated by Copilot at 07a989f</samp>

This pull request adds a new function to query sync participation data from Postgres, refactors the validator handler to use the new function, and adds a config field to support the Bigtable schema version 2 for the Altair fork.


D13ce (approved):
utACK

PR #2595, opened 2023-10-06
(NOBIDS) fix orphaned attestation view

peterbitfly:
### <samp>🤖 Generated by Copilot at d43fb68</samp>

Changed the Bigtable query filter to fetch more attestation data for validators. This enables showing the attestation history for the last 32 epochs on the explorer website.

### DIFF
--- a/db/bigtable.go
+++ b/db/bigtable.go
@@ -1130,7 +1130,7 @@ func (bigtable *Bigtable) GetValidatorAttestationHistory(validators []uint64, st
 	res := make(map[uint64][]*types.ValidatorAttestation, len(validators))
 	resMux := &sync.Mutex{}
 
-	filter := gcp_bigtable.LatestNFilter(1)
+	filter := gcp_bigtable.LatestNFilter(32)
 
 	g, gCtx := errgroup.WithContext(ctx)
 	g.SetLimit(concurrency)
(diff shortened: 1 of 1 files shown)


D13ce (changes requested):
Your change works well but I **believe** you have missed a spot where the same change has to be applied as well, namely within `func (bigtable *Bigtable) GetValidatorMissedAttestationHistory(validators []uint64, startEpoch uint64, endEpoch uint64) (map[uint64]map[uint64]bool, error) {`:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/d43fb68c8ceb7e2e2b823d54ecd89264d75b0fc7/db/bigtable.go#L1424

Here is a real-life test-case: https://holesky.beaconcha.in/validator/1461234#attestations
For epoch 1744, the validator had to attest for slot [55832](https://holesky.beaconcha.in/slot/55832) and his attestation was included in both an orphaned slot ([55833](https://holesky.beaconcha.in/slot/55833)) and a proposed one ([55835](https://holesky.beaconcha.in/slot/55835)).

Because of the orphaned slot, the function returns a single missed attestation when called like this: `GetValidatorMissedAttestationHistory([]uint64{1461234}, 1744, 1744)`

Using 32 for the `LatestNFilter` fixes that.

D13ce (approved):
lgtm

PR #2613, opened 2023-10-17
(NOBIDS) Update Ethereum clients list info

one-three-three-seven:
### <samp>🤖 Generated by Copilot at 928c701</samp>

Update external links in `ethClientsServices.html` template. Use better sources for eth2 client and service information.


D13ce (approved):
lgtm

PR #2616, opened 2023-10-18
(NOBIDS) add slot to proposal notifications

peterbitfly:
### <samp>🤖 Generated by Copilot at e4cdb72</samp>

Add slot number to block proposal notifications. This change enhances the `services/notifications.go` file to include the slot number of each block proposal event in the notification data sent to users.


D13ce (approved):
utACK

PR #2618, opened 2023-10-18
(BIDS-2550) fixed contract invocation detection by adding metadata column

remoterami:
### <samp>🤖 Generated by Copilot at 4ba935d</samp>

This pull request adds a feature to detect and display contract invocation status for eth1 transactions. It adds a new field `invokes_contract` to the `Transaction` message and the transaction data in the bigtable. It also updates the handlers and the rpc code to use this field and show the contract status in the execution block and transaction pages.


rpc/erigon.go
```diff
@@ -283,6 +283,11 @@ func (client *ErigonClient) GetBlock(number int64, traceMode string) (*types.Eth
 					}
 
 					c.Transactions[trace.TransactionPosition].Itx = append(c.Transactions[trace.TransactionPosition].Itx, tracePb)
+					if len(trace.TraceAddress) == 0 {
```
Buttaa, line 286: Rare comment by me on code: I don't know if we want to use Erigon specific API. If we ever consider changing the EL this might break stuff. If I interpret this incorrectly, just ignore me :)

RomiRand: That's not touched, just using another field of the reply. We are already using an Erigon specific endpoint [here](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2618/files#diff-a36de22b09d32f97df1d4b745a6cc61e3242d7693494334a65dec6747032932aL228), but there's a fallback to geth in place [here](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2618/files#diff-a36de22b09d32f97df1d4b745a6cc61e3242d7693494334a65dec6747032932aL290). Both cases should have the same result.

Eisei24 (approved):
utACK

types/eth1.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.27.1
+// 	protoc-gen-go v1.31.0
```
D13ce, line 3: **Question:** Why was this required?

types/eth1.pb.go
```diff
@@ -2414,6 +2477,18 @@ func file_eth1_proto_init() {
...
+			switch v := v.(*Eth1ERC721Indexed); i {
+			case 0:
+				return &v.state
+			case 1:
+				return &v.sizeCache
+			case 2:
+				return &v.unknownFields
+			default:
+				return nil
+			}
```
D13ce, line 2489: **Question:** Do you know why this was added? From my pov, this should have been here before. Just curious.

types/eth1.proto
```diff
@@ -161,6 +166,7 @@ message Eth1TransactionIndexed {
     bytes tx_fee = 8;
     bytes gas_price = 9;
     bool is_contract_creation = 10;
+    // unused, should mark reserved!
```
D13ce, line 169: **Nitpick:** This only relates to `invokes_contract`, right? Personally I would change this to
```suggestion
    // invokes_contract is unused, should mark reserved!
```
because for example here https://github.com/gobitfly/eth2-beaconchain-explorer/blob/27e923389afeca26ed17417fe933f7999048042d/types/eth1.pb.go#L1231-L1233 it looks like everything below that comment might be unused.

Also applies to https://github.com/gobitfly/eth2-beaconchain-explorer/blob/27e923389afeca26ed17417fe933f7999048042d/types/eth1.proto#L200.

db/bigtable_eth1.go
```diff
@@ -3594,6 +3734,184 @@ func (bigtable *Bigtable) GetAddressNames(addresses map[string]string) error {
...
+// returns account state after the given execution state
+// -1 is latest (e.g. "txIdx" = -1 returns the contract state after execution of "block", "block" = -1 returns the state at chain head)
+func (bigtable *Bigtable) GetAddressIsContractAt(requests []isContractAtRequest) ([]types.ContractInteractionType, error) {
+	results := make([]types.ContractInteractionType, len(requests))
+	if len(requests) == 0 {
+		return results, nil
+	}
+
+	// get histories
+	histories := make(map[string][]isContractInfo, 0)
```
D13ce, line 3800: **Minor:** You should set the capacity to `len(requests)`.

db/bigtable_eth1.go
```diff
@@ -3594,6 +3734,184 @@ func (bigtable *Bigtable) GetAddressNames(addresses map[string]string) error {
...
+
+	// evaluate requests; CONTRACT_NONE is default
+	for i, request := range requests {
+		if _, ok := histories[request.address]; !ok {
+			continue
+		}
+		if histories[request.address] == nil {
+			continue
+		}
+		history := histories[request.address]
```
D13ce, line 3817: **Nitpick:** Why not write
```suggestion
		if history, ok := histories[request.address]; !ok {
			continue
		}
		if history == nil {
			continue
		}
```
?

db/bigtable_eth1.go
```diff
@@ -1252,6 +1280,11 @@ func (bigtable *Bigtable) TransformItx(blk *types.Eth1Block, cache *freecache.Ca
 	bulkData = &types.BulkMutations{}
 	bulkMetadataUpdates = &types.BulkMutations{}
 
+	ContractUpdateWrites := &types.BulkMutations{
+		Keys: make([]string, 0),
+		Muts: make([]*gcp_bigtable.Mutation, 0),
+	}
```
D13ce, line 1286: **Question:** Why didn't you use `bulkData` for that? By NOT using it, the `ContractUpdateWrites` keys are not stored via `SaveBlockKeys`. Is that intentional?

db/bigtable_eth1.go
```diff
@@ -1960,7 +2021,7 @@ func (bigtable *Bigtable) TransformWithdrawals(block *types.Eth1Block, cache *fr
 	return bulkData, bulkMetadataUpdates, nil
 }
 
-func (bigtable *Bigtable) GetEth1TxForAddress(prefix string, limit int64) ([]*types.Eth1TransactionIndexed, string, error) {
+func (bigtable *Bigtable) GetEth1TxForAddress(prefix string, limit int64) ([]*types.Eth1TransactionIndexed, []string, error) {
```
D13ce, line 2024: **Nitpick:** Should now be called
```suggestion
func (bigtable *Bigtable) GetEth1TxsForAddress(prefix string, limit int64) ([]*types.Eth1TransactionIndexed, []string, error) {
```

db/bigtable_eth1.go
```diff
@@ -2113,11 +2174,30 @@ func (bigtable *Bigtable) GetAddressTransactionsTableData(address []byte, search
...
 		return nil, err
 	}
 
+	idxs := make([]int64, len(keys))
+	for i, k := range keys {
+		tx_idx, err := strconv.Atoi(strings.Split(k, ":")[6])
+		if err != nil {
+			return nil, fmt.Errorf("error parsing Eth1InternalTransactionIndexed tx index: %v", err)
+		}
+		tx_idx = 10000 - tx_idx
```
D13ce, line 2188: **Question:** This hardcoded 10000 should be a constant for better readability. Also applies to https://github.com/gobitfly/eth2-beaconchain-explorer/blob/133e51a61a11bab1dd3fce36676e2a70befe7ff1/db/bigtable_eth1.go#L2647.

If I am not mistaken, this comes from places like here: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/133e51a61a11bab1dd3fce36676e2a70befe7ff1/db/bigtable_eth1.go#L1049
Those then of course should also use the new constant.

WDYT?

handlers/eth1Block.go
```diff
@@ -182,13 +188,18 @@ func GetExecutionBlockPageData(number uint64, limit int) (*types.Eth1BlockPageDa
 			}
 		}
 
+		var isContractInteraction types.ContractInteractionType
```
D13ce, line 191: **Minor:** The name `isContractInteraction` is misleading as it suggest a boolean value. This is more than true or false though, this holds an interaction type.
This applies to various other code parts too.

Similar to this, `GetAddressIsContractAtBlock` for example is kind of misleading too. A better name would be something like `GetAddressContractInteractionsAtBlock`.
This also applies to various other code parts too.

D13ce (commented):
@remoterami , as discussed in DMs, I have also reviewed this a bit and left some questions and minor suggestions. ptal

Eisei24 (commented):
Please have a look at the comments from me and @D13ce.

types/eth1.pb.go
remoterami: There's no particular reason, I just installed the latest version. I'll downgrade my protoc-gen-go again

types/eth1.pb.go
remoterami: Oh that's weird indeed.. I have no idea honestly

types/eth1.proto
remoterami: That's correct, I'll change it :)

Should we actually mark the field [reserved](https://protobuf.dev/programming-guides/proto3/#fieldreserved) though? So we have cleaned this up and prevent another dev in the future from relying on the value being set and correct

db/bigtable_eth1.go
remoterami: Well, to nitpick on the nitpick, because variables are only defined in the scope in which they are declared :stuck_out_tongue: so `history` wouldn't be known after the first `if`. But by moving the if I can still simplify this a lot, thanks

db/bigtable_eth1.go
remoterami: Oh yeah that's correct.. I forgot about re-orgs, my bad, thank you :sweat_smile: I'll take care of that

db/bigtable_eth1.go
remoterami: Yes, totally! I'm also a big proponent of avoiding all "magic numbers" everywhere. Didn't do it because, well, if that's how it's done I thought  I might shouldn't change it - but I'll gladly do.

handlers/eth1Block.go
Eisei24: The names of the methods were changed but we may also want to rename the returned variable name in that case.
E.g. it was
`txIsContractList, err := BigtableClient.GetAddressIsContractAtTransactions(transactions, idxs)`
now it is
`txIsContractList, err := BigtableClient.GetAddressContractInteractionsAtTransactions(transactions, idxs)`
with `txIsContractList` remaining the same.

@D13ce WDYT?

db/bigtable_eth1.go
```diff
@@ -4001,7 +4014,7 @@ func (bigtable *Bigtable) GetBlockKeys(blockNumber uint64, blockHash []byte) ([]
 	}
 
 	if row == nil {
-		return nil, fmt.Errorf("keys for block %v not found", blockNumber)
+		return make([]string, 0), nil
```
Eisei24, line 4017: **question/issue:** What was your intended behaviour here?

Before we threw an error if the block which was supposed to be deleted could not be found (i.e. no rows) but now we don't.

**Side note:** I see that the result is then immediately used to check if the length is zero.
For that a nil slice or an empty slice makes no difference, both return 0.
Nil slice and empty slice are only different for json encoding.

Was the reason to remove the error so that there is no confusion because you already deleted the contract states info from the bigtable first?
If that is the case why not delete the normal info that was already in `DeleteBlock` first and then delete the contract states?
In that case if `GetBlockKeys` throws this error nothing gets deleted for the block.

Eisei24 (changes requested):
The behaviour in `DeleteBlock` is still the most important point of discussion.

types/eth1.proto
D13ce: @remoterami , I honestly don't have a strong opinion on that. Feel free to either resolve this comment without any further changes or to implement the changes.

handlers/eth1Block.go
D13ce: Jeah, that should be consistent. We should change that.

db/bigtable_eth1.go
```diff
@@ -1142,19 +1183,51 @@ func (bigtable *Bigtable) TransformItx(blk *types.Eth1Block, cache *freecache.Ca
...
+			return nil, nil, fmt.Errorf("unexpected number of transactions in block expected at most %d but got: %v, tx: %x", TX_PER_BLOCK_LIMIT-1, i, tx.GetHash())
 		}
-		iReversed := reversePaddedIndex(i, 10000)
+		iReversed := reversePaddedIndex(i, TX_PER_BLOCK_LIMIT)
 
-		for j, idx := range tx.GetItx() {
-			if j > 999999 {
-				return nil, nil, fmt.Errorf("unexpected number of internal transactions in block expected at most 999999 but got: %v, tx: %x", j, tx.GetHash())
+		for j, itx := range tx.GetItx() {
+			if j >= ITX_PER_TX_LIMIT {
```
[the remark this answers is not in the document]
remoterami: Yes, I'm pretty sure `999999` was a typo, otherwise [reversePaddedIndex(j, 100000)](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2618/files/a076907a039c3a3e9073a222a4a2128f0833262c#diff-143a07e5aa31442d87e5dd9584d1525fc98e8b2fef161a08af276d65f60184acL1155) wouldn't really make sense.
Using the constants already paying off :smile:

db/bigtable_eth1.go
remoterami: The motivation for this change was to be able to handle empty blocks correctly. So `the block which was supposed to be deleted could not be found (i.e. no rows)` is not really always true I think, and this was my attempt of dealing with it. While not relevant for the main networks, it still is for the local testnet and potentially other forks with smaller communities.

Note that the block itself does get deleted anyways (see below though :sweat_smile:), this is only about [keys that were written by this block (entities & indices](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2618/files#diff-143a07e5aa31442d87e5dd9584d1525fc98e8b2fef161a08af276d65f60184acR4076), of which there might be none.

But after looking at it again, there's still an issue here: Because of the zero-length check early-out in `DeleteBlock`, we actually never [delete the block](https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2618/files#diff-143a07e5aa31442d87e5dd9584d1525fc98e8b2fef161a08af276d65f60184acR4098). So I'll remove that early-out.
<br />
Just a side note (as the premise isn't true), but just to clarify:
> why not delete the normal info that was already in DeleteBlock first and then delete the contract states?

Because I think removing the block from `tableBlocks` feels like the finishing move of that function, confirming its successful deletion. If anything else before that has failed, that step should not be executed but rather there needs to be some investigation what's wrong.
<br />
I could also be missing something of course; Lmk if you agree and this answers your question, or if there's anything else to be taken care of :)

Eisei24: You mean empty blocks like this?
https://beaconcha.in/block/15596410#overview

Don't we still save block keys even for such empty blocks?
We save them here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/3744839a466c64bfe48ef4dbc9f0e4b50f8cc9a3/db/bigtable_eth1.go#L694
and get data from all `transform` methods
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/3744839a466c64bfe48ef4dbc9f0e4b50f8cc9a3/db/bigtable_eth1.go#L679
The list of them is here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/a076907a039c3a3e9073a222a4a2128f0833262c/cmd/eth1indexer/main.go#L173
which includes
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/3744839a466c64bfe48ef4dbc9f0e4b50f8cc9a3/db/bigtable_eth1.go#L761
for generic blocks information like its hash.

Eisei24 (commented):
Please check my block keys comment.

db/bigtable_eth1.go
remoterami: I reverted this change; As discussed, I misinterpreted the consequences of another issue related to general reorg handling.

db/bigtable_eth1.go
```diff
@@ -1154,6 +1154,54 @@ func decodeIsContractUpdateTs(ts gcp_bigtable.Timestamp) (block_number, tx_idx,
 	return block_number, tx_idx, trace_idx
 }
 
+func (bigtable *Bigtable) TransformContract(blk *types.Eth1Block, cache *freecache.Cache) (bulkData *types.BulkMutations, bulkMetadataUpdates *types.BulkMutations, err error) {
+	bulkData = &types.BulkMutations{}
+	bulkMetadataUpdates = &types.BulkMutations{}
+	ContractUpdateWrites := &types.BulkMutations{}
```
Eisei24, line 1160: **nitpick:** `ContractUpdateWrites` should have a lowercase first letter, `contractUpdateWrites`.

Eisei24 (approved):
utACK

A small nitpick can still be adapted.

Eisei24 (dismissed):
lgtm, tested for tx and itx of address 0x70f5d9a2716aa99ad24b7461167e57906c7a6789 on Prater.
utACK for various edge cases and contracts that suicided.

db/bigtable_eth1.go
```diff
@@ -4048,6 +4455,18 @@ func (bigtable *Bigtable) GetMethodLabel(id []byte, invokesContract bool) string
 	return method
 }
 
+// get a method label for its byte signature with defaults
+func (bigtable *Bigtable) GetAddressLabel(id string, invoke_overwrite types.ContractInteractionType) string {
+	switch invoke_overwrite {
+	case types.CONTRACT_CREATION:
+		return "Contract Creation"
+	case types.CONTRACT_DESTRUCTION:
+		return "Contract Destruction"
```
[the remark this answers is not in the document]
remoterami: I changed the max column width to address this; As we're carefully truncating every output anyways and the number of affected tables is limited, I thinks that's justified instead of thinking about alternative wording.

Eisei24 (approved):
lgtm

Eisei24 (commented):
Only a minor issue for this BIDS itself.

However as discussed we have an issue with the "itx for a tx" page, e.g.
https://prater.beaconcha.in/tx/0xebc8edfcd300271d458da454b3d63e212031e8d860c73322fceffffb988525f4#internal-txns
https://goerli.etherscan.io/tx/0xebc8edfcd300271d458da454b3d63e212031e8d860c73322fceffffb988525f4#internal
where the "Type Trace Address" column is empty and the Advanced mode does nothing.

Please fix this here or leave a comment on how you want to handle this bug.

Eisei24 (changes requested):
I found the issue, please fix.

Eisei24 (approved):
lgtm

PR #2626, opened 2023-10-19
(NOBIDS) Show correct max effective balance for deposited valis

D13ce:
This PR causes the validator page to show the correct max effective balance for validators that are "deposited".


Eisei24 (approved):
lgtm

PR #2628, opened 2023-10-20
(BIDS-2586) Use correct formatting function for total balance

D13ce:
This PR causes the total balance on the dashboard to use the correct formatting function.

### DIFF
--- a/handlers/common.go
+++ b/handlers/common.go
@@ -318,8 +318,7 @@ func GetValidatorEarnings(validators []uint64, currency string) (*types.Validato
 	earnings.LastWeekFormatted = utils.FormatIncomeClEl(earnings.Income7d, currency)
 	earnings.LastMonthFormatted = utils.FormatIncomeClEl(earnings.Income31d, currency)
 	earnings.TotalFormatted = utils.FormatIncomeClEl(earnings.IncomeTotal, currency)
-	// earnings.TotalChangeFormatted = utils.FormatIncome(income.ClIncomeTotal+currentDayClIncome+int64(totalDeposits), currency, true)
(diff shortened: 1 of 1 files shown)


handlers/common.go
```diff
@@ -318,8 +318,7 @@ func GetValidatorEarnings(validators []uint64, currency string) (*types.Validato
 	earnings.LastWeekFormatted = utils.FormatIncomeClEl(earnings.Income7d, currency)
 	earnings.LastMonthFormatted = utils.FormatIncomeClEl(earnings.Income31d, currency)
 	earnings.TotalFormatted = utils.FormatIncomeClEl(earnings.IncomeTotal, currency)
-	// earnings.TotalChangeFormatted = utils.FormatIncome(income.ClIncomeTotal+currentDayClIncome+int64(totalDeposits), currency, true)
-	earnings.TotalBalance = utils.FormatElCurrency(totalBalance, currency, 5, true, false, false)
+	earnings.TotalBalance = "<b>" + utils.FormatClCurrency(totalBalance, currency, 5, true, false, false) + "</b>"
```
Eisei24, line 321: **question:** Should the text be changed to bold?
It is not bold for the balances of the individual validators.

Eisei24 (commented):
It works but I have a question.

handlers/common.go
D13ce: The dashboard on mainnet that runs on an older version has this value bold, see for example https://beaconcha.in/dashboard?validators=142321
image

I did not want to change the behavior so I added the <b>.

Eisei24 (approved):
lgtm

PR #2630, opened 2023-10-23
(BIDS-2411) Demote tx pending error to info

D13ce:
This PR demotes the error for a pending tx (that is not even part of the mempool yet) to an info log entry. It further more improves error logging in this context.


LuccaBitfly (approved):
utAck

PR #2631, opened 2023-10-23
(BIDS-2591) Improve return value for dashboard effectiveness

D13ce:
This PR causes `/dashboard/data/effectiveness` to return 204 (no content) if no active validator is part of the query. Before, 400 (bad request) was returned which is not correct as a dashboard for only an exited validator is still valid.

### DIFF
--- a/handlers/dashboard.go
+++ b/handlers/dashboard.go
@@ -988,7 +988,8 @@ func DashboardDataEffectiveness(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if len(activeValidators) == 0 {
-		w.WriteHeader(http.StatusNoContent)
+		// valid 200 response with empty data
+		w.Write([]byte(`{}`))
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


handlers/dashboard.go
```diff
@@ -988,7 +988,7 @@ func DashboardDataEffectiveness(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if len(activeValidators) == 0 {
-		http.Error(w, "Invalid query", http.StatusBadRequest)
+		w.WriteHeader(http.StatusNoContent)
```
LuccaBitfly, line 991: Incorrect usage of `StatusNoContent`. Also I'm not sure `w.WriteHeader` actually produces a response, just that it changes what the response code will be once it does happen, pls confirm.

LuccaBitfly (changes requested):

handlers/dashboard.go
D13ce: Thanks for your valuable input, I have adapted the code based on your suggestions.

LuccaBitfly: Thanks for your valuable response, I have resolved this conversation based on your adaptations.

LuccaBitfly (approved):
lgtm

PR #2647, opened 2023-10-27
(BIDS-2567) Check if validator index is availablefor stats

D13ce:
This PR adds a check for the given validator index and shows the validator not found page if the index is not available yet.

### DIFF
--- a/handlers/validator.go
+++ b/handlers/validator.go
@@ -1887,8 +1887,9 @@ func ValidatorStatsTable(w http.ResponseWriter, r *http.Request) {
 	} else {
 		// Request came with a validator index number
 		index, err = strconv.ParseUint(vars["index"], 10, 64)
-		// Request is not a valid index number
+
 		if err != nil || index > math.MaxInt32 { // index in postgres is limited to int
+			// Request is not a valid index number
 			logger.Warnf("error parsing validator index: %v", err)
 			validatorNotFound(data, w, r, vars, "/stats")
 			return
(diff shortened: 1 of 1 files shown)


Eisei24 (commented):
Please have a look at my suggestion.

D13ce:
> Please have a look at my suggestion.

@Eisei24, I would love to, but I don't see any :smile:

LuccaBitfly (changes requested):
unacceptable

handlers/validator.go
```diff
@@ -1897,6 +1898,17 @@ func ValidatorStatsTable(w http.ResponseWriter, r *http.Request) {
...
+	// Check if validator index is available
+	var dbIndex uint64
+	err = db.ReaderDb.Get(&dbIndex, "SELECT validatorindex FROM validators WHERE validatorindex = $1", index)
+	if err != nil {
+		if err != sql.ErrNoRows {
+			utils.LogError(err, "error checking for index in validators table", 0, errFields)
+		}
+		validatorNotFound(data, w, r, vars, "/stats")
+		return
+	}
```
Eisei24, line 1910: **suggestion:** Since we only want to check whether the index exists we can use the EXISTS keyword.
```suggestion
	var doesIndexExist bool
	err = db.ReaderDb.Get(&doesIndexExist, "SELECT EXISTS (SELECT validatorindex FROM validators WHERE validatorindex = $1)", index)
	if err != nil || !doesIndexExist {
		if err != nil {
			utils.LogError(err, "error checking for index in validators table", 0, errFields)
		}
		validatorNotFound(data, w, r, vars, "/stats")
		return
	}
```

Eisei24 (commented):
Sorry about that, I wrote the lost comment again.

handlers/validator.go
```diff
@@ -1887,8 +1887,9 @@ func ValidatorStatsTable(w http.ResponseWriter, r *http.Request) {
 	} else {
 		// Request came with a validator index number
 		index, err = strconv.ParseUint(vars["index"], 10, 64)
-		// Request is not a valid index number
+
 		if err != nil || index > math.MaxInt32 { // index in postgres is limited to int
```
LuccaBitfly, line 1891: also consider:
```go
        index, err = strconv.ParseUint(vars["index"], 10, 31)
        if err != nil {
```
See https://go.dev/play/p/ARwkAt0Ey8S.

Eisei24 (approved):
lgtm

LuccaBitfly (approved):
acceptable

PR #2650, opened 2023-10-28
(NOBIDS) remove slots cache as it is no longer required

peterbitfly:
copilot:summary

### DIFF
--- a/rpc/lighthouse.go
+++ b/rpc/lighthouse.go
@@ -33,8 +33,6 @@ type LighthouseClient struct {
 	endpoint            string
 	assignmentsCache    *lru.Cache
 	assignmentsCacheMux *sync.Mutex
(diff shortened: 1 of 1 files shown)


D13ce (approved):
lgtm

One addition though: I have checked all uses of `func (lc *LighthouseClient) GetBlockBySlot(slot uint64)` and stumbled upon `func (lc *LighthouseClient) GetNewBlockChan()`.
Currently, this is not used and I suggest to remove it.

peterbitfly:
> lgtm
> 
> One addition though: I have checked all uses of `func (lc *LighthouseClient) GetBlockBySlot(slot uint64)` and stumbled upon `func (lc *LighthouseClient) GetNewBlockChan()`. Currently, this is not used and I suggest to remove it.

Done

PR #2653, opened 2023-10-30
(BIDS-2369) add total withdrawal columns

LuccaBitfly:
Same as https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2653 but without query optimizations

### <samp>🤖 Generated by Copilot at 959551c</samp>

This pull request adds support for exporting and tracking the total number and amount of withdrawals by validators in the `validator_stats` table. It introduces a new command `export-stats-totals` to the misc tool, which allows selecting and exporting specific columns from the table. It also adds two new columns (`withdrawals_total` and `withdrawals_amount_total`) to the table and the corresponding fields and db tags to the `ValidatorStatsTableDbRow` struct. It includes a SQL migration file to add and drop the new columns.


Eisei24 (approved):
lgtm

PR #2657, opened 2023-10-31
Bids 2622/fix address page with many tokens

D13ce:
This PR limits the amount of ERC20 tokens shown on the frontend's address page and returned via the `ApiEth1Address` API endpoint.
It also implements a new API endpoint to get all ERC20 tokens for an address via pagination: `ApiEth1AddressERC20Tokens`


cmd/explorer/main.go
```diff
@@ -311,11 +311,7 @@ func main() {
 		apiV1Router.HandleFunc("/execution/{addressIndexOrPubkey}/produced", handlers.ApiETH1AccountProducedBlocks).Methods("GET", "OPTIONS")
 
 		apiV1Router.HandleFunc("/execution/address/{address}", handlers.ApiEth1Address).Methods("GET", "OPTIONS")
-		apiV1Router.HandleFunc("/execution/address/{address}/transactions", handlers.ApiEth1AddressTx).Methods("GET", "OPTIONS")
```
peterbitfly, line 314: why do you remove the handlers here?

db/bigtable_eth1.go
```diff
@@ -3273,9 +3297,16 @@ func (bigtable *Bigtable) GetMetadataForAddress(address []byte) (*types.Eth1Addr
...
 				})
-			}
 
-			if column.Column == ACCOUNT_METADATA_FAMILY+":"+ACCOUNT_COLUMN_NAME {
+				if !isNativeEth {
+					tokenCount++
+
+					// check if token limit is reached (requires that ETH balance is always first)
+					if tokenCount >= limit {
+						break
```
peterbitfly, line 3306: I am not sure if it is ok to break here, as it might cause the `ret.Name` field not to be filled?

db/bigtable_eth1.go
```diff
@@ -3286,22 +3317,9 @@ func (bigtable *Bigtable) GetMetadataForAddress(address []byte) (*types.Eth1Addr
...
-			priceJ = decimal.NewFromBigInt(new(big.Int).SetBytes(ret.Balances[j].Metadata.Price), 0)
-		}
-
-		mulI := decimal.NewFromFloat(float64(10)).Pow(decimal.NewFromBigInt(new(big.Int).SetBytes(ret.Balances[i].Metadata.Decimals), 0))
-		mulJ := decimal.NewFromFloat(float64(10)).Pow(decimal.NewFromBigInt(new(big.Int).SetBytes(ret.Balances[j].Metadata.Decimals), 0))
-		mkI := priceI.Mul(decimal.NewFromBigInt(new(big.Int).SetBytes(ret.Balances[i].Balance), 0).DivRound(mulI, 18))
-		mkJ := priceJ.Mul(decimal.NewFromBigInt(new(big.Int).SetBytes(ret.Balances[j].Balance), 0).DivRound(mulJ, 18))
-
-		return mkI.Cmp(mkJ) >= 0
+		return bytes.Compare(ret.Balances[i].Token, ret.Balances[j].Token) < 0
```
peterbitfly, line 3322: Shouldn't bigtable return the columns anyway in sorted order?

cmd/explorer/main.go
D13ce: As far as I can tell, those where not used anywhere.

db/bigtable_eth1.go
D13ce: For every token we request the metadata within a unique go-routine. Those finish in an undefined order so sorting is needed.

db/bigtable_eth1.go
D13ce: Good point, fixed.

peterbitfly:
utACK

peterbitfly (approved):

PR #2659, opened 2023-11-02
(NOBIDS) Send server error instead of user error

D13ce:
This PR causes the API endpoints `/execution/address/{address}` and `/execution/address/{address}/erc20tokens` to return 500 instead of 400 when metadata could not be fetched.
Furthermore, `ECR20TokensPerAddressLimit` is changed to const.


LuccaBitfly (approved):
lgtm

PR #2660, opened 2023-11-02
(NOBIDS) fix stats export tool end

LuccaBitfly:
### <samp>🤖 Generated by Copilot at 3e573c0</samp>

Fix off-by-one error in validator batch calculation and refactor validator ranking logic. This improves the `cmd/misc/main.go` script that ranks validators by various metrics.


D13ce (approved):
utACK

PR #2671, opened 2023-11-06
(BIDS-2547) added ❗to failed transactions

remoterami:
### <samp>🤖 Generated by Copilot at ab52558</samp>

This pull request adds a new `FormatTransactionHash` function to the `utils` package and updates various functions and pages to use it. The function formats eth1 transaction hashes with a link and a warning symbol if the transaction failed or reverted. The purpose is to improve the user interface and feedback for eth1 transactions on the explorer website.


LuccaBitfly (approved):
Maybe a tooltip for the ! would be nice? Otherwise lgtm

PR #2682, opened 2023-11-09
Fix wrong sync committee participation association

manuelsc:
Note that this PR restores functionality of the sync committee stats and does not aim to fully support the edge case of a validator being part of the same sync committee twice as this requires a wider refactor of how we handle sync committees on the frontend. Will create a jira task for that special case.

Main culprit of current approach was that the de duplication on statistics exporter level shuffled the sync committee index. Removed de duplication in this level since a validator can be elected into the same sync committee multiple times and kept current frontend handling (which does not handle this case at all atm, but so far this hasn't occurred on mainnet or prater yet.)

**Requires a re-export via misc tool, command:**
```
misc \
--command=export-sync-committee-periods \
--dry-run=false \
--config path_to_config.yml \
&& misc \
--command=export-sync-committee-validator-stats \
--dry-run=false \
--config path_to_config.yml \
&& misc \
--command=export-stats-totals \
--columns=missed_attestations_total,participated_sync_total,missed_sync_total,orphaned_sync_total \
--day-start=0 \
--day-end=0
```

_Fun Fact: Sync period 827 is utterly wrong on prater db, every one of the 512 validators has a second entry with a different committee indices. Should be fixed via re-export also._

### <samp>🤖 Generated by Copilot at b3baa4a</samp>

This pull request fixes various bugs and inconsistencies in the sync committee data for validators, both in the exporter and the API handler. It also improves the SQL queries and adds comments to make the code more clear and reliable.

### DIFF
--- a/cmd/misc/main.go
+++ b/cmd/misc/main.go
@@ -55,7 +55,7 @@ var opts = struct {
 
 func main() {
 	configPath := flag.String("config", "config/default.config.yml", "Path to the config file")
-	flag.StringVar(&opts.Command, "command", "", "command to run, available: updateAPIKey, applyDbSchema, initBigtableSchema, epoch-export, debug-rewards, debug-blocks, clear-bigtable, index-old-eth1-blocks, update-aggregation-bits, historic-prices-export, index-missing-blocks, export-epoch-missed-slots, migrate-last-attestation-slot-bigtable, export-genesis-validators, update-block-finalization-sequentially, nameValidatorsByRanges, export-stats-totals, export-sync-committee, export-sync-committee-ph2")
+	flag.StringVar(&opts.Command, "command", "", "command to run, available: updateAPIKey, applyDbSchema, initBigtableSchema, epoch-export, debug-rewards, debug-blocks, clear-bigtable, index-old-eth1-blocks, update-aggregation-bits, historic-prices-export, index-missing-blocks, export-epoch-missed-slots, migrate-last-attestation-slot-bigtable, export-genesis-validators, update-block-finalization-sequentially, nameValidatorsByRanges, export-stats-totals, export-sync-committee-periods, export-sync-committee-validator-stats")
 	flag.Uint64Var(&opts.StartEpoch, "start-epoch", 0, "start epoch")
 	flag.Uint64Var(&opts.EndEpoch, "end-epoch", 0, "end epoch")
 	flag.Uint64Var(&opts.User, "user", 0, "user id")
@@ -362,10 +362,10 @@ func main() {
 		}
 	case "export-stats-totals":
 		exportStatsTotals(opts.Columns, opts.StartDay, opts.EndDay, opts.DataConcurrency)
-	case "export-sync-committee":
-		exportSyncCommittee(rpcClient, bt, opts.StartDay, opts.EndDay, opts.DryRun, false)
-	case "export-sync-committee-ph2":
-		exportSyncCommittee(rpcClient, bt, opts.StartDay, opts.EndDay, opts.DryRun, true)
+	case "export-sync-committee-periods":
+		exportSyncCommitteePeriods(rpcClient, bt, opts.StartDay, opts.EndDay, opts.DryRun)
+	case "export-sync-committee-validator-stats":
+		exportSyncCommitteeValidatorStats(rpcClient, bt, opts.StartDay, opts.EndDay, opts.DryRun, true)
 	case "fix-exec-transactions-count":
 		err = fixExecTransactionsCount()
 	default:
@@ -1241,7 +1241,7 @@ OUTER:
 	logrus.Infof("finished all exporting stats totals for columns '%v' for days %v - %v, took %v", columns, dayStart, dayEnd, time.Since(start))
 }
 
-func exportSyncCommittee(rpcClient rpc.Client, bt *db.Bigtable, startDay, endDay uint64, dryRun, skipPhase1 bool) {
+func exportSyncCommitteePeriods(rpcClient rpc.Client, bt *db.Bigtable, startDay, endDay uint64, dryRun bool) {
 	var currEpoch = uint64(0)
 
 	firstPeriod := utils.SyncPeriodOfEpoch(utils.Config.Chain.ClConfig.AltairForkEpoch)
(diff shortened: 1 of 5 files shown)


handlers/api.go
LuccaBitfly, line 1: The queries you touched here have to be reworked in the follow-up BIDS you talked about in the description, right?

LuccaBitfly (changes requested):

exporter/sync_committees.go
```diff
@@ -75,66 +117,36 @@ func exportSyncCommitteeAtPeriod(rpcClient rpc.Client, p uint64) error {
...
 	}
 
-	return tx.Commit()
+	return result, nil
+}
+
+type SyncCommittee struct {
+	Period         uint64 `json:"period"`
+	ValidatorIndex uint64 `json:"validatorindex"`
+	CommitteeIndex uint64 `json:"committeeindex"`
```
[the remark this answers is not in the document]
manuelsc: No specific reason, figured we might need them at some point if we rewrite the frontend.

exporter/sync_committees.go
```diff
@@ -58,7 +59,48 @@ func exportSyncCommittees(rpcClient rpc.Client) error {
 	return nil
 }
 
-func exportSyncCommitteeAtPeriod(rpcClient rpc.Client, p uint64) error {
+func ExportSyncCommitteeAtPeriod(rpcClient rpc.Client, p uint64, providedTx *sqlx.Tx) error {
```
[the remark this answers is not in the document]
manuelsc: I see where you are coming from but I think the way the IDE handles this via "go to references" does a much better job at handling this case. A comment must be updated each time there's a new usage of this method with this param and gets out of date real quick. A quick usage search in the IDE reveals in an instant all usages with this param.

handlers/api.go
manuelsc: The queries have been adjusted to mimic the current status quo as to not break things. Before this PR we de-duplicated on the exporter level, which is why this de-duplication must not happen on the queries it uses. This is the first step of getting all the data in the db to allow the follow up bids to improve on the frontend

cmd/misc/main.go
```diff
@@ -359,6 +362,10 @@ func main() {
 		}
 	case "export-stats-totals":
 		exportStatsTotals(opts.Columns, opts.StartDay, opts.EndDay, opts.DataConcurrency)
+	case "export-sync-committee-periods":
+		exportSyncCommitteePeriods(rpcClient, bt, opts.StartDay, opts.EndDay, opts.DryRun)
+	case "export-sync-committee-validator-stats":
+		exportSyncCommitteeValidatorStats(rpcClient, bt, opts.StartDay, opts.EndDay, opts.DryRun, true)
```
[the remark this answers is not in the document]
manuelsc: I've talked with StefanF but he knows the misc tool more from a usage perspective rather than a technical one. Could you point me in the direction where this routine is you are talking about? I see that there is a `sync_duties_exported` but I found no usage for it outside of writing it after the export is done.

cmd/misc/main.go
```diff
@@ -1233,3 +1240,291 @@ OUTER:
 
 	logrus.Infof("finished all exporting stats totals for columns '%v' for days %v - %v, took %v", columns, dayStart, dayEnd, time.Since(start))
 }
+
+func exportSyncCommitteePeriods(rpcClient rpc.Client, bt *db.Bigtable, startDay, endDay uint64, dryRun bool) {
+	var currEpoch = uint64(0)
```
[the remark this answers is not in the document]
LuccaBitfly: You gave a thumbs up but didn't change it, forgot or won't do?

cmd/misc/main.go
```diff
@@ -1233,3 +1240,294 @@ OUTER:
...
+	}
+
+	logrus.Infof("getting exported state for day %v", day)
+
+	var err error
+	maxValidatorIndex := uint64(1999999) //bt.GetMaxValidatorindexForEpoch(lastEpoch)
+	// if err != nil {
+	// 	logrus.Errorf("error getting max validator index for epoch %v: %v", lastEpoch, err)
+	// 	return err
+	// }
```
LuccaBitfly, line 1357: Is this intentional or leftover debug code? IIRC `bt.GetMaxValidatorindexForEpoch` doesn't work for all networks, something like this would be safer:
https://github.com/manuelsc/eth2-beaconchain-explorer/blob/bf2fd38536d3615b5cac044512325cb361434b25/cmd/misc/main.go#L1195-L1196

cmd/misc/main.go
```diff
@@ -1233,3 +1240,294 @@ OUTER:
...
+		if err := db.GatherValidatorSyncDutiesForDay(validators, day, validatorData, validatorDataMux); err != nil {
+			return fmt.Errorf("error in GatherValidatorSyncDutiesForDay: %w", err)
+		}
+		return nil
+	})
+
+	var statisticsData1d []*types.ValidatorStatsTableDbRow
+	g.Go(func() error {
+		var err error
+		statisticsData1d, err = db.GatherStatisticsForDay(int64(day) - 1) // convert to int64 to avoid underflows
```
LuccaBitfly, line 1383: Gathering all statistics data seems a bit overkill, maybe just fetch sync columns in a query manuelly?

cmd/misc/main.go
```diff
@@ -1233,3 +1240,294 @@ OUTER:
...
+
+	err = g.Wait()
+	if err != nil {
+		return err
+	}
+
+	onlySyncValidatorData := make([]*types.ValidatorStatsTableDbRow, 0, len(validatorData))
+	for index := range validatorData {
+
+		if validatorData[index].ParticipatedSync > 0 || validatorData[index].MissedSync > 0 || validatorData[index].OrphanedSync > 0 {
```
LuccaBitfly, line 1398: You have to check the total columns too, these validators & their respective total columns have to get updated as well.

cmd/misc/main.go
```diff
@@ -1233,3 +1240,294 @@ OUTER:
...
+		validatorData = append(validatorData, &types.ValidatorStatsTableDbRow{
+			ValidatorIndex: i,
+			Day:            int64(day),
+		})
+	}
+
+	g := &errgroup.Group{}
+
+	g.Go(func() error {
+		if err := db.GatherValidatorSyncDutiesForDay(validators, day, validatorData, validatorDataMux); err != nil {
```
LuccaBitfly, line 1374: If I look into the function I see no use for the first param, maybe remove it? In any case, creating the `validators` slice is useless, you can just pass nil.

cmd/misc/main.go
```diff
@@ -1233,3 +1240,294 @@ OUTER:
...
+	// calculate cl income data & update totals
+	for _, data := range onlySyncValidatorData {
+
+		previousDayData := &types.ValidatorStatsTableDbRow{
+			ValidatorIndex: math.MaxUint64,
+		}
+
+		if day == 0 {
+			previousDayData.ValidatorIndex = data.ValidatorIndex
+		}
```
LuccaBitfly, line 1418: I don't get what lines 1412 to 1418 do, pls explain. Why not do it like here: https://github.com/manuelsc/eth2-beaconchain-explorer/blob/bf2fd38536d3615b5cac044512325cb361434b25/db/statistics.go#L185-L192

cmd/misc/main.go
```diff
@@ -1233,3 +1240,294 @@ OUTER:
...
+	if startDay > 0 {
+		firstEpoch, _ := utils.GetFirstAndLastEpochForDay(startDay)
+		firstPeriod = utils.SyncPeriodOfEpoch(firstEpoch)
+	}
+
+	if endDay <= 0 {
+		var err error
+		currEpoch, err = db.GetLatestFinalizedEpoch()
+		if err != nil {
+			logrus.WithError(err).Errorf("error getting latest finalized epoch")
```
LuccaBitfly, line 1261: `utils.LogError` should always be used instead of `logger.Error`, forgot to mention in my first review round. You don't have to change it everywhere you see it, but every time you write or touch a `logger.Error`, it has to be become a `utils.LogError`.

cmd/misc/main.go
```diff
@@ -1233,3 +1240,294 @@ OUTER:
...
+				missed_sync = $3,
+				missed_sync_total = $4,
+				orphaned_sync = $5,
+				orphaned_sync_total = $6
+				WHERE day = $7 AND validatorindex = $8`,
+				data.ParticipatedSync, data.ParticipatedSyncTotal,
+				data.MissedSync, data.MissedSyncTotal,
+				data.OrphanedSync, data.OrphanedSyncTotal,
+				data.Day, data.ValidatorIndex)
+		}
```
LuccaBitfly, line 1472: Is doing 1 update query per validator efficient? Why not do something with `INSERT VALUES`?

LuccaBitfly (changes requested):
Sorry, wasn't aware that the sync reexport tools are gone so I didn't look at `exportSyncCommitteeValidatorStats` in my first review.

cmd/misc/main.go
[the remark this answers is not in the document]
LuccaBitfly: Nvm, reexport tools are gone and the existence of `exportSyncCommitteeValidatorStats` is valid.

cmd/misc/main.go
manuelsc: The goal is to minimize bug surfaces by using code the exporter uses too. It is overkill but this code runs once per network which is why I prefer the inefficient but battle tested one

cmd/misc/main.go
manuelsc: This is a filter to sync out all validators that had sync committees that day. Total will be updated further below, but if the validator was not in a sync committee that day total will stay the same anyway.

cmd/misc/main.go
manuelsc: It only updates validators with sync committee per day, so that's 512 updates per day. Which is fine performance wise and the code runs once so I did not see the need to optimize it any further. We can keep it simple and more readable therefore.

cmd/misc/main.go
LuccaBitfly: Let's assume you use this tool over a day range that includes 2 distinct sync committee periods, with no validators being in both periods:
Day `x`: validators in sync period `y` have sync duties => their sync stats are updated; sync period `y+1` starts
Day `x+1`: validators in sync period `y` have **no** sync duties => their sync stats are not updated, even though their total changed the previous day.
Pls correct me if I'm wrong.

cmd/misc/main.go
LuccaBitfly: Was confused about the if day == 0 part, new version is understandable again :)

LuccaBitfly:
Build fails lol

cmd/misc/main.go
manuelsc: Am not sure what you are getting at. On Day `x` sync stats are updated including the total. So the total stats on `x+1` for the previous day are already correct?

Issue #2681, opened 2023-11-08
Issue with Missing Validators in Paginated API Endpoint /api/v1/validator/withdrawalCredentials/{withdrawalCredentialsOrEth1address}


nicosampler:
**Description:**
I have encountered an issue with the paginated API endpoint at https://gnosischa.in/api/v1/validator/withdrawalCredentials/{withdrawalCredentialsOrEth1address. The issue revolves around missing validators when using pagination.

**Problem Description:**
I have been testing this endpoint by setting the limit to 200 and increasing the offset in increments of 200 (e.g., 0, 200, 400, etc.) with the goal of reaching a response with fewer than 200 results. The problem I've found is that when the pagination reaches a request with less than 200 validators, the sum of all validators returned is not coincident with the number of validators the address controls. 

**Possible cause**
I suspect the issue may be related to the sorting of the query in this [function](https://github.com/gobitfly/eth2-beaconchain-explorer/blob/6ff245c18e407dfc5c1f4827beb5ed719bf4918a/handlers/api.go#L3768). It appears that the query is not sorted by any specific field, which could potentially lead to inconsistencies in the results.

I kindly request a review and investigation into this issue. The objective is to ensure that the pagination strategy works consistently, returning the correct number of validators.

D13ce:
@nicosampler , many thanks for your report! We have created an internal ticket for your request. :bow:

D13ce:
@nicosampler , just so you know, the PR tackling your issue has just been merged and will be part of our next release.

PR #2685, opened 2023-11-10
(BIDS-2663) Add sorting to postgres query using pagination

D13ce:
This PR adds sorting to the `api1/validator/withdrawalCredentials/{withdrawalCredentialsOrEth1address}` API endpoint to ensure correct pagination.

This should fix https://github.com/gobitfly/eth2-beaconchain-explorer/issues/2681.

### DIFF
--- a/handlers/api.go
+++ b/handlers/api.go
@@ -3784,6 +3784,7 @@ func ApiWithdrawalCredentialsValidators(w http.ResponseWriter, r *http.Request) 
 		pubkey
 	FROM validators
 	WHERE withdrawalcredentials = $1
+	ORDER BY validatorindex ASC
 	LIMIT $2
 	OFFSET $3
 	`, credentials, limit, offset)


Eisei24 (approved):
utACK
I could not find withdrawal credentials where the API returned the wrong amount of validators on any network. Adding an `ORDER BY` would also not make a difference.

But if I interpret the bug as
- Make an API call with `?limit=200&offset=0` => Get 200 validator results
- Make an API call with `?limit=200&offset=200` => Get 200 validator results
- There are only ~390 different validators because they appear in both API calls

then the changes makes sense and should fix the issue.

D13ce:
> utACK I could not find withdrawal credentials where the API returned the wrong amount of validators on any network. Adding an `ORDER BY` would also not make a difference.
> 
> But if I interpret the bug as
> 
>     * Make an API call with `?limit=200&offset=0` => Get 200 validator results
> 
>     * Make an API call with `?limit=200&offset=200` => Get 200 validator results
> 
>     * There are only ~390 different validators because they appear in both API calls
> 
> 
> then the changes makes sense and should fix the issue.

I have written a small test program that iterates over 50k validators based on a single withdrawal credential. Without ordering, I have missed some every run. With ordering I received all 50k validators every run.

PR #2686, opened 2023-11-12
(BIDS-2670) batch machine metric inserts

peterbitfly:
### <samp>🤖 Generated by Copilot at 3b8030f</samp>

This pull request improves the Bigtable integration for storing machine metrics. It introduces a new type `BulkMutation` to represent batched mutations, and implements a buffering and batching mechanism in `db/bigtable.go` using a channel and a goroutine.


db/bigtable.go
```diff
@@ -110,26 +112,82 @@ func InitBigtable(project, instance, chainId, redisAddress string) (*Bigtable, e
...
+		tableValidators:                btClient.Open("beaconchain_validators"),
+		tableValidatorsHistory:         btClient.Open("beaconchain_validators_history"),
+		chainId:                        chainId,
+		redisCache:                     rdc,
+		lastAttestationCacheMux:        &sync.Mutex{},
+		v2SchemaCutOffEpoch:            utils.Config.Bigtable.V2SchemaCutOffEpoch,
+		machineMetricsQueuedWritesChan: make(chan types.BulkMutation, MAX_BATCH_MUTATIONS),
+	}
+
+	go bt.commitQueuedMachineMetricWrites()
```
manuelsc, line 131: Minor nitpick, feel free to ignore. Only really needed for frontend and api instances, makes no sense to run this in frontend data updater, misc tool notification collector etc.

db/bigtable.go
```diff
@@ -110,26 +112,82 @@ func InitBigtable(project, instance, chainId, redisAddress string) (*Bigtable, e
...
+
+	batchSize := 10000
+
+	muts := &types.BulkMutations{
+		Keys: make([]string, 0, batchSize),
+		Muts: make([]*gcp_bigtable.Mutation, 0, batchSize),
+	}
+
+	tmr := time.NewTicker(time.Second * 10)
+	for {
```
manuelsc, line 149: Minor issue as this is not really a problem with the current use case, but it might be in future if that changes: tmr chan never closes and loop runs infinite, could consider closing the channel and stoping the loop in case the machineMetricsQueuedWritesChan channel is closed.

db/bigtable.go
```diff
@@ -110,26 +112,82 @@ func InitBigtable(project, instance, chainId, redisAddress string) (*Bigtable, e
...
+				} else {
+					logger.Errorf("error writing queued machine metrics to bigtable: %v", err)
+				}
+			}
+
+		case <-tmr.C:
+			if len(muts.Keys) > 0 {
+				err := bigtable.WriteBulk(muts, bigtable.tableMachineMetrics)
+
+				if err == nil {
```
manuelsc, line 175: Super minor, feel free to ignore: duplicate code block, this happens in both select cases

manuelsc (approved):
Found no issues with the core logic. Only minor nitpicks but nothing that would be a show stopper in case we need this urgently.

PR #2690, opened 2023-11-13
(BIDS-2669) api: fix ordering of result of /api/v1/validator/leaderboard

guybrush:
### <samp>🤖 Generated by Copilot at ed84339</samp>

Fixed a bug in the `ApiValidatorLeaderboard` function that returned the wrong order of validators. Updated the SQL query in `handlers/api.go` to sort by rank7d in ascending order.

### DIFF
--- a/handlers/api.go
+++ b/handlers/api.go
@@ -2487,7 +2487,7 @@ func ApiValidatorLeaderboard(w http.ResponseWriter, r *http.Request) {
 				rank7d, 
 				validatorindex
 			FROM validator_performance 
-			ORDER BY rank7d DESC LIMIT 100`)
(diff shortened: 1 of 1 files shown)


Eisei24 (commented):
This fixes the bug with the reverse order but I noticed that the rank is not consistent.
Here is an example from Prater:

- On `/validators/leaderboard` the last rank is 606106 which is for validatorindex 318888
- When opening the validator `/validator/318888` it has rank 606091
- The API also gives rank 606091 for that validator but also has other validators for the lower ranks up to 606106 (which would be validatorindex 602078 right now)

Is this a known issue?

Eisei24 (approved):
lgtm

PR #2697, opened 2023-11-14
(NOBIDS) rename api error func

LuccaBitfly:
Renames two functions, cmon just approve, trust me

### <samp>🤖 Generated by Copilot at 77cf0be</samp>

No summary available (Limit exceeded: required to process 53072 tokens, but only 50000 are allowed per call)

### DIFF
--- a/handlers/api.go
+++ b/handlers/api.go
@@ -230,7 +230,7 @@ func ApiEthStoreDay(w http.ResponseWriter, r *http.Request) {
 	} else {
 		day, e := strconv.ParseInt(vars["day"], 10, 64)
 		if e != nil {
-			sendErrorResponse(w, r.URL.String(), "invalid day provided")
+			sendBadRequestResponse(w, r.URL.String(), "invalid day provided")
 			return
 		}
 		rows, err = db.ReaderDb.Query(query+` AND day = $1;`, day)
@@ -238,7 +238,7 @@ func ApiEthStoreDay(w http.ResponseWriter, r *http.Request) {
 
 	if err != nil {
 		logger.Errorf("error retrieving eth.store data: %v", err)
-		sendErrorResponse(w, r.URL.String(), "could not retrieve db results")
+		sendBadRequestResponse(w, r.URL.String(), "could not retrieve db results")
 		return
 	}
 	defer rows.Close()
@@ -274,7 +274,7 @@ func ApiEpoch(w http.ResponseWriter, r *http.Request) {
 
 	epoch, err := strconv.ParseInt(vars["epoch"], 10, 64)
 	if err != nil && vars["epoch"] != "latest" && vars["epoch"] != "finalized" {
-		sendErrorResponse(w, r.URL.String(), "invalid epoch provided")
+		sendBadRequestResponse(w, r.URL.String(), "invalid epoch provided")
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


handlers/api.go
```diff
@@ -4095,12 +4095,12 @@ func adjustQueryResults(data []interface{}, adjustQueryEntriesFuncs ...func(map[
 	return nil
 }
 
-// SendErrorResponse exposes sendErrorResponse
```
D13ce, line 4098: **Minor:** This comment now is outdated. Also, search for `sendErrorResponse` within the sources and you will find two more locations. Those are both code that is commented out but I suggest to either remove it completely or to fix it.

handlers/api.go
```diff
@@ -4095,12 +4095,12 @@ func adjustQueryResults(data []interface{}, adjustQueryEntriesFuncs ...func(map[
...
-func SendErrorResponse(w http.ResponseWriter, route, message string) {
-	sendErrorResponse(w, route, message)
+// SendBadRequestResponse exposes sendErrorResponse
+func SendBadRequestResponse(w http.ResponseWriter, route, message string) {
+	sendBadRequestResponse(w, route, message)
 }
 
-func sendErrorResponse(w http.ResponseWriter, route, message string) {
+func sendBadRequestResponse(w http.ResponseWriter, route, message string) {
 	sendErrorWithCodeResponse(w, route, message, 400)
```
D13ce, line 4104: **Nitpick:** I know this does not come from your PR but as you have touched this function I suggest to replace the hardcoded 400 with `http.StatusBadRequest`.

Related: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/77cf0bef9cdc9d872321f989479eca66e4ce334c/handlers/api.go#L4107-L4109
and `http.StatusInternalServerError`.

handlers/api.go
```diff
@@ -4095,12 +4095,12 @@ func adjustQueryResults(data []interface{}, adjustQueryEntriesFuncs ...func(map[
 	return nil
 }
 
-// SendErrorResponse exposes sendErrorResponse
-func SendErrorResponse(w http.ResponseWriter, route, message string) {
-	sendErrorResponse(w, route, message)
+// SendBadRequestResponse exposes sendErrorResponse
+func SendBadRequestResponse(w http.ResponseWriter, route, message string) {
+	sendBadRequestResponse(w, route, message)
 }
```
D13ce, line 4101: **Question:** Why not remove `sendBadRequestResponse` and only have `SendBadRequestResponse`? I know, it was like this before, but I am asking as you have touched it.
IF you want to change it, you should also take a look at https://github.com/gobitfly/eth2-beaconchain-explorer/blob/77cf0bef9cdc9d872321f989479eca66e4ce334c/handlers/api.go#L4123-L4126 for consistency.

D13ce (commented):
ptal

handlers/api.go
LuccaBitfly: remove

handlers/api.go
LuccaBitfly: fix

handlers/api.go
LuccaBitfly: fix

D13ce (approved):
looks lgtm to me

PR #2698, opened 2023-11-14
Nobids/sort batch bigtable inserts

peterbitfly:
Sort keys prior to batched bigtable writes

### <samp>🤖 Generated by Copilot at afd5783</samp>

This pull request refactors the `db` package to use a common `WriteBulk` function and a `BulkMutations` struct for writing bulk mutations to Bigtable tables. It also improves code readability and consistency by moving common Bigtable functions to a separate file, using a constant parameter for batch size, and removing unused code. Additionally, it adds a utility function to get the parent function name for logging purposes.


db/bigtable_common.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+		for _, e := range errs {
+			if e != nil {
+				return e
+			}
+		}
+		logger.Infof("%s: wrote from %v to %v rows to bigtable in %.1fs", callingFunctionName, start, numKeys, time.Since(startTime).Seconds())
+		if err != nil {
+			return err
+		}
```
D13ce, line 65: **Nitpick:** This has already been checked a few lines above.
```suggestion
```

db/bigtable_common.go
```diff
@@ -0,0 +1,133 @@
...
+	"eth2-exporter/utils"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	gcp_bigtable "cloud.google.com/go/bigtable"
+)
+
+func (bigtable *Bigtable) WriteBulk(mutations *types.BulkMutations, table *gcp_bigtable.Table, batchSize int) error {
```
D13ce, line 15: **Question:** Having two `for` loops here is odd and feels error prone as you would have to make changes in two places. Why not have a single `for` loop?

db/bigtable_common.go
```diff
@@ -0,0 +1,133 @@
...
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	return nil
+}
+
+func (bigtable *Bigtable) DeleteRowsWithPrefix(prefix string) {
```
D13ce, line 72: This is currently not used and does require some refactoring. I suggest to delete it. Once it is used, it can be implemented properly with your new `WriteBulk` functionality.

D13ce (changes requested):

types/bigtable.go
```diff
@@ -54,3 +56,37 @@ type GasNowHistory struct {
 	Fast     *big.Int
 	Rapid    *big.Int
 }
+
+type GetBlockTimings struct {
+	Headers  time.Duration
+	Receipts time.Duration
+	Traces   time.Duration
+}
```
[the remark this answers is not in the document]
peterbitfly: moved back to the eth1.go file

db/bigtable.go
```diff
@@ -545,8 +546,10 @@ func (bigtable *Bigtable) SaveValidatorBalances(epoch uint64, validators []*type
 	// start := time.Now()
 	ts := gcp_bigtable.Timestamp(0)
 
-	muts := make([]*gcp_bigtable.Mutation, 0, MAX_BATCH_MUTATIONS)
-	keys := make([]string, 0, MAX_BATCH_MUTATIONS)
+	muts := &types.BulkMutations{
+		Keys: make([]string, 0, len(validators)),
+		Muts: make([]*gcp_bigtable.Mutation, 0, len(validators)),
+	}
```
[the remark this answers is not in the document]
peterbitfly: done

db/bigtable_common.go
```diff
@@ -0,0 +1,133 @@
...
+func (bigtable *Bigtable) WriteBulk(mutations *types.BulkMutations, table *gcp_bigtable.Table, batchSize int) error {
+
+	callingFunctionName := utils.GetParentFuncName()
+
+	ctx, done := context.WithTimeout(context.Background(), time.Minute*5)
+	defer done()
+
+	// pre-sort mutations for efficient bulk inserts
+	sort.Sort(mutations)
+	length := batchSize
```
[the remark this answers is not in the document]
peterbitfly: done

db/bigtable.go
```diff
@@ -2852,7 +2757,7 @@ func (bigtable *Bigtable) ClearByPrefix(table string, family, prefix string, dry
 		if len(mutsDelete.Keys) == 1000000 {
```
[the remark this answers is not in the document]
peterbitfly: we still need to commit in batches here as loading all keys to be deleted in memory first is not feasible as the delete function could be used to delete millions of rows

db/bigtable_common.go
peterbitfly: deleted in favor of the ClearByPrefix functions which does the same

db/bigtable_common.go
peterbitfly: I don't understand the question

D13ce: This is because I wrote it wrong, sorry. Here is the fixed version:

---

**Question:** Having two parts calling `ApplyBulk` instead of doing everything within the `for` loop feels error prone as you would have to make changes in two places whenever changes are needed. Why not have everything related to `ApplyBulk` within the `for` loop and get rid of this part:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/507bb0a4ac1516f0eb939147e4a349bae1f10bf4/db/bigtable_common.go#L50-L67

D13ce: As discussed in DMs: Won't be changed.

D13ce (approved):
lgtm, really nice PR :tada:

PR #2700, opened 2023-11-14
Batch remove from watchlist for App

manuelsc:
copilot:summary

### DIFF
--- a/cmd/explorer/main.go
+++ b/cmd/explorer/main.go
@@ -328,6 +328,7 @@ func main() {
 		apiV1AuthRouter.HandleFunc("/validator/{pubkey}/add", handlers.UserValidatorWatchlistAdd).Methods("POST", "OPTIONS")
 		apiV1AuthRouter.HandleFunc("/validator/{pubkey}/remove", handlers.UserValidatorWatchlistRemove).Methods("POST", "OPTIONS")
 		apiV1AuthRouter.HandleFunc("/dashboard/save", handlers.UserDashboardWatchlistAdd).Methods("POST", "OPTIONS")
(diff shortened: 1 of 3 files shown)


cmd/explorer/main.go
```diff
@@ -328,6 +328,7 @@ func main() {
 		apiV1AuthRouter.HandleFunc("/validator/{pubkey}/add", handlers.UserValidatorWatchlistAdd).Methods("POST", "OPTIONS")
 		apiV1AuthRouter.HandleFunc("/validator/{pubkey}/remove", handlers.UserValidatorWatchlistRemove).Methods("POST", "OPTIONS")
 		apiV1AuthRouter.HandleFunc("/dashboard/save", handlers.UserDashboardWatchlistAdd).Methods("POST", "OPTIONS")
+		apiV1AuthRouter.HandleFunc("/dashboard/remove", handlers.UserDashboardWatchlistRemove).Methods("POST", "OPTIONS")
```
LuccaBitfly, line 331: Question: Why do we not use the `DELETE` method? E.g.
```suggestion
		apiV1AuthRouter.HandleFunc("/dashboard/watchlist", handlers.UserDashboardWatchlistRemove).Methods("DELETE", "OPTIONS")
```
Would be more in line with REST guidelines.

manuelsc: I don't think we use other HTTP methods outside of GET and POST anywhere in the api. But you are right, this sounds like a case for v2 if we want to be more conform to the standard. Right now I'd argue for consistency since we don't use other methods right now.

LuccaBitfly (approved):
lgtm

PR #2707, opened 2023-11-15
(BIDS-2548) add total deposits columns

Eisei24:
### <samp>🤖 Generated by Copilot at ba02a37</samp>

This pull request adds two new metrics to the validator statistics: the total number and amount of deposits made by each validator. These metrics are stored in the `validator_stats` table, exported in the CSV file, and represented in the `ExporterValidatorStats` struct.


Eisei24:
This PR prepares everything for the reexport but BIDS-2769 has to be done first to get rid of day -1 which contains deposits.
The reexport of total stats only handles days >= 0.

db/migrations/tbd_add_total_deposits.sql
```diff
@@ -0,0 +1,13 @@
...
+ALTER TABLE validator_stats ADD COLUMN IF NOT EXISTS deposits_total INT;
+ALTER TABLE validator_stats ADD COLUMN IF NOT EXISTS deposits_amount_total BIGINT;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+SELECT 'down SQL query - remove total deposits count and amount columns from stats';
+ALTER TABLE validator_stats DROP COLUMN IF EXISTS deposits_total;
+ALTER TABLE validator_stats DROP COLUMN IF EXISTS deposits_amount_total;
+-- +goose StatementEnd
```
LuccaBitfly, line 13: image
I think this means that file should end with newline?

LuccaBitfly (approved):
utAck

db/migrations/tbd_add_total_deposits.sql
Eisei24: We have some migration files that don't end on a newline but since most of them do I added it.

PR #2713, opened 2023-11-16
(BIDS-2644) Add underflow protection for missed attestations

D13ce:
This PR adds an underflow protection for the validator page (missed attestation count). It is required for wrong data on Prater currently. Data will be fixed later.


Eisei24 (approved):
lgtm, quickfix works as expected.

PR #2714, opened 2023-11-17
Bids 2616/validators data filter error

D13ce:
This PR fixes errors that are thrown when the validator overview page is shown with a filter that results in 0 validators.
It furthermore improves and unifies the error handling within `validators.go`.


Eisei24 (approved):
lgtm

PR #2717, opened 2023-11-20
(BIDS-2594) Add wrapper function and remove unused parameter

D13ce:
This PR adds a new wrapper function called `FormatAddressWithLimitsInAddressPageTable`.
It furthermore removes an unused parameter from `formatAddress`.

### DIFF
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -38,8 +38,11 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-// when changing this, you will have to update the swagger docu for func ApiEth1Address too
-const ECR20TokensPerAddressLimit = uint64(200)
+const (
+	ECR20TokensPerAddressLimit    = uint64(200) // when changing this, you will have to update the swagger docu for func ApiEth1Address too
+	digitLimitInAddressPagesTable = 18
+	nameLimitInAddressPagesTable  = 18
+)
 
 var ErrBlockNotFound = errors.New("block not found")
 
@@ -2010,9 +2013,6 @@ func (bigtable *Bigtable) GetAddressTransactionsTableData(address []byte, pageTo
 		fromName := names[string(t.From)]
 		toName := names[string(t.To)]
 
-		from := utils.FormatAddress(t.From, nil, fromName, false, false, !bytes.Equal(t.From, address))
-		to := utils.FormatAddress(t.To, nil, toName, false, false, !bytes.Equal(t.To, address))
-
 		method := bigtable.GetMethodLabel(t.MethodId, t.InvokesContract)
 
 		tableData[i] = []interface{}{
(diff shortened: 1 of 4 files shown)


utils/eth1.go
```diff
@@ -196,15 +210,16 @@ func formatAddress(address []byte, token []byte, name string, isContract bool, l
...
 	} else {
-		// link & token
 		if token != nil {
+			// link & token
 			ret += fmt.Sprintf(`<a href="/`+link+`/0x%x#erc20Txns" target="_parent" data-html="true" data-toggle="tooltip" data-placement="bottom" title="" data-original-title="%s">%s</a>`, address, tooltip, name)
-		} else { // just link
-			ret += fmt.Sprintf(`<a href="/`+link+`/0x%x`+urlFragment+`" target="_parent" data-html="true" data-toggle="tooltip" data-placement="bottom" title="" data-original-title="%s">%s</a>`, address, tooltip, name)
+		} else {
+			// just link
+			ret += fmt.Sprintf(`<a href="/`+link+`/0x%x" target="_parent" data-html="true" data-toggle="tooltip" data-placement="bottom" title="" data-original-title="%s">%s</a>`, address, tooltip, name)
```
Eisei24, line 222: **suggestion:** It's odd to add `link` to the text like this when we already have an `Sprintf`.
How about an `%s` instead?

db/bigtable_eth1.go
```diff
@@ -2849,7 +2841,7 @@ func (bigtable *Bigtable) GetEth1ERC721ForAddress(prefix string, limit int64) ([
 	return data, indexes[len(indexes)-1], nil
 }
 
-func (bigtable *Bigtable) GetAddressErc721TableData(address string, pageToken string) (*types.DataTableResponse, error) {
+func (bigtable *Bigtable) GetAddressErc721TableData(address []byte, pageToken string) (*types.DataTableResponse, error) {
```
Eisei24, line 2844: **issue:** This `string` => `[]byte` breaks the already present use of `address` namely for `pageToken`. For that you need to change `%s` => `%x`.
Currently we can't find data because of wrong keys and so the Erc721 tab never shows up.

The issue also for `GetAddressErc1155TableData`.

You can use **Prater** address **0xb64a30399f7f6b0c154c2e7af0a3ec7b0a5b131a** to test this.

Eisei24 (changes requested):
Please update the ERC721 and ERC1155 handling.

utils/eth1.go
D13ce: You are right, fixed.

db/bigtable_eth1.go
D13ce: VERY good catch, thank you. Also :+1: for including a test address.
=> Fixed.

Eisei24 (approved):
lgtm, as discussed I adjusted the cutoff values so the behaviour stays the same.

PR #2725, opened 2023-11-23
Bids 2673/fix sync table on validator page

D13ce:
This PR splits up the single slot range query for the sync duty history and the sync participation into sub ranges that contain consecutive slots only. This enables the sync table on the `/validator` page to show pages containing more than one sync period again.
It furthermore adds to comments explaining both `GetValidatorSyncDutiesHistory` and `GetSyncParticipationBySlotRange`.

### DIFF
--- a/handlers/validator.go
+++ b/handlers/validator.go
@@ -2053,8 +2053,7 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
 		endIndex := start
 
 		// retrieve sync duties and sync participations
-		syncDuties := make(map[uint64]map[uint64]*types.ValidatorSyncParticipation)
-		syncDuties[validatorIndex] = make(map[uint64]*types.ValidatorSyncParticipation, length)
+		syncDuties := make(map[uint64]*types.ValidatorSyncParticipation, length)
 		participations := make(map[uint64]uint64, length)
 		{
 			// the slot range for the given table page might contain multiplie sync periods and therefore we may need to split the queries to avoid fetching potentially thousands of duties at once
@@ -2086,7 +2085,7 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
 					return
 				}
 				for slot, duty := range sdh[validatorIndex] {
-					syncDuties[validatorIndex][slot] = duty
+					syncDuties[slot] = duty
 				}
 
 				par, err := db.GetSyncParticipationBySlotRange(slotRange.StartSlot, slotRange.EndSlot)
@@ -2126,13 +2125,11 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
 			participation := participations[slot]
 
 			status := uint64(0)
-			if syncDuties[validatorIndex] != nil {
-				if syncDuties[validatorIndex][slot] != nil {
-					status = syncDuties[validatorIndex][slot].Status
-				}
-				if _, ok := missedSlotsMap[slot]; ok {
-					status = 3
-				}
+			if syncDuties[slot] != nil {
+				status = syncDuties[slot].Status
+			}
+			if _, ok := missedSlotsMap[slot]; ok {
+				status = 3
 			}
 
 			tableData = append(tableData, []interface{}{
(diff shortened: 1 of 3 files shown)


handlers/validator.go
```diff
@@ -2053,30 +2052,59 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
 		}
 		endIndex := start
 
-		endSlot := slots[endIndex]
-		startSlot := slots[startIndex]
+		// retrieve sync duties and sync participations
+		syncDuties := make(map[uint64]map[uint64]*types.ValidatorSyncParticipation)
```
Eisei24, line 2056: **suggestion:** Can this be made into a normal map? The validatorIndex "dimension" is never relevant for anything.

Eisei24 (commented):
lgtm but please have a look at the possible improvement.

handlers/validator.go
D13ce: Good call, fixed.

handlers/validator.go
LuccaBitfly, line 1: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/e2d8db10c57f3fc9f427c9a18374ab5299bc361d/handlers/validator.go#L2026
could change this to early return.

handlers/validator.go
```diff
@@ -2053,30 +2052,58 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
 		}
 		endIndex := start
 
-		endSlot := slots[endIndex]
-		startSlot := slots[startIndex]
+		// retrieve sync duties and sync participations
+		syncDuties := make(map[uint64]*types.ValidatorSyncParticipation, length)
+		participations := make(map[uint64]uint64, length)
+		{
+			// the slot range for the given table page might contain multiplie sync periods and therefore we may need to split the queries to avoid fetching potentially thousands of duties at once
```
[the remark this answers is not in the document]
D13ce: Fixed

handlers/validator.go
D13ce: Okay, done. I have used this opportunity to refactor the code a bit in the context of readability (no behavior change, just formatting).

handlers/validator.go
```diff
@@ -2006,152 +2006,157 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
...
+	data := &types.DataTableResponse{
+		Draw:            draw,
+		RecordsTotal:    0,
+		RecordsFiltered: 0,
+		Data:            tableData,
+	}
 
-		latestProposedSlot := services.LatestProposedSlot()
+	if len(syncPeriods) == 0 {
+		// no sync periods for this validator => early exit with empty tableData
```
Eisei24, line 2031: **issue:** There is no early return here in case of empty `syncPeriods` as was inteded.
You just return in case there is an encode error.

If you navigate to e.g. `/validator/428593/sync?draw=1&start=0&length=10` for a validator without syncs on Prater you get past this point and trigger a panic at https://github.com/gobitfly/eth2-beaconchain-explorer/blob/dd8e64b12f9902c7fc750cfc99150c660b8002f2/handlers/validator.go#L2079 because `slots` is empty.

The same line also triggers a panic if the user searches for entries that lie beyond the end.
E.g. `/validator/31484/sync?draw=10001&start=100000&length=10` (on Prater) has 2 sync and therefore `len(slots) = 16384`. As a result `slots[100000]` crashes.

Eisei24 (changes requested):
Please fix the mentioned issues.

handlers/validator.go
D13ce: Jeah, that's because I did not implement an "Early Return" but only an "Early" i.e. an "Early Return" without the proper "Return".
I am sorry, I will fix that quite soon. Shame on me.

D13ce: Fixed.

handlers/validator.go
```diff
@@ -2057,6 +2058,9 @@ func ValidatorSync(w http.ResponseWriter, r *http.Request) {
 	}
 
 	totalCount = uint64(len(slots))
+	if start >= totalCount {
+		start = totalCount - 1
+	}
 
 	startIndex := start + length - 1
 	if startIndex > uint64(len(slots)-1) {
```
Eisei24, line 2066: **suggestion:** Could be
```suggestion
	if startIndex >= totalCount{
```
and the line below it `startIndex = totalCount - 1`

Eisei24 (approved):
lgtm, if you want you can change the suggestion.

handlers/validator.go
D13ce: Done.

PR #2730, opened 2023-11-27
(BIDS-2330) misc: add command fix-blocks

guybrush:
copilot:summary


cmd/misc/main.go
```diff
@@ -1536,3 +1549,354 @@ func reExportSyncCommittee(rpcClient rpc.Client, p uint64, dryRun bool) error {
...
+
+		finalizedSlot, err := strconv.ParseUint(res.Data.Message.Slot, 10, 64)
+		if err != nil {
+			return fmt.Errorf("error parsing finalized slot: %w", err)
+		}
+		iStart := i
+		iEnd := i + 1000
+		if iEnd > finalizedSlot+1 {
+			iEnd = finalizedSlot + 1
+		}
```
D13ce, line 1750: **Minor:** This should also include something like
```Go
if iEnd > end {
	iEnd = end
}
```

Because right now it could happen that `start` is something like `end - 3` and in that case end would still be `iEnd` would still result in `start + 1000`.

D13ce (commented):
utACK

Just found a minor issue, feel free to ignore.

PR #2731, opened 2023-11-27
(BIDS 1823) Adjusting time duration formatting

thib-wien:
Adjusts the format of numerous time durations (for example, time.Hour*24 -> utils.Day)

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 1d9e813</samp>

This pull request simplifies and improves the readability of time durations in various files of the eth2-beaconchain-explorer project. It replaces the use of `time` package expressions with predefined constants from the `utils` package or shorter forms, such as `time.Minute` and `time.Hour`. This change affects database, exporter, handler, mail, RPC, and service functions.


Eisei24 (changes requested):
Occurences I found that can still be changed:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/f5b2ace3a16b8188e4ea7adfa481f714f90092cc/cmd/explorer/main.go#L623
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/bc91f8c46890ba3e6fbe7950e218ec38300b3118/db/db.go#L101
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/bc91f8c46890ba3e6fbe7950e218ec38300b3118/db/db.go#L117
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/af5a829657c9d69dc5579365778d9df552c9cd7d/services/monitoring.go#L263
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/5c75bec9775489074041a7ad1079aff7a5c239d6/utils/utils.go#L368 (there is also a comment with the same code in the method above it which can be deleted)

I also noticed that we set `utils.Year` as
```
	Month    = 30 * Day
	Year     = 12 * Month
```
which is just 360 days.
`utils.Year` is used once in the code. Check that it makes sense that it should be 365 days there instead and if so update it to `Year = 365 * Day`.
This change is too small to create its own BIDS for it so lets just do it here.

thib-wien:
@Eisei24 Thanks for your remarks :)

* I worked on the occurrences 1, 2, 3, 5 that you found.
I was not sure that the spirit of the code was based on the second or on the minute. So I follow your preference.

* Regarding your 4th finding, I did not change this occurrence because the unit of time that this part of the code uses to think is the minute.
I talked about it on Friday morning during our meeting and I got a reply from my mentor who thinks that leaving them as they are makes sense.
If after discussing with him you find an agreement, I will follow it.

* In your point 5, you suggest me to delete a commented code. I did it. In the function with this comment, I also replaced "24" with a constant calculated during compilation, as you prefer constants to hard coded numbers.
Not sure whether my change is relevant.

thib-wien:
@Eisei24 Regarding your request to study `utils.Year` (which appears in your email but not here), I also noticed its inaccuracy and [asked about it this morning on our Discord](https://discord.com/channels/808317185873018881/808634305978302475/1178634483612790815).

Peter doesn't suggest to leave it or to modify it, he is maybe unsure because it might introduce unexpected behaviors and display different statistics. I do not know the code well enough yet to be sure it would be safe.

A concrete example:

Changing `utils.Year` to 365 days will modify statistics in various part of the site (it is used only to define `handlers.year` in _common.go_).
The reason is that year is assigned to `proposalTimeFrame` in function `getProposalLuck` (in _common.go_) and this function will call `calcExpectedSlotProposals` (in the same same source file) in order to calculate something with it. Then, `getProposalLuck` will use the returned result (see its last line) to calculate the luck of the validator.

(the duration of `utils.Month` could be made more accurate (on average) too)

Eisei24 (approved):
lgtm

It is totally fine to leave the 4th occurrence as it is, I have no strong feelings one way or the other.
I wrote the 360 day issue at first but then edited it away because I noticed it was already mentioned in discord and seemed accepted.

PR #2732, opened 2023-11-27
(BIDS-2641) hide "ETH" confusable tokens

LuccaBitfly:
Imports library to hide token symbols that look like "ETH", even if they do some unicode magic to not be the exact string "ETH".


templates/validator/tables.html
```diff
@@ -732,7 +732,7 @@ <h6 class="">This table displays the deposits received and processed by the beac
               data-html="true"
               title="
                 CL: {{ formatPercentageWithPrecision .Income.Apr7d.Cl 2 }}% | {{ formatPercentageWithPrecision .Income.Apr31d.Cl 2 }}% | {{ formatPercentageWithPrecision .Income.Apr365d.Cl 2 }}% <br> 
-                El: {{ formatPercentageWithPrecision .Income.Apr7d.El 2 }}% | {{ formatPercentageWithPrecision .Income.Apr31d.El 2 }}% | {{ formatPercentageWithPrecision .Income.Apr365d.El 2 }}%">
+                EL: {{ formatPercentageWithPrecision .Income.Apr7d.El 2 }}% | {{ formatPercentageWithPrecision .Income.Apr31d.El 2 }}% | {{ formatPercentageWithPrecision .Income.Apr365d.El 2 }}%">
```
Eisei24, line 735: Good find.

utils/utils.go
```diff
@@ -1374,34 +1375,29 @@ func FormatPoolPerformance(val float64) template.HTML {
...
+func isMaliciousToken(symbol string) bool {
+	urls := xurls.Relaxed.FindAllString(symbol, -1)
+	isConfusable := confusables.IsDangerous(symbol, []string{"latin"})
+	return len(urls) > 0 || isConfusable || symbol == "ETH"
+}
+
 func FormatTokenSymbolHTML(tmpl template.HTML) template.HTML {
-	tmplString := (string(tmpl))
+	tmplString := FormatTokenSymbol(string(tmpl))
 	symbolTitle := FormatTokenSymbolTitle(tmplString)
```
Eisei24, line 1399: **issue:** This doesn't work because the first line already changes the symbol from invalid "ETH" to "[hidden-symbol] ⚠️" and you input that into `FormatTokenSymbolTitle` which it doesn't have a problem with so it returns "".

utils/utils.go
```diff
@@ -1374,34 +1375,29 @@ func FormatPoolPerformance(val float64) template.HTML {
...
+	if isMaliciousToken(symbol) {
 		return "[hidden-symbol] ⚠️"
 	}
 	return symbol
 }
 
+func isMaliciousToken(symbol string) bool {
+	urls := xurls.Relaxed.FindAllString(symbol, -1)
+	isConfusable := confusables.IsDangerous(symbol, []string{"latin"})
+	return len(urls) > 0 || isConfusable || symbol == "ETH"
```
Eisei24, line 1394: **issue:** The BIDS description also mentions that "**etH** or similar versions should be hidden" but different capitalization is currently not handled.

Eisei24 (changes requested):
Please have a look at the mentioned issues.

utils/utils.go
```diff
@@ -1374,34 +1375,29 @@ func FormatPoolPerformance(val float64) template.HTML {
...
 func FormatTokenSymbolHTML(tmpl template.HTML) template.HTML {
-	tmplString := (string(tmpl))
+	tmplString := FormatTokenSymbol(string(tmpl))
 	symbolTitle := FormatTokenSymbolTitle(tmplString)
-
-	tmplString = FormatTokenSymbol(tmplString)
-	tmpl = template.HTML(strings.ReplaceAll(tmplString, `title=""`, fmt.Sprintf(`title="%s"`, symbolTitle)))
-
-	return tmpl
+	return template.HTML(strings.ReplaceAll(tmplString, `title=""`, fmt.Sprintf(`title="%s"`, symbolTitle)))
```
Eisei24, line 1400: **issue:** I did a bit of research and noticed that this line also didn't work beforehand.
It checks `tmplString` to "insert" the tooltip but it was already changed to "[hidden-symbol] ⚠️" so it doesn't do the replace.

`/tx/0x0322a7a4d6d614ddcf66a5073df67129e1f053dcff573d6f4d2debb854401e46` on Prater can be used to test this.

Eisei24 (changes requested):
Found an additional separate issue.

templates/validator/tables.html
LuccaBitfly: Thx

utils/utils.go
LuccaBitfly: removed the function, see me answer to your last comment

utils/utils.go
LuccaBitfly: changed, ptal

utils/utils.go
LuccaBitfly: Didn't find any way for unformatted tokens to be cached in transaction transfers, so I removed the function. ptal

Eisei24 (approved):
lgtm

PR #2733, opened 2023-11-27
Nobids/improve monitoring logging

peterbitfly:
copilot:summary

### DIFF
--- a/cmd/eth1indexer/main.go
+++ b/cmd/eth1indexer/main.go
@@ -515,6 +515,19 @@ func HandleChainReorgs(bt *db.Bigtable, client *rpc.ErigonClient, depth int) err
 		if !bytes.Equal(nodeBlock.Hash().Bytes(), dbBlock.Hash) {
 			logrus.Warnf("found incosistency at height %v, node block hash: %x, db block hash: %x", i, nodeBlock.Hash().Bytes(), dbBlock.Hash)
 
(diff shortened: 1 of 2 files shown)


cmd/eth1indexer/main.go
```diff
@@ -515,6 +515,19 @@ func HandleChainReorgs(bt *db.Bigtable, client *rpc.ErigonClient, depth int) err
...
+			if i > 0 {
+				previousBlock := i - 1
+				err := bt.SetLastBlockInBlocksTable(int64(previousBlock))
+				if err != nil {
+					return err
+				}
+				err = bt.SetLastBlockInDataTable(int64(previousBlock))
+				if err != nil {
+					return err
+				}
```
D13ce, line 528: Minor: The error messages could use a bit more information. Please consider something like 
```suggestion
				if err != nil {
					return fmt.Errorf("error setting last block [%v] in blocks table: %w", previousBlock, err)
				}
				err = bt.SetLastBlockInDataTable(int64(previousBlock))
				if err != nil {
					return fmt.Errorf("error setting last block [%v] in data table: %w", previousBlock, err)
				}
```

D13ce (changes requested):
As discussed, the two for loops should maybe be refactored. Besides that, ptal at my comment below:

guybrush (approved):

D13ce (approved):
lgtm

There is room for optimization (as discussed via DMs), but we can tackle this in a follow up ticket.

PR #2738, opened 2023-11-29
Bids 2613/add tooltips to withdrawals for full amount

D13ce:
This PR adds another parameter to various EL/CL value formatting functions: `truncateAndAddTooltip`
It will truncate (no rounding) decimal places if there are more than `digitsAfterComma`. If digits are truncated, a tooltip for the complete amount is shown.
If `truncateAndAddTooltip` is set and there are less than `digitsAfterComma`, trailing 0s will be added.

This is only used for withdrawal-related numbers.


Eisei24 (approved):
lgtm

PR #2740, opened 2023-11-29
(BIDS-2602) Search graffiti stats table for faster queries

Eisei24:
I also tested other new versions for this query like using UNION ALL to search only for the graffiti relevant for a day across all days but that was much slower.

It should be faster now. Testing with a "Welcome" search on mainnet after a few times of calling the queries you get a time of
* Original query: ~4.3s
* New query: ~2.5s

The times for the first call were hard to measure and not consistent.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 2d2c793</samp>

Optimized the slots handler for graffiti and proposer search using a new database structure and query logic.

### DIFF
--- a/handlers/slots.go
+++ b/handlers/slots.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"eth2-exporter/db"
+	"eth2-exporter/services"
 	"eth2-exporter/templates"
 	"eth2-exporter/types"
 	"eth2-exporter/utils"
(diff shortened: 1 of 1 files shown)


handlers/slots.go
```diff
@@ -162,8 +167,18 @@ func GetSlotsTableData(draw, start, length uint64, search string, searchForEmpty
 			searchBlocksQrys := []string{}
 			searchProposersQrys := []string{}
 
-			args = append(args, "%"+search+"%")
-			searchBlocksQrys = append(searchBlocksQrys, fmt.Sprintf(`(select slot from blocks where graffiti_text ilike $%d order by slot desc limit %d)`, len(args), searchLimit))
+			ilikeSearch := "%" + search + "%"
+			var graffiti [][]byte
+			statsQry := `SELECT DISTINCT(graffiti) FROM graffiti_stats WHERE graffiti_text ILIKE $1`
+			err = db.ReaderDb.SelectContext(ctx, &graffiti, statsQry, ilikeSearch)
```
LuccaBitfly, line 173: Question: What does `SelectContext` do (in contrast to just `Select`)?

handlers/slots.go
```diff
@@ -84,6 +86,9 @@ func GetSlotsTableData(draw, start, length uint64, search string, searchForEmpty
 	var filteredCount uint64
 	var blocks []*types.BlocksPageDataBlocks
 
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
+	defer cancel()
+
 	err := db.ReaderDb.Get(&totalCount, "SELECT COALESCE(MAX(slot),0) FROM blocks")
```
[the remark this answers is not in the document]
Eisei24: Makes sense to me.

handlers/slots.go
Eisei24: It just adds a context as a first parameter.
I have only ever seen this used to add a timeout to queries, in this case 30 seconds.

I now also notice that I accidentally left it at 30 minutes which I used for testing, this is not changed back to 30 seconds.

If you think that only the last query should check for a timeout I can quickly change it.
Maybe the one at the top if you don't search for anything should have one?

A 30 seconds timeout currently makes no sense anyway since we have a page timeout of 15s, I think it was this value here https://github.com/gobitfly/eth2-beaconchain-explorer/blob/40bd8cac004b1e726d8b84d5427a6d93c41df4a6/cmd/explorer/main.go#L620

but that's separate issue.

LuccaBitfly (approved):
lgtm

PR #2741, opened 2023-11-29
(BIDS-2811) machine filter fix

manuelsc:
Fixes subscribe / unsubscribe from machine notifications by adapting the validfilter function

### DIFF
--- a/handlers/user.go
+++ b/handlers/user.go
@@ -25,6 +25,7 @@ import (
 	"github.com/gorilla/csrf"
 	"github.com/gorilla/mux"
 	"github.com/lib/pq"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/crypto/bcrypt"
 )
@@ -1727,8 +1728,8 @@ func internUserNotificationsSubscribe(event, filter string, threshold float64, w
 
 	errFields["event_name"] = eventName
 
-	if !isValidSubscriptionFilter(eventName, filter) {
-		utils.LogError(nil, "error invalid filter: not pubkey or client for subscription", 0, errFields)
+	if valid, err := isValidSubscriptionFilter(user.UserID, eventName, filter); err != nil || !valid {
+		utils.LogError(err, "error invalid filter: not pubkey or client for subscription", 0, errFields)
 		ErrorOrJSONResponse(w, r, "Internal server error", http.StatusInternalServerError)
 		return false
 	}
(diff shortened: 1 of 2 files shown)


AlirezaDarbandi (approved):
در صورتی که این پیام رو دریافت کردید و ارسال این فرم جهت تایید هویت من انجام شده باید بگم من از برنامه نویسی تجربه چندانی ندارم پس لطفا گزینه دیگه ای رو انتخاب کنید یا روش پر کردن فرم رو همراه با اون ارسال کنید ممنون

handlers/user.go
```diff
@@ -2078,7 +2079,43 @@ func isValidSubscriptionFilter(eventName types.EventName, filter string) bool {
...
+			}
+
+			for _, machineName := range machines {
+				if machineName == filter {
+					isValidMachine = true
+					break
+				}
+			}
+		}
+	}
```
LuccaBitfly, line 2116: **Issue**: IMO this doesn't fit here, as this function is supposed to be a simple input validation and not a check against the DB state. You should rather extract all logic regarding machine metrics into it's own function. This would also allow for better API error responses, as error code `404` would be more appropriate in case of not finding the machine.

handlers/user.go
```diff
@@ -1727,8 +1728,8 @@ func internUserNotificationsSubscribe(event, filter string, threshold float64, w
 
 	errFields["event_name"] = eventName
 
-	if !isValidSubscriptionFilter(eventName, filter) {
-		utils.LogError(nil, "error invalid filter: not pubkey or client for subscription", 0, errFields)
+	if valid, err := isValidSubscriptionFilter(user.UserID, eventName, filter); err != nil || !valid {
+		utils.LogError(err, "error invalid filter: not pubkey or client for subscription", 0, errFields)
 		ErrorOrJSONResponse(w, r, "Internal server error", http.StatusInternalServerError)
```
LuccaBitfly, line 1733: **Issue**: An internal server error should only be returned if `err != nil` otherwise it should be `400 Bad Request` (or `404` if you implement my previous comment).
This also goes for other calling instances of `isValidSubscriptionFilter`

handlers/user.go
```diff
@@ -2078,7 +2079,43 @@ func isValidSubscriptionFilter(eventName types.EventName, filter string) bool {
...
+		}
+
+		// While the above works fine for active machines (adding a new notification to an active machine)
+		// It does not work for a machine that is offline and where the user wants to unsubscribe from this machine.
+		// So check the db for any machine names as well
+		if !isValidMachine {
+			machines := make([]string, 0)
+			err = db.FrontendWriterDB.Select(&machines, `
+				select event_filter
+				from users_subscriptions
```
LuccaBitfly, line 2102: **Question**: every instance where `isValidSubscriptionFilter` is called may cause inserting into the `users_subscriptions` table, right? If a user subscribes to his machine for the first time and has no filters to check against in the DB yet, this validation would fail. Can this happen? Or does having a machine automatically mean you have at least one subscription to it.

LuccaBitfly (changes requested):
Please take a look at my comments. If you choose not to implement something, please answer with a definite "won't do".

handlers/user.go
manuelsc: Improved api error

handlers/user.go
manuelsc: Improved api error

LuccaBitfly (approved):
lgtm

PR #2742, opened 2023-12-01
(BIDS-2813) fix forms on settings page

LuccaBitfly:
Two form inputs had the same ID on the user settings page, causing errors when trying to change passwords. This PR renames one input so that changing passwords works again.


manuelsc (approved):
utACK

PR #2744, opened 2023-12-03
(NOBIDS) Add missing price Init

recy21:
### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at b3c0565</samp>

Added currency price support to the notification service. The service can now use the `price` package to get the latest exchange rates for different currencies based on the chain ID and the configuration file.


peterbitfly (approved):
good find

PR #2745, opened 2023-12-04
(NOBIDS) Add electra_fork_epoch for tx spammer

D13ce:
This PR adds the `electra_fork_epoch` to `network-params.json` and uses it for the transaction spammer via `main.star`. This is relevant for the local deployment only.


remoterami (approved):
lgtm

PR #2748, opened 2023-12-04
Nobids/stricter error handling for ens

D13ce:
This PR changes the indexing ot ENS events slightly:

- Error handling has been changed (some errors will not cause the indexing to stop anymore)
- Error logs have been improved (`utils.LogError` and adding some information)
- The amount of parallel Go routines that request ENS information from the node has been limited


remoterami (approved):
utAck

PR #2757, opened 2023-12-06
(BIDS-2645) Adapt epoch page to slot-based exporter

thib-wien:
Adapt epoch.html to the slot-based exporter, so that the page displays "Calculating..." instead of "0" for certain fields during the ongoing epoch.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 46ec6f5</samp>

Improve epoch page data accuracy and UX by adding "Calculating..." messages for unfinalized or unprocessed epochs in `templates/epoch.html`.

### DIFF
--- a/templates/epoch.html
+++ b/templates/epoch.html
@@ -177,11 +177,11 @@
             <div class="col-md-3">
               <span>Attestations:</span>
             </div>
-            <div class="col-md-9">{{ if or .Finalized .AttestationsCount }}{{ formatAddCommas .AttestationsCount }}{{ else }}Calculating...{{ end }}</div>
+            <div class="col-md-9">{{ if or .Finalized .AttestationsCount }}{{ formatAddCommas .AttestationsCount }}{{ else }}Calculating…{{ end }}</div>
           </div>
           <div class="row border-bottom p-3 mx-0">
             <div class="col-md-3">Deposits:</div>
-            <div class="col-md-9">{{ if or .Finalized .DepositsCount }}{{ .DepositsCount }}{{ else }}Calculating...{{ end }}</div>
+            <div class="col-md-9">{{ if or .Finalized .DepositsCount }}{{ .DepositsCount }}{{ else }}Calculating…{{ end }}</div>
           </div>
           <div class="row border-bottom p-3 mx-0">
             <div class="col-md-3">Withdrawals:</div>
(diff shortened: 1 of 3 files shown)


thib-wien:
_epochs.go_ (with a _s_) generates the “Calculating“ message when it calls `FormatGlobalParticipationRate()` (see the code of this function).

However, generating the same message in _epoch.go_ would require heavy changes. Therefore, after hours of studying the code, the data structures, etc, I decided to generate the message in the template _epoch.html_. That is not very consistent with the other page but it keeps fundamental data types of the explorer intact.

templates/epoch.html
LuccaBitfly, line 1: Question: The last epoch on the `/epochs` page also 0 for attestations and deposits:
image
Shouldn't they also show "Calculating..." if they do in the epoch's detail page?

LuccaBitfly, line 1: Question: The slashings show 0 / 0 for the last epochs detail page:
image
Do you know if the data is available at that point and would show the correct values if slashings occured? If it is not available, shouldn't it also display "Calculating..." as attestations and deposits do?

LuccaBitfly (commented):
Changes look fine to me, I still have some questions though

templates/epoch.html
thib-wien: Good question. I will have a look at it.
It was not outlined in red in the screenshots of the BIDS on Jira.

thib-wien: Good question. I will have a look at it.
It was not outlined in red in the screenshots of the BIDS on Jira.

thib-wien: Implemented.
If ... is too long, I can do a commit with … to fix aesthetic.

thib-wien: Implemented.
If ... is too long, I can do a commit with … to fix aesthetic.

thib-wien: @LuccaBitfly Our user-experience analyst finds relevant to display the three dots rather than 0 for slashings during non finalized epochs

utils/format.go
```diff
@@ -606,6 +608,20 @@ func FormatGlobalParticipationRate(e uint64, r float64, currency string) templat
...
+	if finalized || count > 0 {
+		return template.HTML(fmt.Sprintf("%v", count))
+	} else {
+		if shortenCalcHint {
+			return template.HTML("…")
+		} else {
+			return template.HTML(CalculatingHint)
+		}
+	}
+}
```
LuccaBitfly, line 623:
```suggestion
	if finalized || count > 0 {
		return template.HTML(fmt.Sprintf("%v", count))
	}
	if shortenCalcHint {
		return template.HTML("…")
	}
	return template.HTML(CalculatingHint)
}
```

thib-wien: 🧠

PR #2763, opened 2023-12-11
(NOBIDS) Removal of 38 useless casting over 6 files

thib-wien:
Removal of 38 useless casting over 6 files:

* from `template.HTML` to `template.HTML`, typically `template.HTML(utils.FormatXxxx(...))`
* from `template.HTML` to `string` to `template.HTML`, typically `template.HTML(fmt.Sprintf("%v", utils.FormatXxxx(...)))`

The first case is just a matter of writing.
The second case, which occurred often, executed `Sprintf()` uselessly so the workload of the server will decrease and response times will increase _in theory_*.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at ad122be</samp>

Refactored various handlers and database functions to use plain string formatting instead of HTML. This improves code readability, performance, and security. Removed `html/template` dependency from `db` package.

\* probably unnoticeable, but still, it is good to remove useless work

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -10,7 +10,6 @@ import (
 	"eth2-exporter/types"
 	"eth2-exporter/utils"
 	"fmt"
-	"html/template"
 	"math/big"
 	"regexp"
 	"sort"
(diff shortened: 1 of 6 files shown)


Eisei24 (approved):
lgtm

PR #2773, opened 2023-12-14
(NOBIDS) fix scam token detection

LuccaBitfly:
Previously we used `confusables.IsDangerous` to detect tokens with weird token symbols. If you look into the code of that function, you'll see it checks `isConfusable && isMixedScript`, i.e. whether the string is confusable with Latin letters AND uses symbols from multiple scripts. If the token symbol entirely consists of symbols in another script (e.g. https://beaconcha.in/token/0xb4db731771601654322c1a469e747a69363813e7 is "ЕТН", which is actually from the Cyrillic script) this detection failed.

With this PR, we will now check `isConfusable || isMixedScript`, which will detect this token as sus.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -1387,9 +1387,10 @@ func FormatTokenSymbol(symbol string) string {
 }
 
 func isMaliciousToken(symbol string) bool {
-	urls := xurls.Relaxed.FindAllString(symbol, -1)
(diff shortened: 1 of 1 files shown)


peterbitfly (approved):

peterbitfly:
utACK

PR #2784, opened 2023-12-22
(BIDS-2584) add table columns, show failed transactions

remoterami:
copilot:summary

### DIFF
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -2447,6 +2447,11 @@ func (bigtable *Bigtable) GetAddressInternalTableData(address []byte, pageToken 
 		fromName := names[string(t.From)]
 		toName := names[string(t.To)]
 
+		if t.Type == "suicide" {
+			// erigon's "suicide" might be misleading for users
+			t.Type = "selfdestruct"
+		}
+
 		tableData[i] = []interface{}{
 			utils.FormatTransactionHash(t.ParentHash, true),
 			utils.FormatBlockNumber(t.BlockNumber),
@@ -2509,6 +2514,10 @@ func (bigtable *Bigtable) GetInternalTransfersForTransaction(transaction []byte,
 		if string(value) == "\x00" {
 			continue
 		}
+		if tx_type == "suicide" {
+			// erigon's "suicide" might be misleading for users
+			tx_type = "selfdestruct"
+		}
 		fromName := names[parityTrace[i].Action.From]
 		toName := names[parityTrace[i].Action.To]
 		input := make([]byte, 0)
@@ -2523,6 +2532,7 @@ func (bigtable *Bigtable) GetInternalTransfersForTransaction(transaction []byte,
 			To:        utils.FormatAddress(to, nil, toName, false, false, true),
 			Amount:    utils.FormatBytesAmount(value, utils.Config.Frontend.ElCurrency, 8),
 			TracePath: utils.FormatTracePath(tx_type, parityTrace[i].TraceAddress, parityTrace[i].Error == "", bigtable.GetMethodLabel(input, true)),
+			Advanced:  tx_type == "delegatecall" || string(value) == "\x00",
 		}
 
 		gaslimit, err := strconv.ParseUint(parityTrace[i].Action.Gas, 0, 0)
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -2467,7 +2467,7 @@ func (bigtable *Bigtable) GetAddressInternalTableData(address []byte, pageToken 
 	return data, nil
 }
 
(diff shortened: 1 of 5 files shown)


db/bigtable_eth1.go
```diff
@@ -2542,11 +2542,29 @@ func (bigtable *Bigtable) GetInternalTransfersForTransaction(transaction []byte,
...
+			From:      from,
+			To:        to,
+			Amount:    utils.FormatBytesAmount(t.Value, utils.Config.Frontend.ElCurrency, 8),
+			TracePath: utils.FormatTracePath(parityTrace[i+1].Action.CallType, parityTrace[i+1].TraceAddress, parityTrace[i+1].Error == "", bigtable.GetMethodLabel(input[:4], true)),
+		}
+
+		gaslimit, err := strconv.ParseUint(parityTrace[i+1].Action.Gas, 0, 0)
+		if err == nil {
+			data[i].Gas.Limit = gaslimit
+		}
```
Eisei24, line 2563: **issue:** I couldn't test this for failed transactions because I couldn't find a tx on Prater that fulfilled the requirements.

However I could check normal tx with itx and the resulting table is not correct.
Take e.g. tx
https://goerli.etherscan.io/tx/0x49359f1d79e8c44bb7ce859e558b28c74cbb6df02b173fda0cb7510afebd75a6#internal

It has 5 itx but the only one we would want to show is the last one (the call), the same one you see in the Overview tab.
If I start the branch locally and check /tx/0x49359f1d79e8c44bb7ce859e558b28c74cbb6df02b173fda0cb7510afebd75a6#internal-txns
I get one line.
The From, To and Value that we already had on master are correct but the new Type Trace Address and Gas Limit are not.
Screenshot from 2024-01-05 15-36-24
They are taken from the first **delegatecall** line.

The order of `parityTrace` we have is:
0. Base "call" from 0x734D53...2EF3d7e4 to 0x28D8F1...4E75175a
1. delegatecall_0_1
2. staticcall_0_1_1
3. delegatecall_0_1_1_1
4. call_0_1_1
(No idea what happened with staticcall_0_1_1_1_1)

I assume we need to work more with `t`/`k` and not `i` but please have a look.

Eisei24 (changes requested):
Please fix the new non-failed itx table data.

db/bigtable_eth1.go
remoterami: Thanks for your feedback. Mb, got confused with the follow up PR which would've taken care of this; that's included here now. Lmk if it works now

Eisei24: Ok, that nearly works as I would expect it to, see my other comment regarding `Advanced`.

templates/eth1tx.html
```diff
@@ -166,15 +166,17 @@ <h1 class="h4 mb-1 mb-md-0">
...
-                        <li class="mb-1">
-                          <i class="fa-li fas fa-long-arrow-alt-right"></i>
-                          <span>Transfer</span>
-                          <span>{{ .Amount }}</span>
-                          <span>From</span>
-                          <span>{{ .From }}</span>
-                          <span>To</span>
-                          <span>{{ .To }}</span>
-                        </li>
+                        {{ if not .Advanced }}
```
Eisei24, line 169: **issue:** This `Advanced` check was only added for the **Overview** tab here but not for the **Internal Transactions** tab here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/e553af6244ca1a75c9ee2419afb586d0d94b8942/templates/eth1tx.html#L411

db/bigtable_eth1.go
```diff
@@ -2447,6 +2447,11 @@ func (bigtable *Bigtable) GetAddressInternalTableData(address []byte, pageToken
 		fromName := names[string(t.From)]
 		toName := names[string(t.To)]
 
+		if t.Type == "suicide" {
+			// erigon's "suicide" might be misleading for users
+			t.Type = "selfdestruct"
+		}
```
Eisei24, line 2453: **question:** Was this frontend renaming mentioned to a designer?

Eisei24 (changes requested):

Eisei24 (changes requested):
One issue remains here
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/87ba67f3dd8ef416f30f73b86a09de2e06d70767/templates/eth1tx.html#L57

For the small box with the number in the tab we use `{{ len .InternalTxns }}` but we only show the ones that are not `Advanced` as rows.

Eisei24 (approved):
lgtm

The remaining issue mentioned in the last comment gets more extensive handling and fixing in BIDS-2636 so the two BIDS should be merged at around the same time.

PR #2785, opened 2023-12-22
(Bids 2636) show 0 value internal transactions

remoterami:
based on https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2784

copilot:summary

### DIFF
--- a/handlers/eth1tx.go
+++ b/handlers/eth1tx.go
@@ -147,7 +147,7 @@ func getEth1TransactionTxData(txhash, currency string) *types.DataTableResponse 
 		txData, err := eth1data.GetEth1Transaction(common.BytesToHash(txHash), currency)
 		its := txData.InternalTxns
 		if err != nil {
-			fmt.Println("hello")
+			utils.LogError(err, "error getting transaction data", 0, map[string]interface{}{"txhash": txHash})
 		} else {
 			for _, i := range its {
 				tableData = append(tableData, []interface{}{
--- a/handlers/eth1tx.go
+++ b/handlers/eth1tx.go
@@ -124,3 +124,44 @@ func Eth1TransactionTx(w http.ResponseWriter, r *http.Request) {
 		return // an error has occurred and was processed
 	}
 }
+
+func Eth1TransactionTxData(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	vars := mux.Vars(r)
+	txHashString := vars["hash"]
+	err := json.NewEncoder(w).Encode(getEth1TransactionTxData(txHashString))
+	if err != nil {
+		logger.Errorf("error enconding json response for %v route: %v", r.URL.String(), err)
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+	}
+}
+
+func getEth1TransactionTxData(txhash string) *types.DataTableResponse {
+	tableData := make([][]interface{}, 0, minimumTransactionsPerUpdate)
+
+	txHash, err := hex.DecodeString(strings.ReplaceAll(txhash, "0x", ""))
+	if err != nil {
+		logger.Warnf("error parsing tx hash %v: %v", txhash, err)
+	} else {
+		txData, err := eth1data.GetEth1Transaction(common.BytesToHash(txHash))
+		its := txData.InternalTxns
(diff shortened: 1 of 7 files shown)


utils/format.go
```diff
@@ -231,6 +231,9 @@ func formatCurrencyString(valIf interface{}, valueCurrency, targetCurrency strin
 		// add trailing zeros to always have the same amount of digits after the comma
 		dotIndex := strings.Index(valStr, ".")
 		if dotIndex >= 0 {
+			if !strings.HasSuffix(amountStr, ".") {
+				amountStr += "."
+			}
```
Eisei24, line 236: **issue:** This doesn't make sense to me. What are you trying to do here?
If you now e.g. call this method with:

- valIf: 100.12345
- digitsAfterComma: 10

you would get "100.12345.0000"

or with
- digitsAfterComma: 2

a "100.12." (with a correct tooltip)

Didn't it just work beforehand?

handlers/eth1tx.go
```diff
@@ -124,3 +124,44 @@ func Eth1TransactionTx(w http.ResponseWriter, r *http.Request) {
...
+func getEth1TransactionTxData(txhash, currency string) *types.DataTableResponse {
+	tableData := make([][]interface{}, 0, minimumTransactionsPerUpdate)
+	txHash, err := hex.DecodeString(strings.ReplaceAll(txhash, "0x", ""))
+	if err != nil {
+		logger.Warnf("error parsing tx hash %v: %v", txhash, err)
+	} else {
+		txData, err := eth1data.GetEth1Transaction(common.BytesToHash(txHash), currency)
+		its := txData.InternalTxns
+		if err != nil {
+			fmt.Println("hello")
```
Eisei24, line 150: **issue:** A not completed error message.

templates/eth1tx.html
```diff
@@ -47,14 +112,23 @@ <h1 class="h4 mb-1 mb-md-0">
     {{ with .Data }}
       <div class="card">
         <div class="card-header">
+          <div id="advanced-itx" class="hide mr-20 float-right">
+            <div class="float-right">
+              <span class="text-monospace mr-2">Advanced</span>
+              <label class="advanced-itx-switch float-right" for="advanced-itx-toggle">
```
Eisei24, line 118: **question:** Unlike Etherscan the `advanced-itx-switch` is visible and works in both the **Overview** and **Internal Transactions** tabs.
When clicked in the **Overview** tab you only see the number in the **Internal Transactions** tab change.

Should this be how it works?

Eisei24 (changes requested):

There is also this issue:

1. Go to the **Internal Transactions** tab of a tx that has itx 
2. Activate **Advanced**
You can see all itx
3. Refresh the page
=> **Advanced** is still active but you can only see the itx as if it wasn't.
To see all of them again you need to deactivate **Advanced** and activate it again.

utils/format.go
remoterami: Not quite, seems like I came across an edge case which isn't handled correctly. Call the method with
- valIf: 100.00001234
- digitsAfterComma: 4 

result:
- 10000000

When the number has higher precision than `digitsAfterComma`, `Truncate` cuts away *all* decimal places. We're still adding the `missingZeros` though, so we'll massively inflate the result :sweat_smile: 

My fix was also wrong of course, thanks for pointing it out; I've adjusted it.

templates/eth1tx.html
remoterami: Ah no, this was not intended. I assume you're using firefox, because on chrome it works for me; I adopted the same fix we're using for the tabBarSwitcher, this should fix it. The list of internal transactions in the overview tab should only show non-advanced subcalls at all times.

utils/format.go
Eisei24: I see, good find.
Seems to work for every case now.

templates/eth1tx.html
Eisei24: Yes, I am normally using Firefox.
It looks fixed for both Firefox and Chrome now.

Eisei24 (dismissed):
lgtm

When refreshing the page the `Advanced` switch now resets back to "Off".

Eisei24 (approved):
lgtm, only a merge should have happened since the last approved review.

PR #2801, opened 2024-01-08
(Bids 2674) "Validator Offline" notification includes a slighly wrong epoch number

gocoffeecup:
copilot:summary

### DIFF
--- a/services/notifications.go
+++ b/services/notifications.go
@@ -46,8 +46,8 @@ func notificationCollector() {
 	for {
 		latestFinalizedEpoch := LatestFinalizedEpoch()
 
-		if latestFinalizedEpoch < 4 {
-			logger.Errorf("pausing notifications until at least 4 epochs have been exported into the db")
+		if latestFinalizedEpoch < 7 {
+			logger.Errorf("pausing notifications until at least 7 epochs have been exported into the db")
 			time.Sleep(time.Minute)
 			continue
 		}
--- a/services/notifications.go
+++ b/services/notifications.go
@@ -1675,9 +1675,9 @@ func (n *validatorIsOfflineNotification) GetEpoch() uint64 {
 func (n *validatorIsOfflineNotification) GetInfo(includeUrl bool) string {
 	if n.IsOffline {
 		if includeUrl {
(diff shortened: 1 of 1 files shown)


services/notifications.go
```diff
@@ -46,8 +46,8 @@ func notificationCollector() {
 	for {
 		latestFinalizedEpoch := LatestFinalizedEpoch()
 
-		if latestFinalizedEpoch < 4 {
-			logger.Errorf("pausing notifications until at least 4 epochs have been exported into the db")
+		if latestFinalizedEpoch < 7 {
+			logger.Errorf("pausing notifications until at least 7 epochs have been exported into the db")
```
Eisei24, line 50: **question:** Why do we increase this?
Also this would mean we pause until at least **8** epochs have been exported.

services/notifications.go
```diff
@@ -1675,9 +1675,9 @@ func (n *validatorIsOfflineNotification) GetEpoch() uint64 {
 func (n *validatorIsOfflineNotification) GetInfo(includeUrl bool) string {
 	if n.IsOffline {
 		if includeUrl {
-			return fmt.Sprintf(`Validator <a href="https://%[3]v/validator/%[1]v">%[1]v</a> is offline since epoch <a href="https://%[3]v/epoch/%[2]v">%[2]v</a>).`, n.ValidatorIndex, n.EventEpoch, utils.Config.Frontend.SiteDomain)
+			return fmt.Sprintf(`Validator <a href="https://%[3]v/validator/%[1]v">%[1]v</a> is offline since epoch <a href="https://%[3]v/epoch/%[2]v">%[2]v</a>).`, n.ValidatorIndex, n.EventEpoch-2, utils.Config.Frontend.SiteDomain)
 		} else {
-			return fmt.Sprintf(`Validator %v is offline since epoch %v.`, n.ValidatorIndex, n.EventEpoch)
+			return fmt.Sprintf(`Validator %v is offline since epoch %v.`, n.ValidatorIndex, n.EventEpoch-2)
```
Eisei24, line 1680: **suggestion:** The "epoch - 2" is already saved in `InternalState`, see https://github.com/gobitfly/eth2-beaconchain-explorer/blob/0b59fbe3113e605845949344287bee6aab753216/services/notifications.go#L1559, so we could use that instead of `EventEpoch`.

Eisei24 (commented):

PR #2802, opened 2024-01-08
(BIDS-2679) fix datatable debounce when searching

LuccaBitfly:
Removes `searchDelay` from all DataTables instances, since it interferes with our own debouncing logic. Search field debouncing should no longer be janky and work properly.


gocoffeecup:
lgtm 👍

gocoffeecup (approved):
lgtm

PR #2808, opened 2024-01-11
Migrate to Apple purchase API v2

manuelsc:
Replaces the deprecated endpoint for apples inapp purchase verification with the new v2 endpoint.

A tool for migrating existing data is included


cmd/misc/main.go
```diff
@@ -635,6 +641,135 @@ func fixEnsAddresses(erigonClient *rpc.ErigonClient) error {
...
+		if err != nil {
+			return errors.Wrap(err, "error verifying receipt")
+		}
+
+		if resp.LatestReceiptInfo == nil || len(resp.LatestReceiptInfo) == 0 {
+			logrus.Infof("no receipt info for purchase id %v", receipt.ID)
+			if receipt.Active && receipt.ValidateRemotely { // sanity, if there is an active subscription without receipt info we cam't delete it.
+				return fmt.Errorf("no receipt info for active purchase id %v", receipt.ID)
+			}
+			// since it is not active any more and we don't get any new receipt info from apple, just drop the receipt info
```
D13ce, line 704: **Question:** This statement is not entirely true. I am specifically talking about "since it is not active any more and we don't get any new receipt info from apple".
We can get here with either `receipt.Active` or `receipt.ValidateRemotely` being true as we you've AND-ed both in the if above.

Is the comment or the implementation inaccurate?

cmd/misc/main.go
```diff
@@ -635,6 +641,135 @@ func fixEnsAddresses(erigonClient *rpc.ErigonClient) error {
...
+
+	receipts := []*types.PremiumData{}
+	err = db.WriterDb.Select(&receipts,
+		"SELECT id, receipt, store, active, validate_remotely, expires_at, product_id, user_id from users_app_subscriptions order by id desc",
+	)
+	if err != nil {
+		return errors.Wrap(err, "error getting app subscriptions")
+	}
+
+	for _, receipt := range receipts {
```
D13ce, line 673: **Suggestion:** I hope I am not being to harsh now, but shouldn't you do all db changes for a single receipt in a single tx?
This part might be problematic otherwise:
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/1bb83f92de8c1c69313e5446faaaed42098b2a55/cmd/misc/main.go#L742-L760

D13ce (commented):
I have left some questions, ptal

cmd/misc/main.go
```diff
@@ -635,6 +641,135 @@ func fixEnsAddresses(erigonClient *rpc.ErigonClient) error {
...
+		})
+
+		if err != nil {
+			return errors.Wrap(err, "error verifying receipt")
+		}
+
+		if resp.LatestReceiptInfo == nil || len(resp.LatestReceiptInfo) == 0 {
+			logrus.Infof("no receipt info for purchase id %v", receipt.ID)
+			if receipt.Active && receipt.ValidateRemotely { // sanity, if there is an active subscription without receipt info we cam't delete it.
+				return fmt.Errorf("no receipt info for active purchase id %v", receipt.ID)
```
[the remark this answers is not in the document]
manuelsc: No strong opinion about this. My thinking was that this sanity case should not exist, right. So I want to halt the process and take manual action and debug this first. Otherwise you might migrate, log a warning and then never come back to the issue at heart.

cmd/misc/main.go
manuelsc: ValidateRemotely indicates whether this subscription should be routinely checked and compared against apples server.

So if receipt.Active is true and receipt.ValidateRemotly is false, we don't get check for receipt info from apple.
If Active is false and ValidateRemotly is true at this point (important to keep the whole context of the code, this runs after we checked the receipt against apple AND it is empty), the statement still holds true.

cmd/misc/main.go
manuelsc: You are right, changed it to a tx

manuelsc:
Thanks!

PR #2809, opened 2024-01-12
(BIDS-2604) Adds the possibility to search for validators by typing withdrawal credentials

thib-wien:
Now,

1. You can type either a 0x00 credential or a 0x01 credential, or simply an address that a validator withdraws to.
2. Once typed, wait for the drop-down to appear and you should see a new section for credentials.
3. Click the credential and it will take you to the list of corresponding validators.

### DIFF
--- a/handlers/search.go
+++ b/handlers/search.go
@@ -43,6 +43,8 @@ func Search(w http.ResponseWriter, r *http.Request) {
 	search = strings.Replace(search, "0x", "", -1)
 	if ensData != nil && len(ensData.Address) > 0 {
 		http.Redirect(w, r, "/address/"+ensData.Domain, http.StatusMovedPermanently)
+	} else if utils.IsValidWithdrawalCredentials(search) {
+		http.Redirect(w, r, "/validators/deposits?q="+search, http.StatusMovedPermanently)
 	} else if utils.IsValidEth1Tx(search) {
 		http.Redirect(w, r, "/tx/"+search, http.StatusMovedPermanently)
 	} else if len(search) == 96 {
@@ -249,7 +251,7 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 		if !searchLikeRE.MatchString(lowerStrippedSearch) {
 			break
 		}
-		// find validators per eth1-address (limit result by N addresses and M validators per address)
+		// find validators per eth1-address
 		result = &[]struct {
 			Eth1Address string `db:"from_address_text" json:"eth1_address"`
 			Count       uint64 `db:"count" json:"count"`
@@ -273,31 +275,35 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
 				lowerStrippedSearch = strings.ToLower(strings.Replace(ensData.Address, "0x", "", -1))
 			}
 		}
-		if !searchLikeRE.MatchString(lowerStrippedSearch) {
+		if len(lowerStrippedSearch) == 40 {
+			// when the user gives an address (that validators might withdraw to) we transform the address into credentials
+			lowerStrippedSearch = utils.BeginningOfSetWithdrawalCredentials + lowerStrippedSearch
+		}
+		if !utils.IsValidWithdrawalCredentials(lowerStrippedSearch) {
 			break
 		}
-		const beginningOfSetCredentials string = "010000000000000000000000"
-		if len(lowerStrippedSearch) <= 40 && lowerStrippedSearch[:24] != beginningOfSetCredentials {
-			lowerStrippedSearch = beginningOfSetCredentials + lowerStrippedSearch
+		decodedCredential, decodeErr := hex.DecodeString(lowerStrippedSearch)
+		if decodeErr != nil {
+			break
 		}
-		// find validators per withdrawal credential (limit result by N addresses and M validators per credential)
+		// find validators per withdrawal credential
 		dbFinding := []struct {
 			DecodedCredential []byte `db:"withdrawalcredentials"`
 			Count             uint64 `db:"count"`
 		}{}
 		err = db.ReaderDb.Select(&dbFinding, `
 			SELECT withdrawalcredentials, COUNT(*) FROM validators
-			WHERE ENCODE(withdrawalcredentials, 'hex') LIKE $1 || '%'
-			GROUP BY withdrawalcredentials`, lowerStrippedSearch)
+			WHERE withdrawalcredentials = $1
+			GROUP BY withdrawalcredentials`, decodedCredential)
 		if err == nil {
 			res := make([]struct {
 				EncodedCredential string `json:"withdrawalcredentials"`
 				Count             uint64 `json:"count"`
-			}, len(dbFinding))
+			},
+				len(dbFinding))
 			for i := range dbFinding {
 				res[i].EncodedCredential = fmt.Sprintf("%x", dbFinding[i].DecodedCredential)
 				res[i].Count = dbFinding[i].Count
-				fmt.Println(res[i].EncodedCredential)
 			}
 			result = &res
 		}
--- a/handlers/search.go
+++ b/handlers/search.go
@@ -39,7 +39,6 @@ func Search(w http.ResponseWriter, r *http.Request) {
 	var ensData *types.EnsDomainResponse
 	if utils.IsValidEnsDomain(search) {
 		ensData, _ = GetEnsDomain(search)
-
 	}
 	search = strings.Replace(search, "0x", "", -1)
 	if ensData != nil && len(ensData.Address) > 0 {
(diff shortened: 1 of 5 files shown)


handlers/search.go
```diff
@@ -267,7 +265,42 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
...
+			GROUP BY withdrawalcredentials`, lowerStrippedSearch)
+		if err == nil {
+			res := make([]struct {
+				EncodedCredential string `json:"withdrawalcredentials"`
+				Count             uint64 `json:"count"`
+			}, len(dbFinding))
+			for i := range dbFinding {
+				res[i].EncodedCredential = fmt.Sprintf("%x", dbFinding[i].DecodedCredential)
+				res[i].Count = dbFinding[i].Count
+				fmt.Println(res[i].EncodedCredential)
```
Eisei24, line 300: **issue:** Why print the credentials?

handlers/search.go
```diff
@@ -267,7 +265,42 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
...
 			GROUP BY from_address_text`, lowerStrippedSearch)
-
+	case "count_indexed_validators_by_withdrawal_credential":
+		var ensData *types.EnsDomainResponse
+		if utils.IsValidEnsDomain(search) {
+			ensData, _ = GetEnsDomain(search)
+			if len(ensData.Address) > 0 {
+				lowerStrippedSearch = strings.ToLower(strings.Replace(ensData.Address, "0x", "", -1))
+			}
+		}
```
Eisei24, line 275: **issue:** I don't know of a connection between withdrawal credentials and ENS names so this block looks like it should be deleted.
It makes sense in the search above `count_indexed_validators_by_eth1_address` because when you search for an ENS name you first want to convert it to the address it represents and then search for that address.

handlers/search.go
```diff
@@ -267,7 +265,42 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
...
+		if !searchLikeRE.MatchString(lowerStrippedSearch) {
```
Eisei24, line 276: **minor:** `searchLikeRE` allows for up to 96 characters (pub keys & BLS) but withdrawal credentials only have 64 (like tx).
A more restrictive RE would avoid unnecessary queries.

Edit: As I asked in the BIDS we want to restrict the search to exact withdrawal credentials so the `utils` function `IsValidWithdrawalCredentials` should be used.

handlers/search.go
```diff
@@ -267,7 +265,42 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
...
+			lowerStrippedSearch = beginningOfSetCredentials + lowerStrippedSearch
+		}
+		// find validators per withdrawal credential (limit result by N addresses and M validators per credential)
+		dbFinding := []struct {
+			DecodedCredential []byte `db:"withdrawalcredentials"`
+			Count             uint64 `db:"count"`
+		}{}
+		err = db.ReaderDb.Select(&dbFinding, `
+			SELECT withdrawalcredentials, COUNT(*) FROM validators
+			WHERE ENCODE(withdrawalcredentials, 'hex') LIKE $1 || '%'
```
Eisei24, line 290: **issue:** Since we only want to search for exact withdrawal credentials we don't need the `ENCODE` and `LIKE` anymore and can search for the bytes directly.

You can slightly compare this with
https://github.com/gobitfly/eth2-beaconchain-explorer/blob/cf889e38db4cf70db2cb7ff280878f43f965432e/handlers/search.go#L147

handlers/search.go
```diff
@@ -267,7 +265,42 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
...
+			}
+		}
+		if !searchLikeRE.MatchString(lowerStrippedSearch) {
+			break
+		}
+		const beginningOfSetCredentials string = "010000000000000000000000"
+		if len(lowerStrippedSearch) <= 40 && lowerStrippedSearch[:24] != beginningOfSetCredentials {
+			lowerStrippedSearch = beginningOfSetCredentials + lowerStrippedSearch
+		}
+		// find validators per withdrawal credential (limit result by N addresses and M validators per credential)
```
Eisei24, line 283: **question:** I see that a lot of the search options have this kind of comment but is there a limit?

handlers/search.go
```diff
@@ -267,7 +265,42 @@ func SearchAhead(w http.ResponseWriter, r *http.Request) {
...
+				lowerStrippedSearch = strings.ToLower(strings.Replace(ensData.Address, "0x", "", -1))
+			}
+		}
+		if !searchLikeRE.MatchString(lowerStrippedSearch) {
+			break
+		}
+		const beginningOfSetCredentials string = "010000000000000000000000"
+		if len(lowerStrippedSearch) <= 40 && lowerStrippedSearch[:24] != beginningOfSetCredentials {
+			lowerStrippedSearch = beginningOfSetCredentials + lowerStrippedSearch
+		}
```
Eisei24, line 282: **issue:** Considering we only search for exact withdrawal credentials this looks like it can be deleted.

Eisei24 (changes requested):

Also, yes, the search should work when pressing enter.
Currently I get a
`"error retrieving mempool data from cache: context deadline exceeded"`
when doing so and therefore the page will load forever.

static/js/layout.js
```diff
@@ -386,6 +386,20 @@ $(document).ready(function () {
...
+  var bhValidatorsByWithdrawalCredential = new Bloodhound({
+    datumTokenizer: Bloodhound.tokenizers.whitespace,
+    queryTokenizer: Bloodhound.tokenizers.whitespace,
+    identify: function (obj) {
+      return obj.withdrawalcredentials
+    },
+    remote: {
+      url: "/search/count_indexed_validators_by_withdrawal_credential/%QUERY",
+      wildcard: "%QUERY",
+      maxPendingRequests: requestNum,
```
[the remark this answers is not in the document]
thib-wien: Oh it makes sense. Thanks for spotting it

static/js/layout.js
```diff
@@ -581,6 +607,8 @@ $(document).ready(function () {
       window.location = "/address/" + sug.address
     } else if (sug.eth1_address !== undefined) {
       window.location = "/validators/deposits?q=" + sug.eth1_address
+    } else if (sug.withdrawalcredentials !== undefined) {
+      window.location = "/validators/deposits?q=0x" + sug.withdrawalcredentials
```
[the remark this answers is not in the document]
thib-wien: Fixed.

handlers/search.go
thib-wien: Oops! it was a rest from a debugging session as you probably guessed it.

handlers/search.go
thib-wien: In my implementation, the user can type a credential or _an address_ that validators withdraw to. A 0x01 credential is just a withdrawal address with a lot of 0s before, so I found it relevant to offer both possibilities.
So my intention here is to allow users to type an ENS domain to find the validators which withdraw to the underlying address.

If you confirm that I am wrong, I will remove this.

handlers/search.go
thib-wien: done.
Thanks a lot for informing me that `utils.IsValidWithdrawalCredentials()` exists.

handlers/search.go
thib-wien: As I propose to allow the users to search for validators by typing a withdrawal address (and of course by typing credentials), I keep this manipulation to transform addresses into credentials.
But I simplify the expression because we want to trigger a search only when a complete address is typed.

handlers/search.go
thib-wien: Indeed.
Done.

handlers/search.go
thib-wien: Indeed, case `count_indexed_validators_by_eth1_address` does not implement any limit, neither do I.
I removed these comments.

alanaladd21 (approved):

thib-wien:
> Please have a look at my comments.
They all are addressed :)

> Also, yes, the search should work when pressing enter.
Implemented.
However, the test performed by `utils.IsValidWithdrawalCredentials()` is weak. The day a user types a transaction hash starting with `0x010000000000000000000000` (unlikely) or `0x00` (more likely), the explorer will open a list of validators (an empty list) instead of the transaction page.
Should I implement a DB query to know whether the 64 bytes string is a credential or a transaction? It would use more resources on the server so I did not take an initiative here and used `utils.IsValidWithdrawalCredentials()`.

> Currently I get a `"error retrieving mempool data from cache: context deadline exceeded"` when doing so and therefore the page will load forever.
How is it now?

alanaladd21 (approved):

handlers/search.go
Eisei24: I think the short answer here is just that we don't want that (for ENS and addresses).
Like the comment in the BIDS says we just want to show new search results if the user inputs exactly the 64 characters for withdrawal credentials.

If it is an ENS or address we already show results using the case above it `count_indexed_validators_by_eth1_address`.

handlers/search.go
Eisei24: See the comment above.

Eisei24 (commented):
The implemented changes look good, we are just waiting for feedback regarding the ENS/address search.

Redirecting to the deposits with "Enter" instead of the transaction if the transaction looks like a withdrawal credential should be fine.

Eisei24 (approved):
lgtm

PR #2810, opened 2024-01-12
Fix txFee calculation in dencun

alrevuelta:
Unfamiliar with the codebase, but just found out that the transaction reward in dencun fork (seen in devnet-12) is wrong. Most likely because the new tx type `3` is not being handled properly. Its being handled as type `0` or `1` and it should be handled as type `2`.

Example, in slot 317601 of devnet-12:
* 1) The reported tx fee is `5039999999706000` see [here](https://dencun-devnet-12.beaconcha.in/slot/317601).
* 2) But the tx fee is in reality `4284000000000000`, which is the reward that is really sent to the [fee recipient](https://dencun-devnet-12.beaconcha.in/address/0xf97e180c050e5ab072211ad2c213eb5aee4df134#blocks)

What is reported i)
imagen

What in reality the fee recipient gets ii)
imagen

Unfinished PR but this should fix it.

### DIFF
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -914,15 +914,18 @@ func CalculateTxFeesFromBlock(block *types.Eth1Block) *big.Int {
 func CalculateTxFeeFromTransaction(tx *types.Eth1Transaction, blockBaseFee *big.Int) *big.Int {
 	// calculate tx fee depending on tx type
 	txFee := new(big.Int).SetUint64(tx.GasUsed)
-	if tx.Type == uint32(2) {
+	switch tx.Type {
+	case 0, 1:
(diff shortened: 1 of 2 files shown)


alrevuelta:
Same happening in Goerli after the fork
Slot 7415006 shows 0.00784...
But the fee sent to the recipient is 0.007807...

guybrush:
Thank you!

This will land in #2817 (commit will be preserved), also adding some basic stats to the charts_data.

guybrush:
> Same happening in Goerli after the fork
Slot 7415006 shows 0.00784...
But the fee sent to the recipient is 0.007807...

You are right, the numbers are wrong for slot 7415006, it should be `0.01655 ETH` instead of `0.03319 ETH`

This should be fixed now, thank you again.

old:
image

new:
image

PR #2825, opened 2024-01-24
(BIDS-2879) remove total withrdawals from account-page for now

guybrush:


guybrush:
remove total withdrawals field for now until we figured out a better way for total withdrawals

alanaladd21 (approved):

recy21 (approved):

PR #2828, opened 2024-01-30
(BIDS-2981) fix activation estimation for eip-7514

guybrush:


Elenore23 (approved):

services/stats.go
```diff
@@ -199,6 +199,21 @@ func eth1UniqueValidatorsCount() (*uint64, error) {
 }
 
 // GetValidatorChurnLimit returns the rate at which validators can enter or leave the system
```
D13ce, line 201: This tool tip contains the wrong function name (`GetValidatorChurnLimit`).
Also, is it correct? Based on the function name I would assume that this is only about the enter rate and not about the leave rate.

D13ce (changes requested):
The build for this PR fails. You have added a new function `getValidatorActivationChurnLimit` that is not used anywhere. Removing it would solve the build error but it would also render the whole PR almost meaningless as it is the only location where the new config value `MaxPerEpochActivationChurnLimit` is used.

D13ce (approved):
utACK

PR #2830, opened 2024-02-06
(BIDS-2535) added default methods based on contract state

remoterami:


db/bigtable_eth1.go
```diff
@@ -4593,24 +4579,34 @@ func (bigtable *Bigtable) GetSignature(hex string, st types.SignatureType) (*str
...
+	case types.CONTRACT_DESTRUCTION:
+		return "Destruction"
+	case types.CONTRACT_PRESENT:
+		if len(id) == 4 {
 			cacheKey := fmt.Sprintf("M:H2L:%s", method)
 			if _, err := cache.TieredCache.GetWithLocalTimeout(cacheKey, time.Hour, &method); err != nil {
-				sig, err := bigtable.GetSignature(method, types.MethodSignature)
-				if err == nil {
+				if sig, err := bigtable.GetSignature(method, types.MethodSignature); err == nil {
+					cache.TieredCache.Set(cacheKey, method, time.Hour)
```
Eisei24, line 4601: **issue:** This sets the wrong method in the cache.Compare it with the handling before if `sig != nil`.

Eisei24 (changes requested):
Please have a look at my comment.

db/bigtable_eth1.go
remoterami: Good find, fixed

db/bigtable_eth1.go
```diff
@@ -4593,24 +4579,34 @@ func (bigtable *Bigtable) GetSignature(hex string, st types.SignatureType) (*str
...
+	}
+	method := fmt.Sprintf("0x%x", id)
+
+	switch interaction {
+	case types.CONTRACT_NONE:
+		return "Transfer"
+	case types.CONTRACT_CREATION:
+		return "Constructor"
+	case types.CONTRACT_DESTRUCTION:
+		return "Destruction"
```
Eisei24, line 4595: **nitpick:** **Constructor** and **Destruction** are not the same type of word.
If we want to leave it like that it is also fine.

Eisei24 (approved):
utACK, tested a few cases but there are many possibilities.

db/bigtable_eth1.go
remoterami: I see what you mean, but I think this is okay because developers can decide how a smart contract is constructed, while that's not possible for destruction. I chose that wording to highlight that difference, but it's not a very strong opinion either.

PR #2885, opened 2024-06-04
Remove notice about delayed broadcast of BLS-to-execution messages

m3diumrare:


m3diumrare:
Hi @guybrush - is there a process for external contributors?

PR #2887, opened 2024-06-06
(NOBIDS) improve storage of eb in bigtable by storing it as a 2 byte …

guybrush:
…value (backport v2 change ddbda9bc)


invis-bitfly (approved):
lgtm

PR #2965, opened 2024-10-08
feat(BEDS-536): add `re-index-blocks` cmd to fix internal transfers status

Tangui-Bitfly:
* save when a trace revert an consider that all its child are reverted
* update transaction view ([ref](https://beaconcha.in/tx/f7d385f000250c073dfef9a36327c5d30a4c77a0c50588ce3eded29f6829a4cd#overview))

image

* update traces view ([ref](https://beaconcha.in/tx/f7d385f000250c073dfef9a36327c5d30a4c77a0c50588ce3eded29f6829a4cd#overview))

image

* introduce an unit test, this required few refactoring
  * TieredCacher interface to allow mock
  * RedisClient interface to allow mock
  * SQLReaderDb interface to allow mock
  * InitBigtableWithCache to inject mock dependency
  * type extraction inside `types/config.go` to simplify test writing


types/eth1.proto
```diff
@@ -63,7 +69,7 @@ message Eth1Transaction {
     uint64 commulative_gas_used = 18;
     uint64 gas_used = 19;
     bytes logs_bloom = 20;
-    uint64 status = 21;
+    StatusType status = 21;
```
Monika-Bitfly, line 72: We need to make sure that this doesn't cause any conflict when using proto buffer

peterbitfly: I suggest adding this as new field instead of replacing the existing one, otherwise this might cause issues with data that is currently in the db?

guybrush (approved):

PR #2967, opened 2024-10-18
feat(BEDS-536): implement misc `fix-internal-txs-from-node` cmd with batch processing from Erigon node

Monika-Bitfly:
Changes:
- introduce `fix-internal-txs-from-node` command for reindexing internal txs within a specified block range
- utilise batch calls to Erigon node for improved speed during data retrieval:
   - execute three RPC calls for each block:
      - `eth_getBlockByNumber`
      - `eth_getBlockReceipts`
      - `trace_block`


rpc/erigon.go
```diff
@@ -768,3 +770,339 @@ type Eth1InternalTransactionWithPosition struct {
...
+			}
+		}
+	} else {
+		return fmt.Errorf("expected an array for uncles, but got: %s", tempData.Uncles)
+	}
+
+	return nil
+}
+
+func (client *ErigonClient) GetBlocksByBatch(blocksChan chan *types.Eth1Block, erigonClient *ErigonClient) ([]*types.Eth1Block, *types.GetBlockTimings, error) {
```
Tangui-Bitfly, line 847: wrong signature, no need to receive `erigonClient *ErigonClient`

rpc/erigon.go
```diff
@@ -768,3 +770,339 @@ type Eth1InternalTransactionWithPosition struct {
...
+
+		ethBlock.Transactions = append(ethBlock.Transactions, pbTx)
+
+	}
+	timings.Headers = time.Since(startTime)
+
+	return ethBlock, timings
+}
+
+func (client *ErigonClient) processReceiptsAndTraces(ethBlock *types.Eth1Block, receipts []geth_types.Receipt, traces []ParityTraceResult, timings *types.GetBlockTimings) (*types.Eth1Block, *types.GetBlockTimings) {
```
Tangui-Bitfly, line 1036: either
* take `ethBlock *types.Eth1Block` and don't return `*types.Eth1Block` it already a pointer so you can just directly modify it without returning it
* don't take `ethBlock *types.Eth1Block` and return `[]*Eth1InternalTransactionWithPosition` and `[]*types.Eth1Transaction`. But this will require more refactoring

rpc/interfaces.go
```diff
@@ -25,6 +25,7 @@ type Client interface {
 
 type Eth1Client interface {
 	GetBlock(number uint64) (*types.Eth1Block, *types.GetBlockTimings, error)
+	GetBlocksByBatch(blocksChan chan *types.Eth1Block) error
```
Tangui-Bitfly, line 28: this interface is not respected by `ErigonClient`, you can delete that change

rpc/erigon.go
```diff
@@ -768,3 +770,339 @@ type Eth1InternalTransactionWithPosition struct {
...
+		if err != nil {
+			fmt.Printf("\n errror while unmarshalling block results %s\n", err)
+		}
+
+		blockDetail, timings := client.processBlockResult(blockResponse, timings)
+		blockDetail, timings = client.processReceiptsAndTraces(blockDetail, *receiptsResult, *tracesResults, timings)
+		blockDetails = append(blockDetails, blockDetail)
+	}
+
+	return blockDetails, timings, nil
```
Tangui-Bitfly, line 911: there is something wrong with `timings`. It's always overridden by the last block you proceed.
Moreover we need to think about it in the context of that function. I think it use to record the time taken to retrieve a block (through 3 requests). In your case it's now one request for N blocks.
I would say just remove it, and if it's deemed important create a new timing type for batch request.

rpc/erigon.go
```diff
@@ -768,3 +770,339 @@ type Eth1InternalTransactionWithPosition struct {
...
+	Bloom         string                    `json:"logsBloom"`
+	Uncles        []*geth_types.Block       `json:"uncles"`
+	Transactions  []*geth_types.Transaction `json:"transactions"`
+	Withdrawals   []*geth_types.Withdrawal  `json:"withdrawals"`
+	BlobGasUsed   *string                   `json:"blobGasUsed"`
+	ExcessBlobGas *string                   `json:"excessBlobGas"`
+	BaseFee       string                    `json:"baseFee"`
+}
+
+func (b *BlockResponse) UnmarshalJSON(data []byte) error {
```
Tangui-Bitfly, line 798: Have a look at https://github.com/ethereum/go-ethereum/blob/master/ethclient/ethclient.go#L128
you could reuse some of the code done there for uncles check and header parsing

Tangui-Bitfly (changes requested):

rpc/erigon.go
Monika-Bitfly: The Erigon client was always nil during my tests, which is why I passed it here

db/bigtable_eth1.go
```diff
@@ -4910,3 +4909,152 @@ func (bigtable *Bigtable) GetGasNowHistory(ts, pastTs time.Time) ([]types.GasNow
...
+	for i := start; i <= end; i += batchSize {
+		firstBlock := i
+		lastBlock := firstBlock + batchSize - 1
+		if lastBlock > end {
+			lastBlock = end
+		}
+
+		// create batch request for each block height
+		g.Go(func() error {
+			blocksChan := make(chan *types.Eth1Block, batchSize)
```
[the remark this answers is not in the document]
Monika-Bitfly: The plan was to reparse the blocks already stored in the database. I will update the logic to parse the blocks provided in the command instead

Tangui-Bitfly (approved):

PR #2972, opened 2024-10-28
(BEDS-536) Use raw db for resync

Tangui-Bitfly:


cmd/store/main.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/gobitfly/eth2-beaconchain-explorer/types"
+	"github.com/gobitfly/eth2-beaconchain-explorer/utils"
+)
+
+func main() {
+	configPath := flag.String("config", "config/default.config.yml", "Path to the config file")
+	flag.Parse()
+
+	cfg := &types.Config{}
+	err := utils.ReadConfig(cfg, *configPath)
```
Monika-Bitfly, line 21:
```suggestion
	err := utils.ReadConfig(cfg, *configPath)
	if err != nil {
		panic(err)
	}
```

cmd/misc/main.go
```diff
@@ -1628,68 +1628,134 @@ func reIndexBlocks(start uint64, end uint64, bt *db.Bigtable, client *rpc.Erigon
...
+			utils.LogError(err, "error importing ens from events", 0)
+			return
 		}
-		to := from + targetCount - 1
+	}
 
-		errFields["from"] = from
-		errFields["to"] = to
-		errFields["targetCount"] = targetCount
+	cache := freecache.NewCache(100 * 1024 * 1024) // 100 MB limit
```
Monika-Bitfly, line 1643:
```suggestion
	cache := freecache.NewCache(100 * 1024 * 1024) // 100 MB limit
```
We're using this value in multiple places, can we define it as const and place it at the top of the file pls?

cmd/misc/main.go
```diff
@@ -1628,68 +1628,134 @@ func reIndexBlocks(start uint64, end uint64, bt *db.Bigtable, client *rpc.Erigon
...
 
-	transforms := make([]func(blk *types.Eth1Block, cache *freecache.Cache) (*types.BulkMutations, *types.BulkMutations, error), 0)
+	readGroup := errgroup.Group{}
+	readGroup.SetLimit(int(concurrency))
+
+	writeGroup := errgroup.Group{}
+	writeGroup.SetLimit(int(concurrency*concurrency) + 1)
+
+	cache := freecache.NewCache(100 * 1024 * 1024) // 100 MB limit
+	quit := make(chan any)
```
Monika-Bitfly, line 1694:
```suggestion
	quit := make(chan struct{})
```

cmd/misc/main.go
```diff
@@ -1628,68 +1628,134 @@ func reIndexBlocks(start uint64, end uint64, bt *db.Bigtable, client *rpc.Erigon
...
+
+	sink := make(chan *types.Eth1Block)
+	writeGroup.Go(func() error {
+		for {
+			select {
+			case block := <-sink:
```
Monika-Bitfly, line 1700:
```suggestion
			case block, ok := <-sink:
			if !ok {
				return nil 
			}
```

I think we can add check here to handle the case where the channel might be closed

db2/cache.go
```diff
@@ -0,0 +1,118 @@
+package db2
+
+import (
+	"encoding/json"
+	"sync"
+	"time"
+)
+
+// var ttl = 20_000 * time.Millisecond
```
Monika-Bitfly, line 9:
```suggestion
```
we can remove this if it's unused

db2/cache.go
```diff
@@ -0,0 +1,118 @@
...
+var oneBlockTTL = 1 * time.Second
```
Monika-Bitfly, line 10: We can define oneBlockTTL and blocksTTL as const values since they are not modified anywhere in the code

db2/cache.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (c *CachedRawStore) cacheBlock(block *FullBlockRawData, ttl time.Duration) {
+	key := blockKey(block.ChainID, block.BlockNumber)
+	c.cache.Store(key, block)
+
+	var mini MinimalBlock
+	if len(block.Uncles) != 0 {
+		// retrieve the block hash for caching but only if the block has uncle(s)
+		_ = json.Unmarshal(block.Block, &mini)
```
Monika-Bitfly, line 77:
```suggestion
		err := json.Unmarshal(block.Block, &mini)
		if err != nil {
		   logrus.Errorf("failed to unmarshal block: %v", err)
		}
```

db2/client.go
```diff
@@ -0,0 +1,253 @@
...
+	"math/big"
+	"net/http"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+
+	"github.com/gobitfly/eth2-beaconchain-explorer/db2/store"
+)
+
+var ErrNotFoundInCache = fmt.Errorf("cannot find hash in cache")
+var ErrMethodNotSupported = fmt.Errorf("methode not supported")
```
Monika-Bitfly, line 19:
```suggestion
var ErrMethodNotSupported = fmt.Errorf("method not supported")
```

db2/client.go
```diff
@@ -0,0 +1,253 @@
...
+	var resps []*jsonrpcMessage
+	for _, message := range messages {
+		resp, err := r.handle(request.Context(), message)
+		if err != nil {
+			return nil, err
+		}
+		resps = append(resps, resp)
+	}
+
+	respBody, _ := makeBody(isSingle, resps)
```
Monika-Bitfly, line 96:
```suggestion
	respBody, err := makeBody(len(resps) == 1, resps)
	if err != nil {
	   return nil, err
	}
```

db2/client.go
```diff
@@ -0,0 +1,253 @@
...
+	return &http.Response{
+		Body:       respBody,
+		StatusCode: http.StatusOK,
+	}, nil
+}
+
+func (r *BigTableEthRaw) handle(ctx context.Context, message *jsonrpcMessage) (*jsonrpcMessage, error) {
+	var args []interface{}
+	// ignore error
+	_ = json.Unmarshal(message.Params, &args)
```
Monika-Bitfly, line 106:
```suggestion
	err := json.Unmarshal(message.Params, &args)
	if err != nil {
	   return nil, err
	}
```

db2/client.go
```diff
@@ -0,0 +1,253 @@
...
+
+func (r WithFallback) RoundTrip(request *http.Request) (*http.Response, error) {
+	resp, err := r.roundTripper.RoundTrip(request)
+	if err == nil {
+		// no fallback needed
+		return resp, nil
+	}
+
+	var e1 *json.SyntaxError
+	if !errors.As(err, &e1) &&
```
Monika-Bitfly, line 47: Should this condition use || instead of &&? Coz the current logic seems to require that all error checks must be true for the error to be handled

db2/store/bigtable.go
```diff
@@ -0,0 +1,338 @@
...
+	"fmt"
+	"slices"
+	"strings"
+	"time"
+
+	"cloud.google.com/go/bigtable"
+	"golang.org/x/exp/maps"
+)
+
+var ErrNotFound = fmt.Errorf("not found")
```
Monika-Bitfly, line 14: we can define this as const value

db2/store/bigtable.go
```diff
@@ -0,0 +1,338 @@
...
+func NewBigTable(project, instance string, tablesAndFamilies map[string][]string) (*BigTableStore, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	// Create an admin client to manage Bigtable tables
+	adminClient, err := bigtable.NewAdminClient(ctx, project, instance)
+	if err != nil {
+		return nil, fmt.Errorf("could not create admin client: %v", err)
+	}
+	// Ensure the admin client is closed after the operations are done
```
Monika-Bitfly, line 88:
```suggestion
```

db2/store/remote.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, err
+	}
+	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("%s%s", r.url, routeGetRow), bytes.NewReader(b))
+	if err != nil {
+		return nil, err
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
```
Monika-Bitfly, line 112:
```suggestion
	}
	
	defer resp.Body.Close()
	
```

db2/store/remote.go
```diff
@@ -0,0 +1,169 @@
...
+	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("%s%s", r.url, routeGetRowsRange), bytes.NewReader(b))
+	if err != nil {
+		return nil, err
+	}
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
```
Monika-Bitfly, line 149:
```suggestion
	}
	
	defer resp.Body.Close()
	
```

db2/tables.go
```diff
@@ -0,0 +1,23 @@
+package db2
+
+const BlocRawTable = "blocks-raw"
```
Monika-Bitfly, line 3:
```suggestion
const BlockRawTable = "blocks-raw"
```

rpc/erigon.go
```diff
@@ -47,17 +53,37 @@ func NewErigonClient(endpoint string) (*ErigonClient, error) {
...
 	}
 
-	rpcClient, err := geth_rpc.Dial(client.endpoint)
-	if err != nil {
-		return nil, fmt.Errorf("error dialing rpc node: %w", err)
+	var opts []geth_rpc.ClientOption
+	if utils.Config != nil {
+		if utils.Config.RawBigtable.Bigtable.Project != "" && utils.Config.RawBigtable.Bigtable.Instance != "" {
+			if utils.Config.RawBigtable.Bigtable.Emulator {
+				_ = os.Setenv("BIGTABLE_EMULATOR_HOST", fmt.Sprintf("%s:%d", utils.Config.RawBigtable.Bigtable.EmulatorHost, utils.Config.RawBigtable.Bigtable.EmulatorPort))
```
Monika-Bitfly, line 60:
```suggestion
				err := os.Setenv("BIGTABLE_EMULATOR_HOST", fmt.Sprintf("%s:%d", utils.Config.RawBigtable.Bigtable.EmulatorHost, utils.Config.RawBigtable.Bigtable.EmulatorPort))
				if err != nil {
				   return nil, fmt.Errorf("error while setting BIGTABLE_EMULATOR_HOST env: %w", err)
				}
```

rpc/erigon.go
```diff
@@ -708,9 +758,9 @@ func (client *ErigonClient) getTraceParity(block *geth_types.Block) ([]*Eth1Inte
 		if trace.TransactionHash == "" {
 			continue
 		}
-		if trace.TransactionPosition >= len(block.Transactions()) {
-			return nil, fmt.Errorf("error transaction position %v out of range", trace.TransactionPosition)
-		}
+		// if trace.TransactionPosition >= txsLen {
```
Monika-Bitfly, line 761: we can remove this part

db2/cache.go
```diff
@@ -0,0 +1,118 @@
...
+		// once read ensure to delete it from the cache
+		go c.unCacheBlockAfter(key, "", oneBlockTTL)
+		return v.(*FullBlockRawData), nil
+	}
+	// TODO make warning not found in cache
+	block, err := c.db.ReadBlockByNumber(chainID, number)
+	if block != nil {
+		c.cacheBlock(block, oneBlockTTL)
+	}
+	return block, err
```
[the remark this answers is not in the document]
Tangui-Bitfly: no here we want to keep thing like that. We don't want to handle the error here, just cache the block if it's not nil and let the caller handle the error.

db2/cache.go
Tangui-Bitfly: Honestly this error will never happen, and if it do something VERY wrong is saved in our db.
If you really want to handle it, I would prefer to use a panic to show to the reader that this is not supposed to happen EVER.

db2/client.go
Tangui-Bitfly: We check if the error is a known error from our service. If it is, we can use fallback, otherwise we want to report the unexpected error.

cmd/misc/main.go
Tangui-Bitfly: I rather not since I don't know what it does.

db2/client.go
Tangui-Bitfly: `len(resps) == 1` is different than isSingle.
We can receive a batch with only one request, the response need to be wrapped in an array. So len(resps) doesn't give enough informations about the caller context.

db2/store/bigtable.go
Monika-Bitfly: Oh yes you're right, my bad pls ignore this one 👍

Monika-Bitfly (approved):

PR #2998, opened 2025-01-13
add: write end date of premium subscriptions to db

remoterami:


handlers/stripe.go
```diff
@@ -560,7 +560,7 @@ func StripeWebhook(w http.ResponseWriter, r *http.Request) {
 				http.Error(w, "error updating stripe mobile subscription, no users_app_subs id  found for subscription id, customer: "+invoice.Customer.ID, http.StatusInternalServerError)
 				return
 			}
-			err = db.UpdateUserSubscription(tx, appSubID, true, 0, "")
+			err = db.UpdateUserSubscription(tx, appSubID, true, invoice.Lines.Data[0].Period.End, "")
```
manuelsc, line 563: **minor**: check array access if there is an element at position 0

manuelsc (changes requested):

handlers/stripe.go
remoterami: Should be there already, see https://github.com/gobitfly/eth2-beaconchain-explorer/blob/190683467b41b74ffb61bb50fd579022d2c82ad5/handlers/stripe.go#L532-L535

manuelsc (approved):

PR #2307, opened 2023-06-13
(NOBIDS) fix migrations for cuncurrently added indexes

guybrush:
### <samp>🤖 Generated by Copilot at 85085de</samp>

Added `-- +goose NO TRANSACTION` directive to `db/migrations/20230612084855_add_missing_indices.sql` to enable concurrent index creation for various tables and columns. This change aims to optimize the database queries and avoid errors.

PR #3002, opened 2025-01-15
Add notransaction fix

GlacierWalrus:
Disable transaction around `CREATE INDEX CONCURRENTLY` statements

See https://github.com/pressly/goose/issues/292  and #2307


GlacierWalrus:
@peterbitfly would you be able to review/tag somebody else in please?

GlacierWalrus:
No interest it seems

PR #3019, opened 2025-02-12
Gasnow fixes

manuelsc:
- Fix gasnow ordering
- Improved algo
- Display in 1 decimal precision on subpage

### DIFF
--- a/services/services.go
+++ b/services/services.go
@@ -1296,111 +1296,185 @@ func getGasNowData() (*types.GasNowPageData, error) {
 	gpoData.Code = 200
 	gpoData.Data.Timestamp = time.Now().UnixNano() / 1e6
 
+	// Connect to the ETH node.
 	client, err := geth_rpc.Dial(utils.Config.Eth1GethEndpoint)
 	if err != nil {
 		return nil, err
 	}
+	defer client.Close()
+
+	// Get the pending block (which will soon be proposed).
 	var raw json.RawMessage
 	err = client.Call(&raw, "eth_getBlockByNumber", "pending", true)
 	if err != nil {
-		return nil, fmt.Errorf("error retrieving pending block data: %.1000s", err) // limit error message to 1000 characters
+		return nil, fmt.Errorf("error retrieving pending block data: %w", err)
 	}
 
-	// var res map[string]interface{}
-	// err = json.Unmarshal(raw, &res)
-	// if err != nil {
-	// 	return nil, err
-	// }
-
+	// Unmarshal into both a header (for base fee) and a block body (for tx list).
 	var header *geth_types.Header
 	var body rpcBlock
-
-	err = json.Unmarshal(raw, &header)
-	if err != nil {
+	if err = json.Unmarshal(raw, &header); err != nil {
 		return nil, err
 	}
-	err = json.Unmarshal(raw, &body)
-	if err != nil {
+	if err = json.Unmarshal(raw, &body); err != nil {
 		return nil, err
 	}
 	txs := body.Transactions
-
-	sort.Slice(txs, func(i, j int) bool {
-		return txs[i].tx.GasPrice().Cmp(txs[j].tx.GasPrice()) > 0
-	})
-	if len(txs) > 1 {
-		medianGasPrice := txs[len(txs)/2].tx.GasPrice()
-		tailGasPrice := txs[len(txs)-1].tx.GasPrice()
-
-		gpoData.Data.Rapid = medianGasPrice
-		gpoData.Data.Fast = tailGasPrice
-	} else {
-		gpoData.Data.Rapid = new(big.Int)
-		gpoData.Data.Fast = new(big.Int)
+	// Make sure the header contains a base fee.
+	if header.BaseFee == nil {
+		return nil, fmt.Errorf("pending block header has nil BaseFee")
 	}
+	baseFee := header.BaseFee
 
-	err = client.Call(&raw, "txpool_content")
-	if err != nil {
-		return nil, fmt.Errorf("error getting raw json data from txpool_content: %w", err)
-	}
-
-	txPoolContent := &TxPoolContent{}
-	err = json.Unmarshal(raw, txPoolContent)
-	if err != nil {
-		return nil, fmt.Errorf("unmarshal txpoolcontent json error: %w", err)
-	}
-
-	pendingTxs := make([]*TxPoolContentTransaction, 0, len(txPoolContent.Pending))
-
-	for _, account := range txPoolContent.Pending {
-		lowestNonce := 9223372036854775807
-		for n := range account {
-			if n < int(lowestNonce) {
-				lowestNonce = n
-			}
+	// Handle edgecase when there's too few txs in the pending block
+	// Use latest instead
+	if len(txs) < 40 { // 40 chosen as roughly 1/5 of current average tx in a block
+		var rawLatest json.RawMessage
+		err = client.Call(&rawLatest, "eth_getBlockByNumber", "latest", true)
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving latest block data: %w", err)
+		}
+		if err = json.Unmarshal(rawLatest, &body); err != nil {
+			return nil, err
 		}
 
-		pendingTxs = append(pendingTxs, account[lowestNonce])
+		// update tx to latest but keep basefee of pending block
+		// as this is still the target block we suggest gas prices for
+		txs = body.Transactions
 	}
-	sort.Slice(pendingTxs, func(i, j int) bool {
-		return pendingTxs[i].GetGasPrice().Cmp(pendingTxs[j].GetGasPrice()) > 0
+
+	// How full a block is [0, 1], protocol targets 50% usage
+	gasUsage := float64(header.GasUsed) / float64(header.GasLimit)
+
+	// -------------------------
+	// (1) Build our “tip” samples.
+	// -------------------------
+	// For transactions in the pending block, compute effective tip = GasPrice - baseFee.
+	pendingTips := make([]*big.Int, 0, len(txs))
+	for _, tx := range txs {
+		tip := tx.tx.GasTipCap()
+		if tip.Sign() < 0 {
+			tip = big.NewInt(0)
+		}
+		pendingTips = append(pendingTips, tip)
+	}
+	// Sort in ascending order (so that e.g. the 90th percentile is near the high end).
+	sort.Slice(pendingTips, func(i, j int) bool {
+		return pendingTips[i].Cmp(pendingTips[j]) < 0
 	})
 
-	standardIndex := int(math.Max(float64(2*len(txs)), 500))
-
-	slowIndex := int(math.Max(float64(5*len(txs)), 1000))
-	if standardIndex < len(pendingTxs) {
-		gpoData.Data.Standard = pendingTxs[standardIndex].GetGasPrice()
-	} else {
-		gpoData.Data.Standard = header.BaseFee
+	// -------------------------
+	// (2) Compute tip percentiles.
+	// Define a helper to get a percentile from a sorted slice.
+	getPercentile := func(tips []*big.Int, pct float64) *big.Int {
+		if len(tips) == 0 {
+			return big.NewInt(0)
+		}
+		// Calculate an index between 0 and len(tips)-1.
+		index := int(math.Ceil((pct/100.0)*float64(len(tips)))) - 1
+		if index < 0 {
+			index = 0
+		} else if index >= len(tips) {
+			index = len(tips) - 1
+		}
+		return tips[index]
 	}
 
-	if gpoData.Data.Standard.Cmp(header.BaseFee) < 0 {
-		gpoData.Data.Standard = header.BaseFee
+	// Dictates the aggressiveness to scale the percentiles for depending on the block usage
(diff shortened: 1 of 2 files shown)


services/services.go
```diff
@@ -1292,116 +1292,201 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
-	sort.Slice(pendingTxs, func(i, j int) bool {
-		return pendingTxs[i].GetGasPrice().Cmp(pendingTxs[j].GetGasPrice()) > 0
-	})
 
-	standardIndex := int(math.Max(float64(2*len(txs)), 500))
+	gpoData := suggestGasPrices(header.GasUsed, header.GasLimit, baseFee, pendingTips)
+
+	// not available in unit test mode
+	if db.BigtableClient != nil {
+		// Log or store historical data.
```
mccreedy-bitfly, line 1358: Really not a fan of having such checks in our business logic, but since this is in a soon-to-be deprecated package....

services/services.go
```diff
@@ -1292,116 +1292,201 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
-
-		gpoData.Data.Rapid = medianGasPrice
-		gpoData.Data.Fast = tailGasPrice
-	} else {
-		gpoData.Data.Rapid = new(big.Int)
-		gpoData.Data.Fast = new(big.Int)
+	// Make sure the header contains a base fee.
+	if header.BaseFee == nil {
+		return nil, fmt.Errorf("pending block header has nil BaseFee")
 	}
```
mccreedy-bitfly, line 1322: When can this happen?

services/services.go
```diff
@@ -1292,116 +1292,201 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
+	baseFee := header.BaseFee
 
-	err = client.Call(&raw, "txpool_content")
-	if err != nil {
-		return nil, fmt.Errorf("error getting raw json data from txpool_content: %w", err)
-	}
+	// Handle edgecase when there's too few txs in the pending block
+	// Use latest instead
+	if len(txs) < 40 { // 40 chosen as roughly 1/5 of current average tx in a block
```
mccreedy-bitfly, line 1327: Why do we do this though? Why did we choose 40? I understand that it's 1/5 the current average transaction, but what's the significance of this? And why is 1/5 the right ratio?

services/services.go
```diff
@@ -1292,116 +1292,201 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
+		}
 
-	txPoolContent := &TxPoolContent{}
-	err = json.Unmarshal(raw, txPoolContent)
-	if err != nil {
-		return nil, fmt.Errorf("unmarshal txpoolcontent json error: %w", err)
+		// update tx to latest but keep basefee of pending block
+		// as this is still the target block we suggest gas prices for
+		txs = body.Transactions
 	}
```
mccreedy-bitfly, line 1340: Aren't we writing the same thing twice in this larger function? I.e. on line 1304 you also grab the block and pull out the transactions... why can't this be a single function which takes as a parameter something that indicates whether it should load in the "latest" block or the "pending" block? Then, your logic would be:

```
block := loadBlock("pending")
numTransactions = countTransactions(block) // doesnt have to use this structure, just an example
if (numTransactions < 40) {
    block = loadBlock("latest")
}
```

Done this way, we have nicely unit testable functions and the logic here in this higher level function is simplified.

services/services.go
```diff
@@ -1292,116 +1292,201 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
+func suggestGasPrices(gasUsed uint64, gasLimit uint64, baseFee *big.Int, pendingTips []*big.Int) *types.GasNowPageData {
+	gpoData := &types.GasNowPageData{}
+	gpoData.Code = 200
+	gpoData.Data.Timestamp = time.Now().UnixNano() / 1e6
 
-	if gpoData.Data.Slow.Cmp(header.BaseFee) < 0 {
-		gpoData.Data.Slow = header.BaseFee
-	}
+	// How full a block is [0, 1], protocol targets 50% usage
+	gasUsage := float64(gasUsed) / float64(gasLimit)
```
mccreedy-bitfly, line 1379: This isn't used until line 1404 so why is it defined way up here?

services/services.go
```diff
@@ -1292,116 +1292,201 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
+	// gpoData.Data.V2.Standard = types.GasNowV2Pair{
+	// 	Gas:     new(big.Int).Add(predictedNormalBaseFee, normalTip),
+	// 	Tip:     normalTip,
+	// 	BaseFee: predictedNormalBaseFee,
+	// }
+	// gpoData.Data.V2.Slow = types.GasNowV2Pair{
+	// 	Gas:     new(big.Int).Add(predictedSlowBaseFee, slowTip),
+	// 	Tip:     slowTip,
+	// 	BaseFee: predictedSlowBaseFee,
+	// }
```
mccreedy-bitfly, line 1487: Don't include suggestions like this in the code. Instead, you can create a BEDS for it with this information, put it in the backlog, and then link the BEDS here. The reason is because people can't comment on the suggested improvements here, and you also have to know to look here to find the suggestions. Putting it in BEDS and linking here gets us all of that.

services/services.go
```diff
@@ -1292,116 +1292,201 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
+	gasUsage := float64(gasUsed) / float64(gasLimit)
 
-	err = db.BigtableClient.SaveGasNowHistory(gpoData.Data.Slow, gpoData.Data.Standard, gpoData.Data.Fast, gpoData.Data.Rapid)
-	if err != nil {
-		logrus.WithError(err).Error("error updating gas now history")
-	}
+	// Sort in ascending order (so that e.g. the 90th percentile is near the high end).
+	sort.Slice(pendingTips, func(i, j int) bool {
+		return pendingTips[i].Cmp(pendingTips[j]) < 0
+	})
```
mccreedy-bitfly, line 1384: Why does order matter? Why do we want the 90th percentile to be near the high end?

services/services.go
```diff
@@ -1292,116 +1292,201 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
+	// rapid ≥ fast ≥ standard ≥ slow.
+	if rapidSuggestion.Cmp(fastSuggestion) < 0 {
+		fastSuggestion = new(big.Int).Set(rapidSuggestion)
+	}
+	if fastSuggestion.Cmp(normalSuggestion) < 0 {
+		normalSuggestion = new(big.Int).Set(fastSuggestion)
+	}
+	if normalSuggestion.Cmp(slowSuggestion) < 0 {
+		slowSuggestion = new(big.Int).Set(normalSuggestion)
+	}
```
mccreedy-bitfly, line 1438: This shouldn't be possible if the percentile function works correctly. Let's just make sure that works and then we can farther simplify the code by removing this

services/services_test.go
```diff
@@ -0,0 +1,232 @@
...
+// 	}
+// 	if data == nil {
+// 		t.Errorf("Error: data is nil")
+// 	}
+
+// 	fmt.Printf("Rapid: %.1f gwei\n", new(big.Float).Quo(new(big.Float).SetInt(data.Data.Rapid), big.NewFloat(1e9)))
+// 	fmt.Printf("Fast: %.1f gwei\n", new(big.Float).Quo(new(big.Float).SetInt(data.Data.Fast), big.NewFloat(1e9)))
+// 	fmt.Printf("Standard: %.1f gwei\n", new(big.Float).Quo(new(big.Float).SetInt(data.Data.Standard), big.NewFloat(1e9)))
+// 	fmt.Printf("Slow: %.1f gwei\n", new(big.Float).Quo(new(big.Float).SetInt(data.Data.Slow), big.NewFloat(1e9)))
+// }
```
mccreedy-bitfly, line 232: ?

mccreedy-bitfly (commented):
From the commit description: 

> Fix gasnow ordering
> Improved algo

Please provide more information. What was wrong with the gasnow ordering, and how does it do not that it's "fixed"?
What was "improved" about the algo?

services/services.go
```diff
@@ -1292,116 +1292,201 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
+		}
+		// Calculate an index between 0 and len(tips)-1.
+		index := int(math.Ceil((pct/100.0)*float64(len(tips)))) - 1
+		if index < 0 {
+			index = 0
+		} else if index >= len(tips) {
+			index = len(tips) - 1
+		}
+		return tips[index]
+	}
```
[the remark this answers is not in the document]
manuelsc: This usually is a great suggestion, but changing dependencies in v1 felt too intrusive for this rather simple PR. I tried using some of the dependencies  we use for v2 unit tests, which lead to the update of some common dependencies between them and ultimately breaking the build. So I rather not touch any dependencies to keep the scope down and limit the intrusiveness of this PR on v1. 

Will add it as a comment though for when we migrate this to v2.

services/services.go
manuelsc: Yeah I know :/ Unfortunately we don't have much to work with here in v1

services/services.go
manuelsc: Not sure. It's a struct from the Geth lib which we unmarshal from a node response - it is a pointer so it can be nil. Better to check than to be sorry :)

services/services.go
manuelsc: Imagine low activity on chain. A new block propagates, clearing the mempool. Obviously we can not predict gas for the next block with an empty mempool. So we fall back on the current block that just happened to handle this edge case.
Realistically to provide an actual useful estimate we need more than just "zero plus one" transactions. But then how many would allow us to make a useful prediction? We don't want to set the percentiles for rapid such as it targets the first one or two tx in a block as these could easily be outliers, so 80% of 40 rather solves this. At the same time you don't want to set the target too high as then probably more time has passed since the "latest" block has been proposed and the current state of mempool is a better candidate again. 

So these were the thoughts that went into the number. I could see us settle for a bit lower number as well but probably not much higher. 40 felt right to keep that balance but I agree that it is to some degree magical so feel free to debate me on this - or if you have a better way to approach this altogether.

services/services_test.go
manuelsc: Can be used to test locally against a real node with debug output. I think it's nice to have this be part of it but not sure if this is the right place (or what the right place would look like for this kind of stuff)

services/services_test.go
```diff
@@ -0,0 +1,232 @@
...
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(resp)
+	})
+	return httptest.NewServer(handler)
+}
+
+// --- Test Cases --- //
+
+// TestSuggestGasPrices_SufficientTxs creates 50 transactions with tip values 1..50.
+// With gasUsed=10,000,000, gasLimit=20,000,000 (i.e. gasUsage=0.5) and baseFee=100,
```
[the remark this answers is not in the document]
manuelsc: Right so using 50 transactions gives enough granularity to catch off by one or rounding errors in the percentile calculations - issues that might not show up with a smaller dataset. Plus it more realistically mimics production conditions where blocks have many transactions, ensuring our scaling and percentile logic behaves correctly.

services/services.go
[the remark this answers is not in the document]
mccreedy-bitfly: The library I linked claims that it has no dependencies, so it should be easy to integrate

services/services.go
mccreedy-bitfly: Let's use the perecentile library and then I think we dont have to do this right

services/services_test.go
mccreedy-bitfly: Add it as an integration test in v2

services/services.go
mccreedy-bitfly: Well I think there are two cases we should care about: a non-empty block and an empty block. If the block is empty, assuming something went wrong and use the previous block. If it's not empty, use the current block. I'm not sure it makes sense to arbitrarily choose 40 (or any number, really) of transactions.

40 is a static number and has no bearing on the total number of transactions in the block or the amount of gas paid. You could have 1 really big transaction with a high gas fee that takes up the entire block, for example. That doesn\t make the gas it paid any less relevant when computing the estimated gas fee in the future, because there's no reason why it couldn't happen concurrently across multiple blocks. 

In general, I'm for keeping things as simple as possible, and adjusting as needed. Right now, that means simple using the current block to estimate gas and "skipping" any empty blocks.

services/services.go
```diff
@@ -1291,117 +1292,165 @@ func gasNowUpdater(wg *sync.WaitGroup) {
...
+			logrus.WithError(err).Error("error updating gas now history")
+		}
+
+		// Get fiat conversion data.
+		gpoData.Data.Price = price.GetPrice(utils.Config.Frontend.ElCurrency, "USD")
+		gpoData.Data.Currency = "USD"
 	} else {
-		gpoData.Data.Standard = header.BaseFee
+		logrus.Error("error saving gas now history: bigtable client not initialized")
 	}
```
mccreedy-bitfly, line 1375: Why do we do this upon every request? When we have a lot of requests, we will try to save it when it's already there. When we have nobody requesting it, then we will have gaps in our history.

mccreedy-bitfly (approved):

PR #3031, opened 2025-03-19
Beds 101/api docs

peterbitfly:
Adopts scalar for a streamlined api docs experience.
Improves clarity of api docs


mccreedy-bitfly (approved):
Leaves much to be desired but this is a massive improvement over what we had before

PR #3036, opened 2025-03-26
fix: validator with index 0 not being treatet as a correct validator …

manuelsc:
…index in blocks/produced endpoint


LuccaBitfly (approved):
lgtm

PR #3038, opened 2025-03-28
fix: add support for type 4 tx, fix bigtable auth, remove obsolete hooks

peterbitfly:


utils/format.go
```diff
@@ -542,6 +542,8 @@ func FormatTransactionType(txnType uint8) string {
 		return "2 (EIP-1559)"
 	case 3:
 		return "3 (Blob, EIP-4844)"
+	case 4:
+		return "3 (Set EOA, EIP-7702)"
```
guybrush, line 546: 4

peterbitfly: fixed

PR #3044, opened 2025-04-08
fix(transformer): adds EL data for pectra

Tangui-Bitfly:
Same PR but better on v2 https://github.com/gobitfly/beaconchain/pull/1490
⚠️ migration file is defined on v2 and not included in this PR

### DIFF
--- a/db/bigtable_eth1.go
+++ b/db/bigtable_eth1.go
@@ -2104,17 +2104,65 @@ func (bigtable *Bigtable) TransformWithdrawals(block *types.Eth1Block, cache *fr
 	return bulkData, bulkMetadataUpdates, nil
 }
 
+type BridgeQueueRequest struct {
+	TxHash         []byte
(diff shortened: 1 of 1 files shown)


db/bigtable_eth1.go
```diff
@@ -2135,16 +2183,17 @@ func (bigtable *Bigtable) TransformConsolidationRequests(blk *types.Eth1Block, c
 
 				logger.Infof("consolidation event: %x %x %x", sourceAddress, sourcePubkey, targetPubkey)
```
Monika-Bitfly, line 2185: suggestion: add check for requestIndex and queueRequests length to make sure it doesn't exceed it
````
if requestIndex >= len(queueRequests) {
    return nil, nil, fmt.Errorf("mismatch between requests and events for block %d", blk.Number)
 }

remoterami (approved):
utAck

PR #3059, opened 2025-04-25
Pending Queue Follow Up

manuelsc:


Elenore23 (commented):
Thx

services/services.go
```diff
@@ -910,11 +910,15 @@ func getIndexPageData() (*types.IndexPageData, error) {
...
-		data.ExitingValidatorsBalance = fmt.Sprintf("%.0f %s", float64(queueData.LeavingEthAmount)/1e9, utils.Config.Frontend.ClCurrency)
-		data.EnteringValidators = queueData.EnteringValidatorCount
-		data.ExitingValidators = queueData.LeavingValidatorCount
+		if queueData != nil {
+			total := queueData.EnteringNewValidatorsEthAmount + queueData.EnteringTopUpEthAmount
+			data.EnteringBalance = fmt.Sprintf("%.0f", float64(total)/1e9)
+			data.EnteringValidatorsBalance = fmt.Sprintf("%.0f", float64(queueData.EnteringNewValidatorsEthAmount)/1e9)
+			data.EnteringValidatorTopup = fmt.Sprintf("%.0f %s", float64(queueData.EnteringTopUpEthAmount)/1e9, utils.Config.Frontend.ClCurrency)
+			data.ExitingValidatorsBalance = fmt.Sprintf("%.0f %s", float64(queueData.LeavingEthAmount)/1e9, utils.Config.Frontend.ClCurrency)
+			data.EnteringValidators = total
```
remoterami, line 919: **issue**: `EnteringValidators` should be the number of new validators, correct? This would be the amount of new ETH

types/frontend.go
```diff
@@ -622,15 +622,15 @@ type ValidatorStateCountRow struct {
...
-	EnteringTotalEthAmount      uint64 // all deposits, topups + new validators
-	EnteringQueueTime           time.Duration
-	EnteringTopUpCount          uint64 // topups
-	TotalActiveEffectiveBalance uint64 // active network effective balance
-	LeavingValidatorCount       uint64
-	LeavingEthAmount            uint64
-	EnteringBalancePerDay       uint64
-	EnteringBalancePerEpoch     uint64
+	EnteringNewValidatorsCount     uint64 // fresh deposits count. Note is not deduplicated
+	EnteringNewValidatorsEthAmount uint64 // new validators
```
remoterami, line 626: **suggestion**: please adjust the comments here a bit. The first is the number of new validators, not a count of deposits. The second comment is also not too precise

remoterami (changes requested):
left some comments, ptal

services/services.go
manuelsc: EnteringValidators has no real usage after pectra, but template requires it be set to some non zero value. It's a cheap workaround I can do without having to refactor / cleanup that template mess - for which there wasn't really time considering the tight deadline

handlers/validator.go
```diff
@@ -636,21 +642,39 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
 				validatorPageData.ActivationEligibilityEpoch = pendingDeposit.EstClearEpoch + 1
+				if !validatorPageData.PendingDepositAboveMinActivation {
+					// countdown to when the deposit will be processed if it won't become eligible
+					validatorPageData.EstimatedActivationTs = utils.EpochToTime(pendingDeposit.EstClearEpoch)
+				}
+
 				isPendingActivation = true
 			}
 
 			if isPendingActivation {
```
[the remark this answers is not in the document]
manuelsc: `EstimatedActivationEpoch` is again because of template mess (*sigh*), it just needs to be a certain value so we can display the estimation when the deposit will be processed, not the estimated activation itself.

You are right about `ActivationEligibilityEpoch`, it was a bit abused to not need a helper var for the calculations below - but there is a UI in template that displays something based on this value. Let me check if I can refactor this quickly in time, otherwise I would pick this up later

remoterami (approved):
lgtm :+1:

types/frontend.go
Elenore23: Sorry but I didn't know strict rules,  I'm just trying to learn something and in that process I'm trying to not disturb now buddy. Thx for your advice I shall try to do my best.  Have a great day.  Irena..

PR #3062, opened 2025-04-28
Exits

manuelsc:


templates/slot/exits.html
```diff
@@ -10,7 +10,19 @@
       </div>
       <div class="row p-1 mx-0">
         <div class="col-md-2">Signature:</div>
-        <div class="col-md-10 text-monospace text-break">0x{{ printf "%x" $exit.Signature }}</div>
+        {{ if gt (len $exit.Signature) 0 }}
+          <div class="col-md-10 text-monospace text-break">0x{{ printf "%x" $exit.Signature }}</div>
+        {{ else }}
+          <div class="col-md-10 text-monospace text-break">-</div>
+        {{ end }}
```
LuccaBitfly, line 17: **nitpick (non-blocking):** Less HTML possible
```suggestion
        <div class="col-md-10 text-monospace text-break">
          {{ if gt (len $exit.Signature) 0 }}
            0x{{ printf "%x" $exit.Signature }}
          {{ else }}
            -
          {{ end }}
        </div>
```

handlers/validator.go
```diff
@@ -979,6 +979,44 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
+		err = db.ReaderDb.Select(&validatorPageData.ConsensusElExits, `
+			SELECT 
+				slot_processed AS slot,
+				block_processed_root AS block_root,
+				blocks_exit_requests.status,
+				COALESCE(reject_reason, '') AS reject_reason,
+				validator_pubkey,
+				'Execution Layer' as triggered_via
+			FROM blocks_exit_requests 
+			LEFT JOIN blocks ON blocks_exit_requests.block_processed_root = blocks.blockroot AND blocks.status = '1'
```
LuccaBitfly, line 992: **question:** Should this be a left join? You'd only want the proposed slots, right? E.g. when writing statistics, we do an inner join https://github.com/gobitfly/beaconchain/blob/19313ed838d4c3d07f4ad71afd47b30f6793ba3c/backend/pkg/commons/db/statistics.go#L1220

^ FYI I wrote that query, so I might be wrong.

LuccaBitfly (changes requested):

handlers/validator.go
manuelsc: Thanks good catch!

LuccaBitfly (approved):
lgtm

PR #3071, opened 2025-05-08
Not all attestations loading in slot page

manuelsc:


handlers/slot.go
```diff
@@ -161,10 +161,10 @@ func Slot(w http.ResponseWriter, r *http.Request) {
...
 }
 
-func getAttestationsData(slot uint64, onlyFirst bool) ([]*types.BlockPageAttestation, error) {
-	limit := ";"
-	if onlyFirst {
-		limit = " LIMIT 1;"
+func getAttestationsData(slot uint64, offset int64) ([]*types.BlockPageAttestation, error) {
+	limit := " LIMIT 1"
+	if offset > 0 {
+		limit += fmt.Sprintf(" OFFSET %d", offset)
```
LuccaBitfly, line 167: **nitpick (non-blocking):** Should do with SQL parameter placeholders (e.g. `LIMIT 1 OFFSET $2`) and not with string concatenation.

LuccaBitfly (approved):
looks good, one comment if you feel like doing it

PR #3074, opened 2025-05-12
fix: incorrectly showing withdrawals for pending validators that do n…

manuelsc:
…ot belong to that validator


MustardseedX (commented):
mistake correction

MustardseedX (commented):
correction to mistakes

PR #3078, opened 2025-05-12
misc: disable withdrawal tab if there is nothing to show

manuelsc:


handlers/validator.go
```diff
@@ -1128,13 +1128,20 @@ func Validator(w http.ResponseWriter, r *http.Request) {
 		return nil
 	})
 
+	var withdrawalCount uint64
+	g.Go(func() error {
+		withdrawalCount, err = db.GetTotalWithdrawalsCount([]uint64{validatorPageData.Index})
+		return err
```
LuccaBitfly, line 1134: **suggestion:** should wrap `err`, as all previous goroutines do.
```suggestion
		if err != nil {
			return fmt.Errorf("error getting withdrawal count: %w", err)
		}
		return nil
```

handlers/validator.go
```diff
@@ -1128,13 +1128,20 @@ func Validator(w http.ResponseWriter, r *http.Request) {
...
+	})
+
 	err = g.Wait()
 	if err != nil {
 		utils.LogError(err, "error getting validator data", 0)
 		http.Error(w, "Internal server error", http.StatusInternalServerError)
 		return
 	}
 
+	validatorPageData.EnableWithdrawalsTab = withdrawalCount > 0 || len(validatorPageData.ExecutionWithdrawals) > 0 || len(validatorPageData.ConsensusElExits) > 0
```
LuccaBitfly, line 1144: **suggestion:** Could combine `CappellaHasHappened` flag into this, so that in the html template you'd only need to check the `EnableWithdrawalsTab` flag.
```suggestion
	hasCapellaHappened := latestEpoch >= (utils.Config.Chain.ClConfig.CappellaForkEpoch)
	hasValidatorWithdrawals := withdrawalCount > 0 || len(validatorPageData.ExecutionWithdrawals) > 0 || len(validatorPageData.ConsensusElExits) > 0
	validatorPageData.EnableWithdrawalsTab = hasCapellaHappened || hasValidatorWithdrawals
```

LuccaBitfly (changes requested):

handlers/validator.go
LuccaBitfly: On a second thought, do we even need to check for `CappellaHasHappened`? If a validator has withdrawals, it must necessarily be true, right?

manuelsc: Yeah technically it's a redundant check. But I didn't want to get into testing whether it actually works on a pre Capella chain so I just kept it. Would argue for it to just leave it as is, wdyt?

LuccaBitfly (approved):
lgtm!!

PR #3082, opened 2025-05-20
feat: reference queue deposit request

remoterami:
Based on gobitfly/beaconchain#1584

### DIFF
--- a/exporter/queues.go
+++ b/exporter/queues.go
@@ -235,7 +235,51 @@ func (qi *PendingQueueIndexer) Index() error {
 		depositsList = append(depositsList, depositsToPostpone...)
 	}
 
-	return qi.save(depositsList)
+	err = qi.save(depositsList)
+	if err != nil {
+		return errors.Wrap(err, "failed to save pending deposits")
+	}
+	return qi.matchDepositRequests()
+}
+
+func (qi *PendingQueueIndexer) matchDepositRequests() error {
+	// matching will be wrong for postponed system-deposits
+	// but likelihood to ever occur for one pubkey, amount, slot combo is effectively 0
+	q := `
+	WITH pdq_ranked AS (
+		SELECT *, ROW_NUMBER() OVER (
+			PARTITION BY pubkey, amount, slot ORDER BY id
+		) AS rn
+		FROM pending_deposits_queue
+	),
+	bdr_ranked AS (
+		SELECT *, ROW_NUMBER() OVER (
+			PARTITION BY pubkey, amount, slot_queued ORDER BY id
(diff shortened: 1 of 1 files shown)


exporter/queues.go
```diff
@@ -235,7 +235,51 @@ func (qi *PendingQueueIndexer) Index() error {
...
+	}
+	return qi.matchDepositRequests()
+}
+
+func (qi *PendingQueueIndexer) matchDepositRequests() error {
+	// matching will be wrong for postponed system-deposits
+	// but likelihood to ever occur for one pubkey, amount, slot combo is effectively 0
+	q := `
+	WITH pdq_ranked AS (
+		SELECT *, ROW_NUMBER() OVER (
```
manuelsc, line 250: **question:** The entire reason for `rn` is so that if we have multiple deposits for the same pubkey, same amount and same slot, we still get to match them correctly, right? 
Is there a way that these two tables could diverge in terms of ordering within the slot? I just fear that even a single diverging entry will basically kill the entire matching for everything after but maybe this is unfounded.

If it is not unfounded, maybe matching "just" on slot, pubkey and amount is the better tradeoff. Sure it's not exact for this specific edge case but if it may lead to breaking matching for other entries we might be better of ignoring that edge-case.

exporter/queues.go
```diff
@@ -235,7 +235,51 @@ func (qi *PendingQueueIndexer) Index() error {
 		depositsList = append(depositsList, depositsToPostpone...)
 	}
 
-	return qi.save(depositsList)
+	err = qi.save(depositsList)
+	if err != nil {
+		return errors.Wrap(err, "failed to save pending deposits")
+	}
+	return qi.matchDepositRequests()
```
manuelsc, line 242: **suggestion:** maybe do save and match within the same transaction

exporter/queues.go
remoterami: > The entire reason for rn is so that if we have multiple deposits for the same pubkey, same amount and same slot, we still get to match them correctly, right?

Yes correct. Basically it's necessary because `pdq` is missing the `index_queued` column we could join on, which `bdr` does have. We could add that, but I think then your fear could manifest :skull: So imo the proposed solution should suffice, If I misunderstood I'd need an example 

> Is there a way that these two tables could diverge in terms of ordering within the slot? I just fear that even a single diverging entry will basically kill the entire matching for everything after but maybe this is unfounded.

No, I don't think so. The `rn` is basically just the SQL way to match the results in order without having an exploding results set

manuelsc (approved):
lgtm

PR #3083, opened 2025-05-21
fix: add reth to ethClients in valid subscription filter check

igorline:
I hope this should fix update notifications propagation for Reth


manuelsc (approved):

manuelsc:
Thank you for you contribution!

PR #3084, opened 2025-05-21
fix: remove rudimental support column value for reth client

igorline:
There's a value specified for seems to be rudimental column support for Reth client that is breaking the layout at https://beaconcha.in/user/ethClients

image


manuelsc:
Thank you for you contribution!

PR #3142, opened 2025-10-13
feat(api_docs): fully document v1 api

peterbitfly:


static/openapi/api/paths/api_v1_user_token.yaml
```diff
@@ -0,0 +1,49 @@
...
+          schema:
+            $ref: ../components/schemas/utils.OAuthErrorResponse.yaml
+    '500':
+      description: Internal Server Error
+      content:
+        application/json:
+          schema:
+            $ref: ../components/schemas/utils.OAuthErrorResponse.yaml
+  security:
+    - OAuthAccessCode: []
```
manuelsc: issue: this is the only endpoint with that security, don't think it got ever defined?

Makefile
```diff
@@ -21,10 +21,6 @@ explorer:
 	mkdir -p bin/
 	echo "Creating frontend asset bundle..."
 	go run cmd/bundle/main.go
-	echo "Generating swagger docs..."
```
manuelsc, line 24: Question: Don't we need a new build process via the redocly cli?

manuelsc (dismissed):
there's some AI querks here and there but it's fine, like we said we can iterate on it.

Makefile
peterbitfly: right, I adapted the build pipeline & the Dockerfile to automatically generate the docs bundle

matthias-bitfly (approved):

PR #3147, opened 2025-10-20
Fe 150/api key management section

rucister:


templates/user/settings.html
```diff
@@ -323,6 +323,18 @@ <h1 class="h4 mb-1 mb-md-0"><i class="mr-2 fas fa-user-circle"></i>Account Setti
               </div>
             {{ end }}
           {{ end }}
+          
+          <div class="alert alert-info alert-dismissible fade show my-3 p-3 d-flex " role="alert">
```
enzo-bitfly: issue: `alert` is meant to be used for dynamic changes that mostly require a time-sensitive action from the user. You can simply remove it, as the text should be announced when opening the tab panel.

enzo-bitfly: issue: `alert` is meant for dynamic changes, and it shouldn't contain interactive elements. You could use `role="region"` (or use a `<section>` with a heading) to convey to assistive technologies that this message is important.

types/config.go
```diff
@@ -299,6 +299,9 @@ type Frontend struct {
 	ElCurrencyDecimals int64         `yaml:"elCurrencyDecimals" envconfig:"FRONTEND_EL_CURRENCY_DECIMALS"`
 	MainCurrency       string        `yaml:"mainCurrency" envconfig:"FRONTEND_MAIN_CURRENCY"`
 	V2NotificationURL  string        `yaml:"v2NotificationUrl" envconfig:"FRONTEND_V2_NOTIFICATION_URL"`
+	APIKeyManagement	struct {
```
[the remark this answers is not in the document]
rucister: Done

templates/user/settings.html
rucister: Changed to a more relaxed `role="status"`.

PR #3152, opened 2025-10-23
feat(user): blur api key in user settings

peterbitfly:


templates/user/settings.html
```diff
@@ -829,11 +829,32 @@ <h3 class="h5">
...
-                            <span style="user-select: all;">{{ .Subscription.ApiKey }}</span>
+                            <span id="user-api-key" style="user-select: all;" data-api-key="{{ .Subscription.ApiKey }}"> <span class="api-key-start" style="user-select: all;"></span><span class="api-key-middle" style="user-select: all; filter: blur(4px);"></span><span class="api-key-end" style="user-select: all;"></span> </span>
                             <a style="color: #fff" id="copy-apiKey" data-toggle="tooltip" data-original-title="Copy API Key" data-clipboard-text="{{ .Subscription.ApiKey }}" class="btn btn-primary mb-1 btn-icon-inside">
                               <span class="fa fa-copy icon-inside text-white"></span>
                             </a>
                           </span>
+                          <script>
+                            ;(function () {
+                              var container = document.getElementById("user-api-key")
+                              if (!container) {
```
enzo-bitfly: question: is there a reason for container not to be always present?

peterbitfly: when we initially launched the explorer the API key functionality was not yet implemented, hence we have a small amount of old users that do not have an API key yet.

PR #3155, opened 2025-10-27
feat: add `enterprise plan card`

enzo-bitfly:
See: FE-153

### DIFF
--- a/templates/payment/pricing.html
+++ b/templates/payment/pricing.html
@@ -138,6 +138,7 @@
     .box-shadow {
       border: 1px solid var(--shadow-light);
     }
+
     .hot-box {
       border: 1px solid #ee7112;
     }
@@ -158,6 +159,14 @@
       appearance: none;
       -moz-appearance: none;
       -webkit-appearance: none;
+    }
+
+    .pricing__enterprise-plan-feature {
+      background: var(--bg-color-light);
+
+      p {
+        font-size: 0.875rem;
+      }
     }
   </style>
 {{ end }}
@@ -193,7 +202,7 @@
 
     <div class="container mt-2 container-pricing">
       <section class="my-5">
-        <div class="card-deck mb-3 justify-content-center text-center position-relative">
+        <div class="card-deck mb-3 justify-content-center text-center position-relative mb-5">
           <div style="margin: 0" class="row justify-content-center">
             <div class="card mx-2 mb-4 box-shadow">
               <div class="card-header">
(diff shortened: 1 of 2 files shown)


templates/payment/pricing.html
```diff
@@ -357,7 +368,43 @@ <h1 class="card-title pricing-card-title">399€* <small class="text-muted">/ mo
...
+        <div class="col-lg-8">
+          <div class="card">
+            <div class="card-body text-center pt-4">
+              <h2 class="mb-4">Enterprise Plan</h2>
+              <p class="col-md-8 mx-auto mb-4">Custom solutions for large-scale operations. Get dedicated support, custom rate limits, and enterprise-grade SLA.</p>
+              <a href="#contact-sales" class="btn btn-lg btn-outline-primary mb-4">Contact Sales</a>
+              <div class="row px-3">
+                <div class="col-md-4 p-2">
+                  <div class="pricing__enterprise-plan-feature card p-3 h-100">
+                    <i class="fas fa-infinity fa-2x mb-3 text-muted"></i>
```
rucister: suggestion: Not sure if a card inside a card would play well visually. I believe at least one of them should not have the drop-shadow.

templates/payment/pricing.html
```diff
@@ -357,7 +368,43 @@ <h1 class="card-title pricing-card-title">399€* <small class="text-muted">/ mo
           </div>
         </div>
       </section>
-      <section class="my-5">
+
+      <section class="row justify-content-center">
```
rucister: suggestion: In my opinion the width of the while block should be equal to the pricing block above and maybe not be a card at all.

templates/payment/pricing.html
```diff
@@ -135,9 +135,14 @@
 
 {{ define "css" }}
   <style>
+    html {
+      scroll-behavior: smooth;
```
rucister: question:  Shouldn't this be a global behavior?

templates/payment/pricing.html
enzo-bitfly: pair programming with @rucister - done with bg color

templates/payment/pricing.html
enzo-bitfly: pair programming with @rucister - done

templates/payment/pricing.html
enzo-bitfly: of course, moved to `layout.css`

PR #3172, opened 2026-07-09
Update README with Highsoft licensing information

nzcommander:
Added note about Highsoft chart library licensing and future plans.


nzcommander (commented):
Good

PR #3174, opened 2026-07-09
Update README with author credit and donation information

nzcommander:
Added donation information and author credit to README.


nzcommander:
Ok

